PR #8, opened 2019-07-19
Indent line comments correctly when they precede a continuation line.

allevato:
If a line comment occurs after a continuation break, then its indentation depends on what comes on the line following it:

* If the following line is an indented continuation line, then the comment should also be indented as a continuation. For example,

  ```swift
  return true
    // comment
    && false
  ```

* Otherwise, if the following line is not a continuation (such as a new statement, a closing delimiter, or the end of the file), then the comment should not be indented as a continuation:

  ```swift
  {
    return true
    // comment
  }
  ```

This is achieved by adding a reset break after the last item in a code block (rather than just _between_ them) and by not resetting the continuation state unconditionally when printing a comment.

Fixes [SR-11107](https://bugs.swift.org/browse/SR-11107).

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/CommentTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/CommentTests.swift
@@ -416,4 +416,17 @@ public class CommentTests: PrettyPrintTestCase {
     
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 100)
   }
+
+  public func testCommentOnContinuationLine() {
+    let input =
+      """
+      func foo() {
+        return true
+          // comment
+          && false
+      }
+      """
+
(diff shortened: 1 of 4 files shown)


Tests/SwiftFormatPrettyPrintTests/CommentTests.swift
```diff
@@ -416,4 +416,27 @@ public class CommentTests: PrettyPrintTestCase {
...
+    let input =
+      """
+      func foo() {
+        return true
+          // comment
+          && false
+      }
+      """
+
+    let expected =
```
akyrtzi: `input` and `expected` are almost identical, do they need to be repeated?

allevato: Updated to compare against `input + "\n"` instead (we do this in some other tests).

PR #11, opened 2019-07-20
[SR-11115] Missing type annotation on multiple declarations

kitasuke:
Produced source code is missing type annotation only when multiple declarations with only one type because the last one only has it. So I set its type to the rest of decls when necessary 

Fixes [SR-11115](https://bugs.swift.org/browse/SR-11115).

### DIFF
--- a/Sources/SwiftFormatRules/OneVariableDeclarationPerLine.swift
+++ b/Sources/SwiftFormatRules/OneVariableDeclarationPerLine.swift
@@ -54,8 +54,14 @@ public final class OneVariableDeclarationPerLine: SyntaxFormatRule {
       // The first binding corresponds to the original `var`/`let`
       // declaration, so it should not have its trivia replaced.
       var isFirst = true
+      // The last binding only has type annotation when multiple
+      // declarations in one line, so make a new binding with its type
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatRules/OneVariableDeclarationPerLine.swift
```diff
@@ -54,8 +54,14 @@ public final class OneVariableDeclarationPerLine: SyntaxFormatRule {
...
       // declaration, so it should not have its trivia replaced.
       var isFirst = true
+      // The last binding only has type annotation when multiple
+      // declarations in one line, so make a new binding with its type
+      let typeAnnotation = varDecl.bindings.last?.typeAnnotation
       for binding in varDecl.bindings {
-        let newBinding = binding.withTrailingComma(nil)
+        var newBinding = binding.withTrailingComma(nil)
+        if typeAnnotation != nil && binding.typeAnnotation == nil {
+            newBinding = newBinding.withTypeAnnotation(typeAnnotation)
```
allevato: Super-nit: Can you use 2-space indentation here?

kitasuke: Oops, fixed it

PR #12, opened 2019-07-20
Add lint/format APIs that take source text directly.

allevato:


Sources/SwiftFormat/SwiftFormatter.swift
```diff
@@ -61,6 +61,21 @@ public final class SwiftFormatter {
...
+  /// Formats the given Swift source code and writes the result to an output stream.
+  ///
+  /// - Parameters:
+  ///   - source: The Swift source code to be formatted.
+  ///   - url: A file URL denoting the filename/path that should be assumed for this source code.
+  ///   - outputStream: A value conforming to `TextOutputStream` to which the formatted output will
+  ///     be written.
+  /// - Throws: If an unrecoverable error occurs when formatting the code.
+  public func format<Output: TextOutputStream>(
+    source: String, assumingFileURL url: URL, to outputStream: inout Output
```
akyrtzi: Why is a `url` necessary for the formatting operation?

allevato: Right now we still emit diagnostics during formatting for things that can't be fixed automatically (for example, a comment with a URL that exceeds the line length limit), so the URL is used so that the originating file can be associated with those diagnostics.

akyrtzi: How about we make this optional (and maybe assign some dummy url if none is given) in case the client doesn't really care about this or the context of where the source text came from is lost or it didn't come from a file in the first place?

allevato: Done.

PR #13, opened 2019-07-21
[SR-11109] swift-format produces invalid enum for raw values

kitasuke:
## Overview

This PR fixes unexpected format for enum with incremental raw values for `Int` or `Float`.

I applogize that I forgot to assign this issue to me on bugs.swift when I started working on, but I gave it a try to fix the issue. However, this approach doesn't look elegant in my opinion. If that's the case, please ignore this PR since @allevato should be working on the fix as well.

Fixes [SR-11109](https://bugs.swift.org/browse/SR-11109).

### DIFF
--- a/Sources/SwiftFormatRules/OneCasePerLine.swift
+++ b/Sources/SwiftFormatRules/OneCasePerLine.swift
@@ -46,9 +46,18 @@ public final class OneCasePerLine: SyntaxFormatRule {
             let newRemovedDecl = createAssociateOrRawCaseDecl(
               fullDecl: caseMember,
               removedElement: element)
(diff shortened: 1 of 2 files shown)


allevato:
Thanks! However, I just pushed https://github.com/apple/swift-format/pull/15 which is a deeper rewrite of the rule simplify it and do a simple linear scan instead of the original behavior that tried to remove and re-add elements/cases, and covers a couple more edge cases with comments/attributes.

allevato:
#15 has now been merged, so I'll close this one out.

kitasuke:
Your PR looks so much better!

PR #14, opened 2019-07-22
Type safe code generation for pipeline-generation

kitasuke:
## Overview
Current way of code generation is string based, so I made a change to make it type safe. However, that loses code readability a lot. It might be a good idea to adapt this way to showcase usage of code gen using SwiftSyntax, but I'm totally okay to keep current way for code readability

## Caution
Current version of SwiftSyntax has a bug to make `StringLiteralExpr` with trivia. We can't merge this PR until [this fix](https://github.com/apple/swift-syntax/pull/135) is merged for the safety.


allevato:
I agree that the readability suffers quite a bit with this approach ☹️   Using string interpolation makes it clear what the final code will look like and makes it easier to adjust.

I think the main problem is that SwiftSyntax is too low-level to do code generation on its own, because you have to manually curate things like spaces, newlines, and even the commas inside function calls. What you really want in a code generator is to say "just generate a function call with these arguments" and have it handle the unimportant details like ensuring that the right elements have trailing commas and that statements are separated by newlines. Then you can take that tree and run it through the swift-format pretty printer to clean it up.

So ultimately what we would need is a layer/DSL on top of SwiftSyntax that makes these kinds of constructions easier. Maybe @akyrtzi has some thoughts in this area?

akyrtzi:
I've reached out to @akkyie for getting https://github.com/akkyie/SyntaxBuilder upstreamed.

kitasuke:
Yes, it would be nice to have DSL on top of SwiftSyntax!

kitasuke:
I'll close this PR since SyntaxBuilder is not ready yet in SwiftSyntax

PR #32, opened 2019-08-07
Migrate leading trivia when adding an access-level modifier.

dylansturg:
The AddModifierRewriter was previously only removing the leading trivia of the declaration's first token. The assumption that the new modifier must be the declaration's first token is incorrect. Instead, the AddModifierRewriter now transfers trivia from the token immediately following the new modifier without making any assumptions about where the modifier will be placed in the declaration. This transfer of trivia only occurs for the first modifier in a declaration. The code that adds modifiers to an existing list of modifiers already correctly moves trivia when adding modifiers.

This fixes a bug where leadingTrivia that belongs on attributes (e.g. `@objc`) was moved to the access-level modifier. Moving the leadTrivia trivia resulted in breaking up the attributes and modifiers in a declaration.

### DIFF
--- a/Sources/SwiftFormatRules/AddModifierRewriter.swift
+++ b/Sources/SwiftFormatRules/AddModifierRewriter.swift
@@ -22,13 +22,11 @@ private final class AddModifierRewriter: SyntaxRewriter {
   }
 
   override func visit(_ node: VariableDeclSyntax) -> DeclSyntax {
-    // Check for modifiers, if none, put accessor keyword before the first token
+    // Check for modifiers, and, if none, insert the modifier and relocate trivia from the displaced
+    // token.
     guard let modifiers = node.modifiers else {
-      guard
-        let newDecl = removeFirstTokLeadingTrivia(node: node)
-        as? VariableDeclSyntax
-      else { return node }
-      return newDecl.addModifier(modifierKeyword)
+      let nodeWithModifier = node.addModifier(modifierKeyword)
+      return nodeByRelocatingTrivia(in: nodeWithModifier) { $0.modifiers }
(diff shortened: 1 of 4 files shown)


Sources/SwiftFormatRules/AddModifierRewriter.swift
```diff
@@ -39,115 +37,127 @@ private final class AddModifierRewriter: SyntaxRewriter {
...
-    let withoutLeadTrivia = replaceTrivia(
+  /// Moves trivia in the given node to correct the placement of potentially displaced trivia in the
+  /// node after the first modifier was added to the given node. The added modifier is assumed to be
+  /// the first and only modifier of the node. After the first modifier is added to a node, any
+  /// leading trivia on the token immediately after the modifier is considered displaced. This
+  /// method moves that displaced trivia onto the new modifier. When there is no displaced trivia,
+  /// this method does nothing and returns the given node as-is.
+  /// - Parameter node: A node that was updated to include a new modifier.
+  /// - Parameter modifiersProvider: A closure that returns all modifiers for the given node.
+  func relocateTrivia<NodeType: DeclSyntax>(
```
allevato: Swift naming conventions for methods (which, to be completely fair, are spottily followed through the formatter codebase) typically use verb phrases for `Void`-returning methods and noun phrases that describe the thing being returned for everything else.

So a better name for this method might be `nodeByRelocatingTrivia(in:for:)`, which adds clarity to the call sites about what the output of the method is.

dylansturg: Renamed. Thanks for the additional context!

PR #35, opened 2019-08-10
[SR-11291] Missing space after semicolon in swift-format

kitasuke:
Missing space only if two code block items exist in one line with semicolon

Fixes [SR-11291](https://bugs.swift.org/browse/SR-11291).

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -848,7 +848,8 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: MemberDeclListItemSyntax) -> SyntaxVisitorContinueKind {
     before(node.firstToken, tokens: .open)
-    after(node.lastToken, tokens: .close, .break(.reset, size: 0))
+    let resetSize = node.semicolon != nil ? 1 : 0
+    after(node.lastToken, tokens: .close, .break(.reset, size: resetSize))
     return .visitChildren
   }
 
@@ -934,7 +935,8 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: CodeBlockItemSyntax) -> SyntaxVisitorContinueKind {
     before(node.firstToken, tokens: .open)
-    after(node.lastToken, tokens: .close, .break(.reset, size: 0))
+    let resetSize = node.semicolon != nil ? 1 : 0
+    after(node.lastToken, tokens: .close, .break(.reset, size: resetSize))
     return .visitChildren
   }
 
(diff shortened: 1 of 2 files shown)


kitasuke:
You're right. Applied same behavior for `MemberDeclBlockItem`.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -934,7 +935,8 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: CodeBlockItemSyntax) -> SyntaxVisitorContinueKind {
     before(node.firstToken, tokens: .open)
-    after(node.lastToken, tokens: .close, .break(.reset, size: 0))
+    let resetSize = node.semicolon != nil ? 1 : 0
```
kitasuke: Let me know if this calculation of reset size should be a common helper function to be called by `CodeBlockItemSyntax` and `MemberDeclListItemSyntax`

allevato: I don't think that's necessary; at this point I don't foresee us wanting to generalize the number of spaces to anything else, so factoring out doesn't buy us much.

It's interesting to observe that these two methods are identical, so we _could_ go as far as factor them out in their entirety into a single helper function, but I don't think we want to do that either. It's conceivable that at some point in the future we'd want to handle multi-statement lines differently than multi-decl lines, and combining the logic would make that more difficult. So I think it's all fine the way it is.

Tests/SwiftFormatPrettyPrintTests/SemicolonTests.swift
```diff
@@ -0,0 +1,24 @@
+public class SemiColonTypeTests: PrettyPrintTestCase {
```
allevato: Can you add a couple test cases like the ones you already have, but with the members wrapped in a type? For example,

```swift
struct Foo {
  var foo = false; var bar = true; var baz = false
}
```

```swift
struct Foo {
  var foo = false
  var bar = true
  var baz = false
}
```

In the existing tests, since the `var`s aren't direct children of a type declaration, they'll be simple `CodeBlockItem`s rather than `MemberDeclListItem`s, so we don't have any test coverage for that part yet.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
kitasuke: Agreed

Tests/SwiftFormatPrettyPrintTests/SemicolonTests.swift
kitasuke: I don't know why I didn't do this... Fixed the test cases!

PR #37, opened 2019-08-11
Remove a lie from the documentation

Trzyipolkostkicukru:
`lineBreakBeforeEachArgument` isn't the default.

### DIFF
--- a/Documentation/Configuration.md
+++ b/Documentation/Configuration.md
@@ -54,9 +54,9 @@ top-level keys and values:
 
 *   `lineBreakBeforeEachArgument` _(boolean)_: Determines the line-breaking
     behavior for generic arguments and function arguments when a declaration is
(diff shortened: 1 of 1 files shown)


Documentation/Configuration.md
```diff
@@ -54,7 +54,7 @@ top-level keys and values:
 
 *   `lineBreakBeforeEachArgument` _(boolean)_: Determines the line-breaking
     behavior for generic arguments and function arguments when a declaration is
-    wrapped onto multiple lines. If true (the default), a line break will be
+    wrapped onto multiple lines. If true, a line break will be
```
allevato: Can you move the "(the default)" part below to follow "If false", and rewrap this paragraph as needed so that it fills up to 80 columns?

Trzyipolkostkicukru: Sure!

PR #38, opened 2019-08-12
Allow attributes, including `available`, to span multiple lines.

dylansturg:
The previous implementation forced all attributes to use a single line, even if the attribute's identifier + contents exceeded the line length. Following some investigation, it appears that the Swift compiler supports multi-line attribute declarations.

The formatter uses the following priority for laying out attributes:
1. Whenever possible, fit the entire attribute on a single line.
2. Otherwise, if the arguments can fit on a line, use line breaks between the attribute's opening paren, the arguments, and the closing paren.
3. Otherwise use the configured argument line breaking option (as is used in argument lists for functions and function like declarations) to add line breaks between the arguments. Prefers keeping individual arguments together, but there's a carve specifically for "labeled availability arguments" to allow breaking at the ":" separator.

In the future, we could add more special case line breaks in specific kinds of attribute arguments (e.g. line breaks between pieces of an Objective-C selector).

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1016,32 +1016,36 @@ private final class TokenStreamCreator: SyntaxVisitor {
     return .visitChildren
   }
 
-  func handleAvailabilitySpec(leftParen: TokenSyntax?, spec: Syntax, rightParen: TokenSyntax?) {
-    var tokens: [TokenSyntax] = []
-    if let leftParen = leftParen {
-      tokens.append(leftParen)
+  func visit(_ node: AttributeSyntax) -> SyntaxVisitorContinueKind {
+    before(node.firstToken, tokens: .open)
+    if node.argument != nil {
+      // Wrap the attribute's arguments in their own group, so arguments stay together with a higher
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1016,35 +1016,41 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
     }
+    after(node.lastToken, tokens: .close)
+    return .visitChildren
   }
 
-  func visit(_ node: AttributeSyntax) -> SyntaxVisitorContinueKind {
-    if let argument = node.argument {
-      handleAvailabilitySpec(leftParen: node.leftParen, spec: argument, rightParen: node.rightParen)
+  func visit(_ node: AvailabilitySpecListSyntax) -> SyntaxVisitorContinueKind {
+    for argument in node.dropLast() {
```
allevato: There's a helper method defined earlier in this file named `insertTokens(_:betweenElementsOf:)` that does this, so you should be able to write:

```swift
insertTokens(.break(.same, size: 1), betweenElementsOf: node)
```

dylansturg: Perfect. Done.

PR #39, opened 2019-08-12
Allow line breaks between Obj-C selector pieces in attributes.

dylansturg:
Previously, Objective-C selector pieces were forced onto 1 line regardless of their length. This is problematic when the selector is verbose or has many arguments, since it will overflow the line length. It's still possible for an Objective-C selector piece whose name is longer than the configured line length to overflow the line length, since it's invalid to break inside of the name of a selector piece.

These breaks are specifically size 0 because Objective-C selector pieces aren't separated by spaces.

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/AttributeTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/AttributeTests.swift
@@ -172,11 +172,13 @@ public class AttributeTests: PrettyPrintTestCase {
       input: input, expected: expected, linelength: 40, configuration: configuration)
   }
 
-  public func testObjCAttributes() {
+  public func testObjCBinPackedAttributes() {
     let input =
       """
       @objc func f() {}
-      @objc(foo:bar:baz)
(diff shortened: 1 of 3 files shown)


Tests/SwiftFormatPrettyPrintTests/AttributeTests.swift
```diff
@@ -185,16 +187,107 @@ public class AttributeTests: PrettyPrintTestCase {
...
+       """
+
+     let configuration = Configuration()
+     configuration.lineBreakBeforeEachArgument = true
+     assertPrettyPrintEqual(
+       input: input, expected: expected, linelength: 40, configuration: configuration)
+   }
+
+  public func testObjCAttributesDiscretionaryLineBreaking() {
+    // The disretionary newlines in the 3rd function declaration are invalid, because new lines
```
allevato: Nit: s/disretionary/discretionary/

dylansturg: Thanks!

PR #42, opened 2019-08-19
Support formatting and linting from standard input.

dylansturg:
Reading standard input is a feature that clang-format supports which is useful for users of swift-format too.

CLI changes:
- You can now not specify any directory/file paths, and the tool will read from standard input up to the first EOF.
- When no directory/file paths are specified, the in-place and recursive options are not allowed.
- You can specify a new assume-filename option if and only if using standard input to provide a filename to include in output diagnostics.

I copied clang-format's approach to invoking standard input mode, where the tool starts reading standard input whenever there were no paths specified.

### DIFF
--- a/Sources/SwiftFormat/SwiftLinter.swift
+++ b/Sources/SwiftFormat/SwiftLinter.swift
@@ -64,7 +64,7 @@ public final class SwiftLinter {
   ///   - url: A file URL denoting the filename/path that should be assumed for this source code.
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func lint(source: String, assumingFileURL url: URL) throws {
-    let sourceFile = try SyntaxParser.parse(url)
+    let sourceFile = try SyntaxParser.parse(source: source)
     try lint(syntax: sourceFile, assumingFileURL: url)
   }
 
--- a/Sources/swift-format/Run.swift
+++ b/Sources/swift-format/Run.swift
@@ -20,21 +20,33 @@ import SwiftSyntax
 /// Runs the linting pipeline over the provided source file.
 ///
 /// If there were any lint diagnostics emitted, this function returns a non-zero exit code.
-/// - Parameter configuration: The `Configuration` that contains user-specific settings.
-/// - Parameter path: The absolute path to the source file to be linted.
+/// - Parameters:
+///   - configuration: The `Configuration` that contains user-specific settings.
+///   - sourceFile: A file handle from which to read the source code to be linted.
+///   - assumingFilename: The filename of the source file, used in diagnostic output.
 /// - Returns: Zero if there were no lint errors, otherwise a non-zero number.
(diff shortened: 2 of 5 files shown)


Sources/SwiftFormat/SwiftLinter.swift
```diff
@@ -64,7 +64,7 @@ public final class SwiftLinter {
   ///   - url: A file URL denoting the filename/path that should be assumed for this source code.
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func lint(source: String, assumingFileURL url: URL) throws {
-    let sourceFile = try SyntaxParser.parse(url)
+    let sourceFile = try SyntaxParser.parse(source: source)
```
allevato: Oopsie, thanks for catching that.

Sources/swift-format/Run.swift
```diff
@@ -17,6 +17,42 @@ import SwiftFormatConfiguration
...
 
+/// Runs the linting pipeline over a source file that will be read from standard input. Optionally,
+/// a filename can be provided to use in the output diagnostics.
+///
+/// If there were any lint diagnostics emitted, this function returns a non-zero exit code.
+/// - Parameters:
+///   - configuration: The `Configuration` that contains user-specific settings.
+///   - assumingFilename: The filename of the file that contains the source code.
+/// - Returns: Zero if there were no lint errors, otherwise a non-zero number.
+func lintMain(configuration: Configuration, assumingFilename: String?) -> Int {
```
allevato: I'm a bit uncomfortable having separate entry points for the `stdin` case and the file case because any logic that we may want to modify in one will need to be kept in sync with the other.

The main tricky thing here is that the `stdin` entry point calls the API that takes a source string and assumed URL, and the file-based API takes the URL directly.

What if we have this function take a `FileHandle` and an assumed filename, and then this function can always read from the file handle and call the `(source:assumingFileURL:)` API? Then the caller of this function can pass in either `FileHandle.standardInput` or a `FileHandle` to the requested file on disk, and we can converge on a single implementation of `lintMain` and `formatMain`.

That means that the `swift-format` tool will no longer use the versions of the API that take a URL directly, but that's fine—those APIs are still useful to have for other potential tool integrations aside from ours.

dylansturg: That makes sense. I agree having to keep changes in sync is not ideal. I've refactored the existing methods (deleting the new methods) to use FileHandle arguments to read the source code.

I ended up refactoring main.swift a bit to share the code in the FileIterator loop, since I needed to add some more identical code to that loop.

Sources/SwiftFormat/SwiftLinter.swift
dylansturg: I spent a while scratching my head... "That's not the source code I put on `stdin`....". 😅

PR #43, opened 2019-08-19
Add option to not indent conditional compilation blocks

alejandro-isaza:
When adding conditional compilation to a large block of code it is often undesirable to have that large block be indented.

### DIFF
--- a/Documentation/Configuration.md
+++ b/Documentation/Configuration.md
@@ -59,6 +59,10 @@ top-level keys and values:
     If false (the default), arguments will be laid out horizontally first, with
     line breaks only being fired when the line length would be exceeded.
 
+*   `indentConditionalCompilationBlocks` _(boolean)_: Determines if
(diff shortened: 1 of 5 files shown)


keith:
Would this also mean this formatting would work:

```swift
func foo() {
#if BAR
    print("hi")
#endif
}
```

Or would it de-ident it to:

```swift
func foo() {
#if BAR
print("hi")
#endif
}
```

allevato:
There's no ability yet for lines starting with `#` to be always be flush-left; this change only affects the indentation of the _contents_ of a conditional compilation block.

So if you write this:

```swift
func foo() {
#if BAR
  print("hi")
#endif
}
```

Then the only possibilities are this if the configuration option is `true` (default):

```swift
func foo() {
  #if BAR
    print("hi")
  #endif
}
```

and this if it's `false`:

```swift
func foo() {
  #if BAR
  print("hi")
  #endif
}
```

PR #48, opened 2019-09-05
Detect and delete semicolons nested in code blocks.

dylansturg:
The DoNotUseSemicolons rule previously only removed semicolons that were on the top level code block items inside of a code block item list. It's possible for additional code block item lists to be nested inside of a code block item; see the new test for an example.

The rule now recursively walks through all children of each code block item to find and remove semicolons.

The leading trivia replacement to add a newline was stripping all newlines from the existing trivia, and then adding 1 newline. This is a problem when the leading trivia already contains 2 or more leading newlines, for example to create a blank line between 2 code block items. The previous behavior would remove the intended blank line. Instead, now a single newline is appended to the leading trivia if and only if it doesn't already contain at least 1 leading newline.

### DIFF
--- a/Sources/SwiftFormatRules/DoNotUseSemicolons.swift
+++ b/Sources/SwiftFormatRules/DoNotUseSemicolons.swift
@@ -22,52 +22,75 @@ import SwiftSyntax
 ///
 /// - SeeAlso: https://google.github.io/swift#semicolons
 public final class DoNotUseSemicolons: SyntaxFormatRule {
-  private func transformItems(_ items: CodeBlockItemListSyntax) -> CodeBlockItemListSyntax {
-    var newItems = Array(items)
+
+  /// Creates a new version of the given node which doesn't contain any semicolons. The node's
+  /// items are recursively modified to remove semicolons, replacing with line breaks where needed.
+  /// Items are checked recursively to support items that contain code blocks, which may have
+  /// semicolons to be removed.
+  /// - Parameters:
+  ///   - node: A node that contains items which may have semicolons or nested code blocks.
+  ///   - nodeCreator: A closure that creates a new node given an array of items.
+  private func nodeByRemovingSemicolons<
+    ItemType: Syntax & SemicolonSyntax & Equatable, NodeType: SyntaxCollection
+  >(from node: NodeType, nodeCreator: ([ItemType]) -> NodeType) -> NodeType
+  where NodeType.Element == ItemType {
+    var newItems = Array(node)
 
     // Because newlines belong to the _first_ token on the new line, if we remove a semicolon, we
     // need to keep track of the fact that the next statement needs a new line.
     var previousHadSemicolon = false
-    for (idx, item) in items.enumerated() {
+    for (idx, item) in node.enumerated() {
 
       // Store the previous statement's semicolon-ness.
       defer { previousHadSemicolon = item.semicolon != nil }
 
+      // Check for semicolons in statements inside of the item, because code blocks may be nested
+      // inside of other code blocks.
+      guard let visitedItem = visit(item) as? ItemType else {
+        return node
+      }
+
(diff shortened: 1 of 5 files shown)


Sources/SwiftFormatRules/DoNotUseSemicolons.swift
```diff
@@ -77,3 +98,12 @@ extension Diagnostic.Message {
   static let removeSemicolonAndMove = Diagnostic.Message(
     .warning, "remove ';' and move the next statement to a new line")
 }
+
+/// Protocol that declares support for accessing and modifying a token that represents a semicolon.
```
allevato: Can you toss this (and the extensions below) in a separate file (SemicolonSyntax.swift)? We have various extensions/convenience things like this that we should look at someday contributing back to swift-syntax, so keeping it separate from the rule will make sure that we don't forget about it as easily.

Sources/SwiftFormatRules/DoNotUseSemicolons.swift
```diff
@@ -22,52 +22,73 @@ import SwiftSyntax
...
       newItems[idx] = newItem
     }
-    return SyntaxFactory.makeCodeBlockItemList(newItems)
+    return nodeCreator(newItems)
   }
 
-  public override func visit(_ node: CodeBlockSyntax) -> Syntax {
-    return node.withStatements(transformItems(node.statements))
+  public override func visit(_ node: CodeBlockItemListSyntax) -> Syntax {
+    return nodeByRemovingSemicolons(from: node) { SyntaxFactory.makeCodeBlockItemList($0) }
```
allevato: It's equivalent, but when you have a closure where you're just passing through the arguments to another function in the same order, you can usually pass a reference to that function directly, so this should also work:

```swift
nodeByRemovingSemicolons(from: node, nodeCreator: SyntaxFactory.makeCodeBlockItemList)
```

My personal style is to prefer function references instead of creating new closure literals when possible, but it's not a *strong* preference and I've also never bothered looking into whether there were particular advantages to one or the other in terms of compiler performance.

Tests/SwiftFormatRulesTests/DoNotUseSemicolonsTests.swift
```diff
@@ -18,4 +18,79 @@ public class DoNotUseSemicolonsTests: DiagnosingTestCase {
...
+                /// This is a doc comment for printing "goodbye".
+                print("goodbye")
+
+                /// This is a doc comment for printing "3".
+                print("3")
+
+                print("4")
+                /** Inline comment. */ print("5")
+
+                print("6")// This is an important statement.
```
allevato: So it looks like the two spaces after the semicolon in the original are the trailing trivia of the semicolon? I guess it's fine to discard that because the pretty printer will fix it up (and if there was a non-comment token here instead, the spaces wouldn't be trailing trivia, so it wouldn't be an issue). It might be worth a comment somewhere in the implementation (or the test?) calling this case out though as something we don't need to worry about.

dylansturg: I added a comment in the implementation that the trailing trivia is being discarded. It seems that the trailing trivia is only spaces up to but not including the first newline or comment. The pretty printer will fix any missing spaces, so I didn't try to merge the trailing trivia into a different token.

Sources/SwiftFormatRules/DoNotUseSemicolons.swift
dylansturg: Done.

Sources/SwiftFormatRules/DoNotUseSemicolons.swift
dylansturg: I forgot about that! Avoiding unnecessary closures seems like a win to me. Changed.

PR #50, opened 2019-09-09
Make linter messages more grammatically and structurally consistent.

dylansturg:
I read through all of the existing linter messages, and found the following rules to apply in general. The changes here bring these linter messages in sync with the format of the other existing linter messages.

- Use sentence fragments, starting with lowercase and **not** ending with a period.
- Use the [imperative form](https://en.wikipedia.org/wiki/Imperative_mood) so that the message is a command, instructing the user what to change.
- Start with a verb describing what needs to be changed, potentially include the conditions under which the change is necessary and/or a rationale.

### DIFF
--- a/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
+++ b/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
@@ -414,21 +414,21 @@ public class WhitespaceLinter {
 
 extension Diagnostic.Message {
   static let trailingWhitespaceError = Diagnostic.Message(
-    .warning, "[TrailingWhitespace]: remove trailing whitespace.")
+    .warning, "[TrailingWhitespace]: remove trailing whitespace")
 
   static func indentationError(_ spaces: Int) -> Diagnostic.Message {
-    return .init(.warning, "[Indentation]: indentation should be \(spaces) spaces.")
+    return .init(.warning, "[Indentation]: indentation should be \(spaces) spaces")
   }
 
   static func spacingError(_ spaces: Int) -> Diagnostic.Message {
-    return .init(.warning, "[Spacing]: should be \(spaces) spaces.")
+    return .init(.warning, "[Spacing]: should be \(spaces) spaces")
   }
 
-  static let removeLineError = Diagnostic.Message(.warning, "[RemoveLine]: remove line break.")
+  static let removeLineError = Diagnostic.Message(.warning, "[RemoveLine]: remove line break")
 
   static func addLinesError(_ lines: Int) -> Diagnostic.Message {
-    return .init(.warning, "[AddLines]: add \(lines) line breaks.")
+    return .init(.warning, "[AddLines]: add \(lines) line breaks")
   }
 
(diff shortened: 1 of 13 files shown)


Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
```diff
@@ -414,21 +414,21 @@ public class WhitespaceLinter {
 
 extension Diagnostic.Message {
   static let trailingWhitespaceError = Diagnostic.Message(
-    .warning, "[TrailingWhitespace]: remove trailing whitespace.")
+    .warning, "[TrailingWhitespace]: remove trailing whitespace")
 
   static func indentationError(_ spaces: Int) -> Diagnostic.Message {
-    return .init(.warning, "[Indentation]: indentation should be \(spaces) spaces.")
+    return .init(.warning, "[Indentation]: indentation should be \(spaces) spaces")
```
allevato: Should we make this one and the one below imperative? We'd need to change the API to take the actual and expected values so that we can do the math and say "indent/unindent by N spaces" and "add/remove N spaces" instead of the indicative mood message we have right now.

If we do it I think it should be a separate PR, I'm just curious about your thoughts.

Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
```diff
@@ -414,21 +414,21 @@ public class WhitespaceLinter {
...
-  static let removeLineError = Diagnostic.Message(.warning, "[RemoveLine]: remove line break.")
+  static let removeLineError = Diagnostic.Message(.warning, "[RemoveLine]: remove line break")
 
   static func addLinesError(_ lines: Int) -> Diagnostic.Message {
-    return .init(.warning, "[AddLines]: add \(lines) line breaks.")
+    return .init(.warning, "[AddLines]: add \(lines) line breaks")
   }
 
-  static let lineLengthError = Diagnostic.Message(.warning, "[LineLength]: line is too long.")
+  static let lineLengthError = Diagnostic.Message(.warning, "[LineLength]: line is too long")
```
allevato: This is a weird one too—I'm not sure how we make it imperative unless we say "shorten the line", which seems demonstrably worse. It's probably best left as is.

Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
dylansturg: I think that would be a good change. It would be consistent with the missing new lines error and is slightly more specific. Given that it's a more substantial change than rewording, I also think a separate PR makes sense.

Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
dylansturg: I spent some time thinking about how to make this imperative, but I couldn't come up with anything that I felt was an improvement. It's difficult to state what exactly needs to change.

I think the current text is a good compromise compared to trying to make a very vague imperative statement (e.g. "make this line shorter") or trying to figure out a specific change to make (e.g. "add a line break at column N"). I vote for keeping the current wording.

PR #52, opened 2019-09-09
Use imperative mood in the diagnostics from the WhitespaceLinter.

dylansturg:
Follow up from comments on https://github.com/apple/swift-format/pull/50.

### DIFF
--- a/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
+++ b/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
@@ -139,7 +139,7 @@ public class WhitespaceLinter {
     // Only run this check at the start of a line.
     guard
       (user.count > 1 && form.count > 1)
-      || (form.count == 1 && form.count == 1 && isFirstCharacter)
+        || (form.count == 1 && form.count == 1 && isFirstCharacter)
     else {
       return
     }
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
```diff
@@ -139,7 +139,7 @@ public class WhitespaceLinter {
     // Only run this check at the start of a line.
     guard
       (user.count > 1 && form.count > 1)
-      || (form.count == 1 && form.count == 1 && isFirstCharacter)
+        || (form.count == 1 && form.count == 1 && isFirstCharacter)
```
dylansturg: @allevato Is this spacing right? It was added when I formatted this file. I'm not sure there's supposed to be an additional indent here.

allevato: The break preceding the `||` is a `.continuation` break, so for this indentation to occur I believe the break following the `guard` would have to be an `.open` break, which sounds right based on changes we've made in the past to get clauses to line up.

I think it's fine; arguably the `||` in this case is a continuation of the expression the line above, not an "equal child" of the `guard` compared to the first line. But I guess it looks odd since the first line is forced onto its own line.

PR #53, opened 2019-09-10
[Format Range] Format in a specific range POC

denis631:
- Add application range var to the Formatting Context
- Pass `TokenSyntax` to the PrettyPrinter
- PrettyPrinter uses the leading and trailing trivia in case when token is not in applicable range.
- Verify its work by having some minimal tests

Previous discussions here: https://forums.swift.org/t/range-formatting/28060

PS. I was thinking it would be a good idea still to make a PR and continue discussions here, potentially improving the PR and eventually merging.

### DIFF
--- a/Sources/SwiftFormat/SwiftFormatter.swift
+++ b/Sources/SwiftFormat/SwiftFormatter.swift
@@ -48,7 +48,7 @@ public final class SwiftFormatter {
   ///     be written.
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func format<Output: TextOutputStream>(
-    contentsOf url: URL, to outputStream: inout Output
+    contentsOf url: URL, to outputStream: inout Output, applicationRange: SourceRange? = nil
   ) throws {
     guard FileManager.default.isReadableFile(atPath: url.path) else {
       throw SwiftFormatError.fileNotReadable
@@ -58,7 +58,7 @@ public final class SwiftFormatter {
       throw SwiftFormatError.isDirectory
     }
     let sourceFile = try SyntaxParser.parse(url)
-    try format(syntax: sourceFile, assumingFileURL: url, to: &outputStream)
+    try format(syntax: sourceFile, assumingFileURL: url, to: &outputStream, applicationRange: applicationRange)
   }
 
   /// Formats the given Swift source code and writes the result to an output stream.
@@ -89,13 +89,14 @@ public final class SwiftFormatter {
   ///     be written.
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func format<Output: TextOutputStream>(
-    syntax: SourceFileSyntax, assumingFileURL url: URL?, to outputStream: inout Output
+    syntax: SourceFileSyntax, assumingFileURL url: URL?, to outputStream: inout Output,
+    applicationRange: SourceRange? = nil
   ) throws {
     let assumedURL = url ?? URL(fileURLWithPath: "source")
 
     let context = Context(
       configuration: configuration, diagnosticEngine: diagnosticEngine, fileURL: assumedURL,
-      sourceFileSyntax: syntax)
+      sourceFileSyntax: syntax, applicationRange: applicationRange)
     let pipeline = FormatPipeline(context: context)
     let transformedSyntax = pipeline.visit(syntax)
 
--- a/Sources/SwiftFormatCore/Context.swift
+++ b/Sources/SwiftFormatCore/Context.swift
@@ -52,12 +52,18 @@ public class Context {
   /// Contains the rules have been disabled by comments for certain line numbers.
   public let ruleMask: RuleMask
 
+  /// Contains the range of source code to be formatted.
+  ///
+  /// `nil` idicates that the whole file should be formatted
+  public let applicationRange: SourceRange?
+
   /// Creates a new Context with the provided configuration, diagnostic engine, and file URL.
   public init(
     configuration: Configuration,
     diagnosticEngine: DiagnosticEngine?,
     fileURL: URL,
-    sourceFileSyntax: SourceFileSyntax
+    sourceFileSyntax: SourceFileSyntax,
+    applicationRange: SourceRange? = nil
   ) {
     self.configuration = configuration
     self.diagnosticEngine = diagnosticEngine
@@ -69,6 +75,7 @@ public class Context {
       syntaxNode: sourceFileSyntax,
       sourceLocationConverter: sourceLocationConverter
     )
+    self.applicationRange = applicationRange
   }
 
   /// Given a rule's name and the node it is examining, determine if the rule is disabled at this
--- a/Sources/SwiftFormatPrettyPrint/Comment.swift
+++ b/Sources/SwiftFormatPrettyPrint/Comment.swift
@@ -64,7 +64,7 @@ struct Comment {
     }
   }
 
-  func print(indent: [Indent]) -> String {
+  func print() -> String {
     switch self.kind {
     case .line, .docLine:
       let separator = "\n" + kind.prefix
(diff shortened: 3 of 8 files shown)


allevato:
Thanks for getting this started! Apologies for the delay in replying to your latest forum post.

This looks like a reasonable start. However, to call attention to something I mentioned in my latest reply on the thread, I think just inserting the entire `TokenSyntax` into the formatting token stream _won't_ do the correct thing on its own, because we also treat comments as special kinds of formatting tokens in the pretty printer (we extract them from the trivia of a token). So I think you could end up with `comment` tokens in the formatting token stream preceding the `syntax` token whose `TokenSyntax` has leading trivia that also contains those comments, and depending on how you're processing both of those things, you might end up printing those comments twice if you have the pretty printer print `.comment` tokens _and_ print all the trivia of the `TokenSyntax` instance?

So I think instead what we need to do is consider that anywhere that we insert "text" (whether it's syntactic text or a comment), we should also keep track of the whitespace that preceded it. That would mean updating the `.comment` and `.syntax` cases to store that information; when we iterate over the trivia in the `TokenStreamCreator`, we'd have to partition it as we scan it to retain that leading whitespace. It also means we probably have to treat trailing trivia with care, in the handful of cases where it's present.

That may help us make a more informed decision about what whitespace to preserve and what whitespace to discard depending on whether we're inside an application range or not.

cc @dylansturg, who has also been thinking about this problem.

dylansturg:
We need support for range formatting to support a mechanism for toggling formatting on chunks of code (e.g. imagine some chunk of code that the author wants to format exclusively by hand). I'm excited to see good progress on this! 

I experimented with a similar approach to this one, where the entire `TokenSyntax` is inserted into the output stream by the PrettyPrinter. I ran into some difficult edge cases trying to suppress the correct non-syntax text (i.e.comments & newlines). I think that Tony's suggestion of slicing the trivia to store the original whitespace on the comment and syntax tokens could avoid those edge cases, and overall be more reliable to implement.

Regarding the implementation in the PrettyPrinter, I think we should continue processing the tokens to record indentation when in a non-formatted section. That will allow the PrettyPrinter to resume using the correct indentation in the next formatted section.

denis631:
Thank you guys for the comments. 
Now as I've written a test that formats a part of a doc comment I have encountered that comment is written twice. This is exactly the issue you guys have been pointing to.

@allevato it took me a while to get what exactly do you mean, but now I get it. 
This means `.syntax` case contains also leadingTriviaTokens array which corresponds to leading trivia represented as `[Token]` array.
Then in the `processTokenAsInOriginalInput` we need to drop all the leading trivia tokens which were already printed.
Imagine we want to range format 2 out of 3 line doc comment. This means we will have an array of tokens that contains `.comment` cases and a `.syntax(syntaxToken, leadingTriviaTokens)` case. Before printing the text of the syntaxToken we need to print the leading ones, but only the ones which _were not_ already printed

AFAIU this would lead to something like this, where `didPrintToken` function returns true if the token was already printed and else otherwise. How it will be computed is yet to be defined
```swift
  private func processTokenAsInOriginalInput(_ token: Token) {
    if case let .syntax(tokenSyntax, leadingTriviaTokens) = token {
      leadingTriviaTokens
        .drop(while: didPrintToken)
        .forEach(write)

      writeRaw(tokenSyntax.printableText)
      tokenSyntax.trailingTrivia.write(to: &outputBuffer)
    }
  }

  private func didPrintToken(_ token: Token) -> Bool {
    return printedTokens.contains(token)
  }
```
Did I understand it correctly?

> I think we should continue processing the tokens to record indentation when in a non-formatted section. That will allow the PrettyPrinter to resume using the correct indentation in the next formatted section.

@dylansturg this is exactly the problem I am facing right now. What is the "correct indentation"? 
Is there a concrete specification of how the range formatting should work in these kinds of edge cases, like formatting only a part of a statement or a part of an expression.
E.g. what should be the result, if we want to format only the line containing function h (line 3) 
```
func f {
func g {
func h {}
}
}
``` 

should the result be indented by 1 indentation unit, or 2 (since it's indentation stack going to have 2 units, one for `func g` and another one for `func h`?

These kinds of choices obviously have an impact on how range formatting should be implemented. Unfortunately, I didn't find any spec for his.

dylansturg:
> What is the "correct indentation"?
Is there a concrete specification of how the range formatting should work in these kinds of edge cases, like formatting only a part of a statement or a part of an expression.

Unfortunately, I'm not aware of a written spec with these answers. 😞@allevato might have a more formal definition of what we should do when formatting a block of code that is inside of a file. We should certainly consider the behavior of other related formatters (e.g. clang-format) and expected use cases (e.g. formatting a block from a source editor extension).

I'm not sure if anyone else will agree, but I think referencing clang-format could be helpful. We might want to see if it makes sense to match how clang-format handles indentation in and around formatted ranges.

allevato:
> Imagine we want to range format 2 out of 3 line doc comment. This means we will have an array of tokens that contains `.comment` cases and a `.syntax(syntaxToken, leadingTriviaTokens)` case. Before printing the text of the syntaxToken we need to print the leading ones, but only the ones which _were not_ already printed

A couple notes here:

* The `.comment` cases should also keep track of leading whitespace trivia that preceded them.
* Your solution still stores the whole `TokenSyntax` in the `.syntax` case, but I think that overcomplicates matters.

Imagine the following scenario:

```swift
    /// doc comment
    func foo() {
    }
```

IIRC, the leading trivia for the `func` keyword looks roughly like this:

```swift
[
  .newlines(1), .spaces(4), .newlines(1),
  .docComment("/// doc comment"),
  .newlines(1), .spaces(4),
]
```

If we use comment pieces as "pivots" (because those show up in the formatting token stream as "text" we care about), we can "partition" that trivia array around them. So, I think we'd want to represent that in the formatting token stream like this:

```swift
[
  ...,
  .comment("/// doc comment", leadingTrivia: [.newlines(1), .spaces(4), .newlines(1)]),
  .syntax("func", leadingTrivia: [.newlines(1), .spaces(4)]),
  ...,
]
```

In other words, as we scan the trivia of a token looking for its comments, we'd also need to build up those lists.

We'd also need to do something about trailing trivia, in the few cases where it occurs. That might just be another associated value in `.syntax`.

I haven't completely thought this through, but at least if we retained that information into the pretty-printer, we might have what we need to be able to make an informed decision about whether to print what was originally there or to print what the layout algorithm says should be printed, depending on whether we're in a format range or not.

There are probably edge cases all over the place here, like what if your range starts/ends in the middle of a large block comment? In some situations we may just have to ignore the exact range and adjust it to something more sensible.

> Unfortunately, I'm not aware of a written spec with these answers. 😞@allevato might have a more formal definition of what we should do when formatting a block of code that is inside of a file. We should certainly consider the behavior of other related formatters (e.g. clang-format) and expected use cases (e.g. formatting a block from a source editor extension).

@dylansturg is correct, we don't really have a spec in this case. My recommendation would be to look to other formatters as prior art and consider what they do, but to also consider what makes sense within the constraints of the algorithm we have implemented.

denis631:
@allevato thank you for your detailed description.
The only piece that is missing from the puzzle is how do other `Token`s fit into this design (e.g. `.open` and `.close` cases). Are they also in the leading trivia part of the `.comment` or `.syntax` cases or do they exist separately in the token stream.
In the example above what we will have is something like this:
```swift
[
  ...,
  .comment("/// doc comment", leadingTrivia: [.newlines(1), .spaces(4), .newlines(1)]),
  .open, 
  .open,
  .open,
  .syntax("func", leadingTrivia: [.newlines(1), .spaces(4)]),
  ...,
]
```
Should these open cases be in `.syntax` case as leading trivia as well?
What about the `.spaces` case. Sometimes it is in the leading trivia but is ignored by the tokenStreamCreator, yet now it should be in the leadingTrivia and if we put everything the leadingTrivia associated object of `.syntax` and `.comment` cases then how we do know when to ingore the cases and when not to?

rabingaire:
Hey @denis631 are you still working on this? I was trying to implement this feature in SourceKit-LSP and I saw your pull request on swift-format. Let me know if you are still working on this or not?

@allevato

denis631:
@rabingaire no I am not atm.
I hope you can benefit from the discussion, but my question written in the previous comment is still opened 🤷‍♂

shahmishal:

PR #54, opened 2019-09-13
Expression grouping improvements to relocate breaks.

dylansturg:
Some types of expressions were breaking at suboptimal locations in the source as a result of a lack of groups. I experimented with grouping around every expression in a SequenceExpr, but there are so many kinds of expressions used in so many contexts that it's difficult to determine appropriate groups in general. A second complicating factor is that the precedence of operators cannot be determined using the parsed syntax. Without knowing operator precedence, it's difficult to correctly group operators in general.

This change uses the following assumptions to make some progress on expression grouping:
1. Logical AND and OR operators (&&, ||) are generally good candidates for breaks and their operands are generally best grouped together.
2. Choices in ternary expressions are generally best grouped together, so that the semantic tokens in the ternary ("?", ":") are easier to find.
3. Member access expressions and function call expressions generally look best grouped together, allowing the breaks to occur at other expressions.

I have added a few new test cases that exercise the revised formatting behaviors, and updated the few existing test cases that were failing to match the new formatting.


Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
```diff
@@ -283,4 +283,26 @@ public class IfStmtTests: PrettyPrintTestCase {
 
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 50)
   }
+
+  public func testConditionExpressionOperatorGrouping() {
+    let input =
+      """
+      if someObj is SuperVerboselyNamedType || someObj is AnotherPrettyLongType  || someObjc == "APlainString" || someObj == 4 {
```
allevato: Can you also add an example where one of the interior expressions (between two `||`s) is so long that it needs to wrap inside? Since you used `.continue` breaks, my hunch is that they won't stack, and you'll end up with something like this:

```swift
if something is SomeType
  || someOtherThing + someDifferentValue
  / someDenominator == 72
  || someLastCondition
{
```

which makes me think we might want to use `.open/.close` breaks around our `&&/||` groups instead of `.continue` breaks. Does that end up working correctly?

Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
```diff
@@ -283,4 +283,26 @@ public class IfStmtTests: PrettyPrintTestCase {
...
+        // do something
+      }
+      """
+
+    let expected =
+      """
+      if someObj is SuperVerboselyNamedType
+        || someObj is AnotherPrettyLongType
+        || someObjc == "APlainString" || someObj == 4
```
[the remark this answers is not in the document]
dylansturg: I like the idea. I tend to dislike bin packing of conditional expressions in particular, because it's easier to miss expressions. The token between the `if` and the first condition is already a space, so I think this might be straightforward.

allevato:
Now that #58 is merged, can you rebase these changes on top of it to sanitize the diff?

Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
dylansturg: It ended up being a little more involved, but I've got the intra-expression continuation indent working.

Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
```diff
@@ -275,10 +278,18 @@ public class PrettyPrinter {
...
         // scope.
-        activeOpenBreaks.append(
-          ActiveOpenBreak(
-            lineNumber: currentLineNumber,
-            contributesContinuationIndent: currentLineIsContinuation))
+        var activeOpenBreak = ActiveOpenBreak(
+          kind: openKind,
+          lineNumber: currentLineNumber,
+          contributesContinuationIndent: currentLineIsContinuation)
+        activeOpenBreak.contributesBlockIndent = openKind == .block
```
allevato: Let's go ahead and make this an initializer argument too, since it's always set immediately after creating the instance.

Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
```diff
@@ -275,10 +278,18 @@ public class PrettyPrinter {
...
+
+        // Check if this break will fire, and apply a continuation indent to its contents.
```
allevato: Can this section (up until the `append` call) be moved above where the `ActiveOpenBreak` is created, so that instead of reassigning a value to `contributesContinuationIndent`, we start a temporary variable (perhaps also named `contributesContinuationIndent`) equal to `currentLineIsContinuation`, then update it based on this, and pass that into the `ActiveOpenBreak` initializer? That, along with the change above, would remove all the state changes in this section of the code, so it may be easier to reason about.

Tests/SwiftFormatPrettyPrintTests/ForInStmtTests.swift
```diff
@@ -186,4 +199,67 @@ public class ForInStmtTests: PrettyPrintTestCase {
 
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 20)
   }
+
+  public func testForStatementWithNestedExpressions() {
```
allevato: Niiiiiiiiiiice 😱

PR #55, opened 2019-09-16
Properly format VariableDecls with multiple bindings.

allevato:


allevato:
cc @dylansturg

Tests/SwiftFormatPrettyPrintTests/VariableDeclTests.swift
```diff
@@ -46,4 +46,47 @@ public class VariableDeclarationTests: PrettyPrintTestCase {
...
+      let a = letsForceTheFirstOneToWrapAsWell, longVariableName = longFunctionCall()
+      """
+
+    let expected =
+      """
+      let a = 100, b = 200, c = 300, d = 400,
+        e = 500, f = 600
+      let a = 5,
+        longVariableName = longFunctionCall(),
+        anotherReallyLongVariableName
```
dylansturg: Would there be any value in adding an example where the indentation goes back one after a continuation in the declaration?

For example:
```
let a = 5,
  longVariableName = longFunctionCall(),
  anotherReallyLongVariableName
    = something,
  someFinalVar = 6
```

allevato: Good call. I've just swapped the two bindings in this example since the effect is the same.

PR #56, opened 2019-09-17
Improve multiline string wrapping.

allevato:
This fixes a couple oddities in the way multiline strings were being
handled:

* Previously, the entire text of the multiline string literal was
  inserted as a single syntax node. This meant that the opening
  quotes (and tokens preceding them up to the closest break) would
  wrap unnecessarily if the string's *total* length was longer than
  the column limit, even if the opening line wasn't.

  Now, we treat each line of the string literal individually, so
  the opening quotes only wrap if they would be the only thing that
  didn't fit on the line.

* The leading whitespace was also preserved before, so we made no
  attempt at reindenting the multiline strings according to what the
  pretty printer was doing. Now we compensate for that so the pretty
  printer ensures that multiline strings line up with other tokens
  based on the current indentation.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -26,6 +26,11 @@ private final class TokenStreamCreator: SyntaxVisitor {
   private let config: Configuration
   private let maxlinelength: Int
 
+  /// Keeps track of the prefix length of multiline string segments when they are visited so that
+  /// the prefix can be stripped at the beginning of lines before the text is added to the token
+  /// stream.
+  private var pendingMultilineStringSegmentPrefixLengths = [TokenSyntax: Int]()
+
   init(configuration: Configuration) {
     self.config = configuration
     self.maxlinelength = config.lineLength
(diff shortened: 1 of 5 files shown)


allevato:
cc @dylansturg

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1550,20 +1574,57 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
+      appendToken(.syntax(String(firstLine)))
+    }
+
+    // Add the remaining lines of the segment, preceding each with a newline and stripping the
+    // leading whitespace so that the pretty-printer can re-indent the string according to the
+    // standard rules that it would apply.
+    for line in lines.dropFirst() as ArraySlice {
+      appendToken(.newline)
+
+      let remainingString = line.dropFirst(prefixCount)
```
dylansturg: Could this be destructive to the string if it was malformed by not having sufficient leading spaces on some line? I'm pretty sure that would **not** compile, so that may not be a worthwhile consideration. It looks like the syntax parser is able to parse such a malformed string into a StringLiteralExpr.

Say if you wrote something like this, and then formatted it (even though it doesn't compile), maybe because you format on save:
```
  let someString = """
my first good string
  """
=>
  let someString = """
   first good string
  """
```

allevato: Thanks for catching that; this stage of the parser is more forgiving than I was expecting.

I've updated this so that lines that contain fewer than the expected number of spaces only strip off the spaces that are there; this forces those lines to align, which I think is a nice "autocorrection" vs. bailing out entirely and leaving the string untouched.

I also realized there were errors in blank line handling because the use of `.newline` tokens was triggering the folding logic when they exceeded the maximum number allowed by the configuration, so the latest version I pushed also includes some necessary changes to the `.newline` logic to support newlines that should always _always_ **_always_** be there.

PR #57, opened 2019-09-17
Support subscripts in AddModifierRewriter.

allevato:
This fixes the `NoAccessLevelOnExtensionDeclaration` rule, where an access level on an `extension` decl was not being moved to subscripts inside it.

This change also adds some test coverage for `typealias`es while I was in here.


allevato:
cc @dylansturg

shali3:
I'm not sure this is the right place to aske, but can you please elaborate why setting an access identifier on an extension is wrong to begin with?
I couldn't find an answer in the comments/docs.

PR #63, opened 2019-09-30
Add a SequenceExpr folding algorithm.

allevato:
This change adapts an algorithm used by the compiler to fold
sequence expressions into expression trees based on the precedence
and associativity of operators.

We will use this in TokenStreamCreator to transform SequenceExprs
into shapes that let us more easily insert groups and breaks that
allow wrapping to occur in a manner consistent with precedence and
associativity, which matches users' expectations.


Sources/SwiftFormatPrettyPrint/OperatorContext.swift
```diff
@@ -0,0 +1,368 @@
...
+  private var infixOperators = [String: InfixOperator]()
+
+  /// A cache of associativities between pairs of operators, computed and cached
+  /// on demand.
+  private var associativityCache = [AssociativityCacheKey: Associativity?]()
+
+  /// Creats a new operator context that is populated with the precedence groups
+  /// operators that are built into the Swift compiler and standard library.
+  public init() {
+    // Add the built-in precedence groups.
```
akyrtzi: Nitpick, I'd recommend to decouple a bit `OperatorContext` from how to populate it. So for example there could be an `addDefaultOperators` function and for convenience a `getDefaultOperatorContext` static function. Later on we could potentially have a function to populate an `OperatorContext` by loading a module (e.g. the stdlib) and for this case not need the hardcoded operators.

Sources/SwiftFormatPrettyPrint/SequenceExprFolding.swift
```diff
@@ -0,0 +1,490 @@
...
+  /// Once folded, the returned expression will have one of three possible
+  /// shapes:
+  ///
+  /// - A subtype of `ExprSyntax` other than `SequenceExprSyntax`
+  /// - A `SequenceExprSyntax` with exactly two elements, where the first is an
+  ///   operand (a subtype of `ExprSyntax`) and the second is a cast expression
+  ///   (either `AsExprSyntax` or `IsExprSyntax`)
+  /// - A `SequenceExprSyntax` with exactly three elements, where the first and
+  ///   third elements are operands (subtypes of `ExprSyntax`) and the second is
+  ///   a `BinaryOperatorExprSyntax`
```
akyrtzi: As part of moving this to SwiftSyntax I'd recommend that we add new syntax nodes (like `BinaryExprSyntax`) that completely replace `SequenceExprSyntax` from the tree. `SequenceExprSyntax` should not be present in a folded tree then (unless there was an unknown operator).

Tests/SwiftFormatPrettyPrintTests/SequenceExprFoldingTests.swift
```diff
@@ -0,0 +1,204 @@
+import SwiftFormatPrettyPrint
+import SwiftSyntax
+import XCTest
+
+final class SequenceExprFoldingTests: XCTestCase {
```
akyrtzi: Could you also add a test demonstrating and testing how to add a custom operator programmatically and have it folded properly?

Sources/SwiftFormatPrettyPrint/OperatorContext.swift
akyrtzi: `Builtin` as in `addBuiltinOperators` is slightly better.

Sources/SwiftFormatPrettyPrint/SequenceExprFolding.swift
allevato: 👍 to this idea; for now, I've added a TODO to track doing this so that we can move forward with other changes that depend on this.

Tests/SwiftFormatPrettyPrintTests/SequenceExprFoldingTests.swift
allevato: Done.

Sources/SwiftFormatPrettyPrint/OperatorContext.swift
allevato: Done.

PR #71, opened 2019-10-29
Don't group around function calls that occur after an assignment oper…

dylansturg:
…ator.

Grouping around these function calls results in suboptimal formatting for expressions that simply assign the result of a function call. There was also inconsistent between assignments and declarations (e.g. using let or var), because there may be SequenceExpr when using let/var decl.

I opted to also exclude the grouping when the function call is the first item in an expression list, because the indentation of following items isn't confusing whereas the indentation would be confusing when the function call follows other operands.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -669,6 +669,13 @@ private final class TokenStreamCreator: SyntaxVisitor {
   }
 
   func visit(_ node: FunctionCallExprSyntax) -> SyntaxVisitorContinueKind {
+    if let calledMemberAccessExpr = node.calledExpression as? MemberAccessExprSyntax {
+      if let base = calledMemberAccessExpr.base, base is IdentifierExprSyntax {
+        before(base.firstToken, tokens: .open)
+        after(calledMemberAccessExpr.name.lastToken, tokens: .close)
+      }
+    }
+
     let arguments = node.argumentList
     if !arguments.isEmpty {
       // If there is a trailing closure, force the right parenthesis down to the next line so it
@@ -1219,7 +1226,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
       // assignment expressions) are covered by 3-element sequence expressions.
       let binOp = iterator.next()!
       let rhs = iterator.next()!
-      maybeGroupAroundSubexpression(rhs)
+      maybeGroupAroundSubexpression(rhs, combiningOperator: binOp)
 
       if shouldRequireWhitespace(around: binOp) {
         if shouldStackIndentation(after: binOp) {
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -2274,14 +2281,46 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
-    case is FunctionCallExprSyntax, is MemberAccessExprSyntax, is SubscriptExprSyntax:
+    case is MemberAccessExprSyntax, is SubscriptExprSyntax:
       before(expr.firstToken, tokens: .open)
       after(expr.lastToken, tokens: .close)
     default:
       break
     }
+
+    // Function calls in expressions are conditionally grouped based on whether they are preceded by
+    // an assigning operator. Assigning the result of a function call is allowed to break the
```
allevato: Instead of using the term "rectangle rule" here (since we don't define it anywhere, and the only references I can find online are in a couple of Google's public Java formatter references), let's replace the sentence starting with "Assigning the result of..." with:

> When a function call expression is assigned to an lvalue, we omit the group around the function call so that the callee and open parenthesis can remain on the same line, if they fit. This is a frequent enough case that the outcome looks better with the exception in place.

At some point we may want to formalize our use of the rectangle rule throughout the pretty printer, but for now, let's avoid it as "jargon."

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -2274,14 +2281,46 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
+  /// Assignment is defined as either being an assignment operator (i.e. `=`) or any operator that
+  /// uses "assignment" precedence.
+  private func isAssigningOperator(_ operatorExpr: ExprSyntax) -> Bool {
+    if operatorExpr is AssignmentExprSyntax {
+      return true
+    }
+    if let binaryOperator = operatorExpr as? BinaryOperatorExprSyntax {
+      let operatorText = binaryOperator.operatorToken.text
+      if let precedence = operatorContext.infixOperator(named: operatorText)?.precedenceGroup,
+        precedence === operatorContext.precedenceGroup(named: .assignment)
```
p4checo: Hi guys 👋 

Sorry to bump into this PR out of the blue, but I'm currently adding `swift-format` to our project and this was one of the "problems" we found with the current implementation, so thanks for tackling it!

Would it be reasonable to extend this rule to not just apply to `AssignmentPrecedence` operators, but to any operator with precedence *equal or higher* than `AssignmentPrecedence`?

In our particular case, we use ReactiveSwift which adds the custom operator `<~` (binding operator) defined as such:

```swift
precedencegroup BindingPrecedence {
	associativity: right

	// Binds tighter than assignment but looser than everything else
	higherThan: AssignmentPrecedence
}
```

It's essentially used whenever an assignment is made (hence the precedence relationship), and it manifests the same "wrong" formatting behavior by indenting expression into the next line if it doesn't fit (which happens most of the time since we're dealing with complex reactive chains).

What do you say? Thanks! 🙏

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: Good point, I forgot about the fact that rectangle rule isn't necessarily widely known.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: Just for context - you can see what operators we currently support [here](https://github.com/apple/swift-format/blob/master/Sources/SwiftFormatPrettyPrint/OperatorContext.swift).

If you're interested, we'd be happy to review a PR that adds support for reading custom operators in from the formatter's configuration!

p4checo: Thanks for the feedback! 🙇 

I'm still **very** unfamiliar with `swift-format`'s codebase, so pardon my ignorance 😅. What you mention makes complete sense. I noticed some types representing operator precedences (e.g. `.assignment`), but didn't dive into the code to see if it was effectively doable or made sense. Guess I was being too optimistic!

However, I do intend to invest some time in `swift-format` as I want to try and make some rules more configurable as well as add support for some new ones, so we can automatically enforce as much of our style guide as possible. This change will definitely be the list of stuff to do, but I think I'll start with some "easier" additions first 😇.

Keep it up!

PR #72, opened 2019-11-03
Add new rule to add single line after function decls

p4checo:
## Motivation

To aid in visual processing and add clarity between function declaration and body, a newline is sometimes useful separating the two.

To enforce that, a new rule `BlankLineAfterFunctionDeclarations` was created, allowing two configurable exceptions:

 - single line functions, as these expressions are already compact and quite readable, so leaving them as is doesn't hinder visual processing.

 - super calls (when on first line), as these can be seen as being very "coupled" to the function declaration itself so it might make sense to keep this coupling without visual separation.

Some helpers were created to help adjust indentation while formatting:

- `Trivia.withSpacesOnly()` to return trivia with whitespace only,

- `Indent.asTrivia` to return a `Trivia` representation of an indent.

## Discussion

This is the first of some rules/configurations I intend to add to `swift-format`, with the goal of eventually using it to apply our own [style guide](https://github.com/Mindera/swift-style-guide) automatically, while also enriching current `swift-format`'s capabilities.

However, I understand that some care must be taken before adding new rules, so that only rules that make sense for the broad community are added. From my perspective, rules should be generic while also being configurable to minimize total number of rules, avoiding having rules that are too specific.

That being said, not sure what's the "process" or "requirements" for adding new rules are (if any), and if my approach was the best one, as I am completely new to the codebase. I basically took some inspiration from `BlankLineBetweenMembers` and went from there, so please be gentle 😇.

### Target branch and tooling notes

I developed this rule under `swift-5.1-branch`, as I was unable to get `swift run generate-pipeline` working with the development snapshot. Even on Xcode 11  with the `swift-5.1-branch` branch, I had to resort to `swift package generate-xcodeproj` as I couldn't run the tests in the "native" Xcode SPM integration due to some linker errors. Everything should work on `master` though.

## 🚧 WIP - missing work 🚧

In the meantime I realized my rule isn't yet applying to all function-like declarations, like `init`'s, `deinit`'s and `subscript`'s. Will do once my approach is validated to avoid throwing away work.


Documentation/Configuration.md
```diff
@@ -82,6 +82,10 @@ An example `.swift-format` configuration file is shown below.
     "blankLineBetweenMembers": {
         "ignoreSingleLineProperties": true
     },
+    "blankLineAfterFunctionDeclarations": {
```
allevato: Technically, the "function declaration" includes everything from the first token (arguably even the preceding doc comment) up through the body and the closing brace. So I was a bit confused when I first read this description, because to me, "after function declaration" would imply "after the final `}`", rather than "after the opening `{` that follows the function signature".

A clearer name for this might be `blankLineAtBeginningOfFunctionBody`.

p4checo: I agree. Still getting used to all this new syntax _lingo_ 😅

p4checo:
> I'll start by saying that this rule is one that is a bit, well, opinionated 😄

I agree 😄. But in a way isn't code formatting in general _mostly_ opinionated? 😂

> At the same time, the set of core rules that we provide should be generally useful to a fairly large audience. If we allow too many rules or knobs for configuration, the code base will quickly grow to be difficult to maintain/test reasonably.

Completely agree, and I know this is not an easy balance to achieve.

> Looking at this rule and the specificity of its behavior (e.g., treating super differently), I'm not sure this is general enough to take on as a core rule. (And in complete fairness, I also want to look at our own existing rule set and prune some that may not have proved as useful as originally intended, or at least make them optional.)

Yeah, we might be pushing it with the `super` exception. I added it because it's a part of our style guide, and also as an exercise (in fact implementing this entire new rule was a great exercise to get to know the codebase and formatting mechanics). However, I think we might be able to drop this `super` exception if the remaining behavior is considered "valid" for the bigger audience.

> That parenthetical statement leads to another issue, which is that right now, we don't have a way to make the rewriter rules disabled by default;
> (...)
> Having a rule like this be enabled by default would be a breaking change, since it would force that behavior on all existing users unless they knew to update their configuration after this went in.

Can't we in a way ensure we don't add any side effects by ensuring the default rule value doesn't modify existing behavior? At least that was my approach/interpretation.

> Making some rules opt-in would be straightforward, but I think I'd like to strategize a bit more about other interesting ways that third-parties could hook into the formatter without forking it or adding rules to the core set. For example, maybe we could provide a way that users could build their own dylibs that exported custom rules, and then have swift-format detect and load those into the pipeline.

I do agree that some sort of dynamic plugin/rule system would be ideal so that any 3rd party could develop their own custom rules freely and use them on their own formatting pipeline. 

In the meantime (and pardon me if this is a stupid idea), would it be easily feasible to use something along the lines of the `XCTestManifests.swift` for Linux? In the sense that we could have some configuration file that pointed to some custom rules `.swift` files, that were then built into the `swift-format` binary. I don't see that big of a overhead in building the binary every time we want to include a new custom rule, as it's not something you do very often, and you can store/distribute the binary easily. Just a thought, anyway.

Given we don't have anything like that for now, do you suggest 3rd parties keep their custom rules in their own forks?

shahmishal:

PR #73, opened 2019-11-04
Add new configuration do add a line break before each generic requirement

p4checo:
## Motivation

When generic requirement lists are wrapped into multiple lines, it might be desirable to place a line break before each requirement, in the same way generic arguments and function arguments are with `lineBreakBeforeEachArgument`.

When enabled, this new configuration will add a new line before each generic requirement, causing all requirements to line up vertically after the `where` clause, with an extra level of indentation. If
disabled, requirements will be laid out horizontally first, with line breaks only being fired when the line length would be exceeded. This also helps enforce the rectangle rule for the entire generic requirement lists in both scenarios.

## Changes

- Add new `lineBreakAfterGenericWhereClause` configuration that is applied on `TokenStreamCreator.visit(_ node: GenericWhereClauseSyntax)`.

- Add new unit tests to validate desired behavior.

- Fix some typos.


Documentation/Configuration.md
```diff
@@ -58,6 +58,14 @@ top-level keys and values:
     argument, forcing the entire argument list to be laid out vertically.
     If false (the default), arguments will be laid out horizontally first, with
     line breaks only being fired when the line length would be exceeded.
+    
+*   `lineBreakAfterGenericWhereClause` _(boolean)_:  Determines the
```
allevato: So the behavior here is very close in spirit to what we do just above for function arguments, but with one difference: for function arguments, the rule we apply is "if a list of function/generic arguments must wrap, then a line break is added before _every_ argument". The rule this PR applies is "if a list of generic requirements must wrap, then a line break is added before _the first_ requirement".

Fundamentally, these are both really just "comma-delimited lists inside decls", so I think it would make sense to treat them uniformly. How would you feel about renaming this to `lineBreakBeforeEachGenericRequirement` and implementing that instead? The change in implementation would be minor; I believe you'd only need to pass `.consistent` as the group style to the `.open` group that you create below.

Tests/SwiftFormatPrettyPrintTests/ClassDeclTests.swift
```diff
@@ -126,7 +126,7 @@ public class ClassDeclTests: PrettyPrintTestCase {
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 30, configuration: config)
   }
 
-  public func testClassInheritence() {
+  public func testClassInheritance() {
```
allevato: Thanks for cleaning up all these typos that I missed (or maybe even added)! 😄

Documentation/Configuration.md
p4checo: That's a good point. I thought about applying this to each requirement (effectively replicating the function arguments) but then decided to only implement it for the first one so that very long generic requirement lists don't end up taking lots of vertical space. Additionally, even though they are both "comma delimited lists inside decls", I see parameters as being in a way more "independent" than generic requirements, so it makes more sense to split them each on its own when broken. On the other side, if we were talking about _generic parameters_,  I would agree with you (and in fact I intend to tackle that one too). Does it make sense?

That being said, if possible I would prefer to keep this behavior of just adding a line break before the first requirement when broken into multiple lines. Perhaps we could convert this to an enum like the one below? (the actual names might suck, but the ideia is there 😇)

```swift
enum GenericRequirementListLineBreakStyle {
    case noExplicitBreak // .none would be better, but we should avoid it because of Optional
    case beforeFirst
    case beforeEach
```

allevato: Sorry for the delay on coming back around to some of these—I've been putting out fires on other projects.

After taking a little bit longer to think about this, I'd like to propose something different, because I realized that what we're doing today is kind of _wrong._ In other words, this (see ClassDeclTests.swift, line 197):

```swift
class MyClass<S, T>
where S: Collection, T: ReallyLongClassName,
  U: LongerClassName
{
  // ...
}
```

...shouldn't be allowed, _even if we're binpacking the requirements list._ (It also violates our internal style guide. Oops!)

When we talk about style rules in our internal discussions, we refer to a concept called [the rectangle rule](https://github.com/google/google-java-format/wiki/The-Rectangle-Rule) to describe how the content of nodes should be able to be "boxed" for clarity, and everything after the `where` clause fails to meet that. We allow the rectangle rule to be broken in some places where an extra line break would be wasteful, but I don't think this is one of those places.

So, what I'd like to see happen is this: we _always_ put a group around the generic requirements so that we break before them if they're ever long enough that they don't fit on the same line following the `where` keyword. Then, we call this setting `lineBreakBeforeEachGenericRequirement` and implement it as such, making it consistent with arguments (when it's true, it lays out vertically; when it's false, it bin-packs).

This avoids overly complicating the setting with too many knobs, and also ends up getting you the behavior you're looking for.

@dylansturg, would you agree with this assessment?

dylansturg: Grouping around the generic requirements sounds good to me. That's less complexity, and it has a nice symmetry with function arguments (where we also force a break after the opening `(` if the arguments must use multiple lines).

p4checo: This is just brilliant! Implementing it like that makes complete sense and both rule "scenarios" end up looking better! The rectangle rule does make sense after all 😄

Tests/SwiftFormatPrettyPrintTests/PrettyPrintTestCase.swift
```diff
@@ -19,20 +19,69 @@ public class PrettyPrintTestCase: XCTestCase {
 
     // Assert that the input, when formatted, is what we expected.
     if let formatted = prettyPrintedSource(input, configuration: configuration) {
-      XCTAssertEqual(
-        expected, formatted,
-        "Pretty-printed result was not what was expected",
-        file: file, line: line)
+      XCTAssertDiff(
```
allevato: Let's revert the changes in this file before merging so that we're not duplicating this code in two places. We should move common code into a separate test-support module, and I'd also like to replace the current diffing logic (which was a quickly written temporary solution) with something more general purpose, but none of that is relevant to the change being made by this PR so it would be better to keep this change localized.

p4checo: That makes sense. I added it in because it's so much easier to spot what's wrong with a diff 😅

PR #74, opened 2019-11-06
Add new configuration do prioritize keeping function output together

p4checo:
## Motivation

When a function's signature doesn't fit a single line because of the output (e.g. `throws -> ResultType`), the default behavior is to break right before the output and indent it one extra level.

However, this might not be the desired visual outcome, and by prioritizing keeping the output together (i.e. `) throws -> ReturnType`), the pretty printer will first break on any generic parameters and parameters before breaking the output into its own line.

## Changes

- Add new `prioritizeKeepingFunctionOutputTogether` configuration that is applied on `TokenStreamCreator`'s `visit(_ node: FunctionDeclSyntax)` and `visit(_ node: SubscriptDeclSyntax)`.

- Add new unit tests to validate desired behavior.

### DIFF
--- a/Documentation/Configuration.md
+++ b/Documentation/Configuration.md
@@ -58,6 +58,16 @@ top-level keys and values:
     argument, forcing the entire argument list to be laid out vertically.
     If false (the default), arguments will be laid out horizontally first, with
     line breaks only being fired when the line length would be exceeded.
+    
+*   `prioritizeKeepingFunctionOutputTogether` _(boolean)_: Determines if 
+    function-like declaration outputs should be prioritized to be together with the
+    function signature right (closing) parenthesis. If false (the default), function 
+    output (i.e. throws, return type) is not prioritized to be together with the 
+    signature's right parenthesis, and when the line length would be exceeded,
+    a line break will be fired after the function signature first, indenting the 
+    declaration output one additional level. If true, A line break will be fired 
+    further up in the function's declaration (e.g. generic parameters, 
+    parameters) before breaking on the function's output.  
 
 *   `indentConditionalCompilationBlocks` _(boolean)_: Determines if
     conditional compilation blocks are indented. If this setting is `false` the body
--- a/Sources/SwiftFormatConfiguration/Configuration.swift
+++ b/Sources/SwiftFormatConfiguration/Configuration.swift
@@ -29,6 +29,7 @@ public class Configuration: Codable {
     case blankLineBetweenMembers
     case lineBreakBeforeControlFlowKeywords
     case lineBreakBeforeEachArgument
+    case prioritizeKeepingFunctionOutputTogether
     case indentConditionalCompilationBlocks
     case rules
   }
@@ -88,6 +89,16 @@ public class Configuration: Codable {
   /// each argument, forcing the entire argument list to be laid out vertically.
   public var lineBreakBeforeEachArgument = false
 
+  /// Determines if function-like declaration outputs should be prioritized to be together with the
+  /// function signature right (closing) parenthesis.
+  ///
+  /// If false (the default), function output (i.e. throws, return type) is not prioritized to be
+  /// together with the signature's right parenthesis, and when the line length would be exceeded,
+  /// a line break will be fired after the function signature first, indenting the declaration output
+  /// one additional level. If true, A line break will be fired further up in the function's
+  /// declaration (e.g. generic parameters, parameters) before breaking on the function's output.
+  public var prioritizeKeepingFunctionOutputTogether = false
+
   /// Determines the indentation behavior for `#if`, `#elseif`, and `#else`.
   public var indentConditionalCompilationBlocks = true
 
@@ -129,10 +140,11 @@ public class Configuration: Codable {
       BlankLineBetweenMembersConfiguration.self, forKey: .blankLineBetweenMembers)
       ?? BlankLineBetweenMembersConfiguration()
     self.lineBreakBeforeControlFlowKeywords
-      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeControlFlowKeywords)
-      ?? true
+      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeControlFlowKeywords) ?? false
     self.lineBreakBeforeEachArgument
-      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeEachArgument) ?? true
+      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeEachArgument) ?? false
+    self.prioritizeKeepingFunctionOutputTogether
+      = try container.decodeIfPresent(Bool.self, forKey: .prioritizeKeepingFunctionOutputTogether) ?? false
     self.indentConditionalCompilationBlocks
       = try container.decodeIfPresent(Bool.self, forKey: .indentConditionalCompilationBlocks) ?? true
     self.rules = try container.decodeIfPresent([String: Bool].self, forKey: .rules) ?? [:]
@@ -148,9 +160,9 @@ public class Configuration: Codable {
     try container.encode(indentation, forKey: .indentation)
     try container.encode(respectsExistingLineBreaks, forKey: .respectsExistingLineBreaks)
     try container.encode(blankLineBetweenMembers, forKey: .blankLineBetweenMembers)
-    try container.encode(
-      lineBreakBeforeControlFlowKeywords, forKey: .lineBreakBeforeControlFlowKeywords)
+    try container.encode(lineBreakBeforeControlFlowKeywords, forKey: .lineBreakBeforeControlFlowKeywords)
     try container.encode(lineBreakBeforeEachArgument, forKey: .lineBreakBeforeEachArgument)
+    try container.encode(prioritizeKeepingFunctionOutputTogether, forKey: .prioritizeKeepingFunctionOutputTogether)
     try container.encode(indentConditionalCompilationBlocks, forKey: .indentConditionalCompilationBlocks)
     try container.encode(rules, forKey: .rules)
   }
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -67,14 +67,6 @@ private final class TokenStreamCreator: SyntaxVisitor {
   /// token stream.
   func before(_ token: TokenSyntax?, tokens: [Token]) {
     guard let tok = token else { return }
-    for preToken in tokens {
-      if case .open = preToken {
-        openings += 1
-      } else if case .close = preToken {
-        assert(openings > 0)
-        openings -= 1
-      }
-    }
     beforeMap[tok, default: []] += tokens
   }
 
@@ -88,14 +80,6 @@ private final class TokenStreamCreator: SyntaxVisitor {
   /// token stream.
   func after(_ token: TokenSyntax?, tokens: [Token]) {
     guard let tok = token else { return }
-    for postToken in tokens {
-      if case .open = postToken {
-        openings += 1
-      } else if case .close = postToken {
-        assert(openings > 0)
-        openings -= 1
-      }
-    }
     afterMap[tok, default: []].append(tokens)
   }
 
@@ -268,6 +252,12 @@ private final class TokenStreamCreator: SyntaxVisitor {
     after(node.funcKeyword, tokens: .break)
     after(node.identifier, tokens: .close)
 
+    // Prioritize keeping ") throws -> <return_type>" together.
+    if config.prioritizeKeepingFunctionOutputTogether {
+      // Due to visitation order, the matching .open break is added in ParameterClauseSyntax.
+      after(node.signature.lastToken, tokens: .close)
+    }
+
     if case .spacedBinaryOperator = node.identifier.tokenKind {
       after(node.identifier.lastToken, tokens: .space)
     }
@@ -320,6 +310,12 @@ private final class TokenStreamCreator: SyntaxVisitor {
     if let firstModifierToken = node.modifiers?.firstToken {
       before(firstModifierToken, tokens: .open)
       after(node.subscriptKeyword, tokens: .close)
+    }
+
+    // Prioritize keeping ") -> <return_type>" together.
+    if config.prioritizeKeepingFunctionOutputTogether {
+      // Due to visitation order, the matching .open break is added in ParameterClauseSyntax.
+      after(node.result.lastToken, tokens: .close)
     }
 
     arrangeAttributeList(node.attributes)
@@ -849,6 +845,14 @@ private final class TokenStreamCreator: SyntaxVisitor {
   func visit(_ node: ParameterClauseSyntax) -> SyntaxVisitorContinueKind {
     after(node.leftParen, tokens: .break(.open, size: 0), .open(argumentListConsistency()))
     before(node.rightParen, tokens: .break(.close, size: 0), .close)
+
+    // Prioritize keeping ") throws -> <return_type>" together.
+    if config.prioritizeKeepingFunctionOutputTogether {
+      // Due to visitation order, this .open corresponds to a .close added in FunctionDeclSyntax
+      // or SubscriptDeclSyntax.
+      before(node.rightParen, tokens: .open)
+    }
+
     return .visitChildren
   }
 
--- a/Tests/SwiftFormatPrettyPrintTests/FunctionDeclTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/FunctionDeclTests.swift
@@ -552,6 +552,53 @@ public class FunctionDeclTests: PrettyPrintTestCase {
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 33)
   }
 
+  public func testBreaksBeforeOrInsideOutput_prioritizingKeepingOutputTogether() {
+    let input =
+    """
+      func name<R>(_ x: Int) throws -> R
+
+      func name<R>(_ x: Int) throws -> R {
+        statement
+        statement
+      }
+      """
+
+    var expected =
+    """
+      func name<R>(
+        _ x: Int
+      ) throws -> R
+
+      func name<R>(
+        _ x: Int
+      ) throws -> R {
+        statement
+        statement
+      }
+
+      """
+    let config = Configuration()
+    config.prioritizeKeepingFunctionOutputTogether = true
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 23, configuration: config)
+
+    expected =
+    """
+    func name<R>(
+      _ x: Int
+    ) throws -> R
+
+    func name<R>(
+      _ x: Int
+    ) throws -> R {
+      statement
+      statement
+    }
+
+    """
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 30, configuration: config)
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 33, configuration: config)
+  }
+
   public func testBreaksBeforeOrInsideOutputWithAttributes() {
     let input =
       """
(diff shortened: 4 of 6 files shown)


Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -88,6 +88,15 @@ public class Configuration: Codable {
...
+  /// Determines if function-like declaration outputs should be prioritized to be together with the
+  /// function signature right (closing) parenthesis.
+  ///
+  /// If false (the default), function output (i.e. throws, return type) is not prioritized to be
+  /// together with the signature's right parenthesis, and when the line length would be exceeded,
+  /// a line break will be fired after the function signature first, indenting the declaration output
+  /// one additional level. If true, A line break will be fired further up in the function's
+  /// declaration (e.g. generic parameters, parameters) before breaking on the function's output.
+  public var prioritizeKeepingFunctionOutputTogether = false
   /// Determines the indentation behavior for `#if`, `#elseif`, and `#else`.
```
allevato: Nit: Add a blank line after this property.

Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -88,6 +88,15 @@ public class Configuration: Codable {
...
 
+  /// Determines if function-like declaration outputs should be prioritized to be together with the
+  /// function signature right (closing) parenthesis.
+  ///
+  /// If false (the default), function output (i.e. throws, return type) is not prioritized to be
+  /// together with the signature's right parenthesis, and when the line length would be exceeded,
+  /// a line break will be fired after the function signature first, indenting the declaration output
+  /// one additional level. If true, A line break will be fired further up in the function's
+  /// declaration (e.g. generic parameters, parameters) before breaking on the function's output.
+  public var prioritizeKeepingFunctionOutputTogether = false
```
allevato: We can't use synthesized `Codable` because we have some custom logic in our `Configuration` type, so you'll also need to add this property manually to the `CodingKeys` enum and to the `init(from: Decoder)` and `encode(to: Encoder)` methods.

Documentation/Configuration.md
```diff
@@ -58,6 +58,16 @@ top-level keys and values:
     argument, forcing the entire argument list to be laid out vertically.
     If false (the default), arguments will be laid out horizontally first, with
     line breaks only being fired when the line length would be exceeded.
+    
+*   `prioritizeKeepingFunctionOutputTogether` _(boolean)_: Determines if
```
allevato: Interesting historical note: this was the original wrapping behavior that we had implemented by default. That was also back when we forced every argument to be on its own line (`lineBreakBeforeEachArgument = true`), so we ended up removing this logic because we felt it was too aggressive in the case of a function that had a lot of arguments but only needed to wrap because of the output clause—in other words, the output clause would force the `)` down onto its own line, which would also force every argument onto its own line.

That's less of an issue with `lineBreakBeforeEachArgument = false`, so having a knob to re-enable this behavior doesn't seem too bad.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -268,6 +268,14 @@ private final class TokenStreamCreator: SyntaxVisitor {
     after(node.funcKeyword, tokens: .break)
     after(node.identifier, tokens: .close)
 
+    // prioritize keeping ") throws -> <return_type>" together.
```
allevato: Unfortunately, this doesn't construct the token stream the way that we need it to be, and the test cases appear to be passing by coincidence (I'm actually surprised).

What's happening here is that we're at the mercy of visitation order. First, `FunctionDeclSyntax` is visited, and we call `before(node.signature.input.rightParen, tokens: .open)`. Then, after this function exits, at some pointer later we visit the children of the function decl, and that includes the `ParameterClauseSyntax` (`node.signature.input`) which also calls `before(node.rightParen, tokens: .break(.close, size: 0), .close)`.

The token stream will be created in that order, so that means you're not actually opening a new group just before the right-paren; you're opening a new group, putting a break inside it, and immediately closing it. If you run an example through the formatter with `--debug-dump-token-stream`, you can see this happening:

```
...
    [SYNTAX "(" Length: 1 Idx: 16]
    [BREAK Kind: open(kind: SwiftFormatPrettyPrint.OpenBreakKind.block) Size: 0 Length: 21 Ignores Discretionary NL: false Idx: 17]
    [OPEN Inconsistent Length: 21 Idx: 18]
      [OPEN Inconsistent Length: 8 Idx: 19]
        [SYNTAX "_" Length: 1 Idx: 20]
        [BREAK Kind: continue Size: 1 Length: 3 Ignores Discretionary NL: false Idx: 21]
        [SYNTAX "x" Length: 1 Idx: 22]
        [SYNTAX ":" Length: 1 Idx: 23]
        [BREAK Kind: continue Size: 1 Length: 4 Ignores Discretionary NL: false Idx: 24]
        [SYNTAX "Int" Length: 3 Idx: 25]
      [CLOSE Idx: 26]
---v
      [OPEN Inconsistent Length: 0 Idx: 27]
        [BREAK Kind: close(mustBreak: true) Size: 0 Length: 0 Ignores Discretionary NL: false Idx: 28]
      [CLOSE Idx: 29]
---^

      [SYNTAX ")" Length: 1 Idx: 30]
      [BREAK Kind: continue Size: 1 Length: 7 Ignores Discretionary NL: false Idx: 31]
      [SYNTAX "throws" Length: 6 Idx: 32]
      [BREAK Kind: continue Size: 1 Length: 5 Ignores Discretionary NL: false Idx: 33]
      [SYNTAX "->" Length: 2 Idx: 34]
      [SPACE Size: 1 Flexible: false Length: 1 Idx: 35]
      [SYNTAX "R" Length: 1 Idx: 36]
    [CLOSE Idx: 37]
  [CLOSE Idx: 38]
...
```

Whereas, what I think we want is this (ignore the lengths/indexes):

```
...
    [SYNTAX "(" Length: 1 Idx: 16]
    [BREAK Kind: open(kind: SwiftFormatPrettyPrint.OpenBreakKind.block) Size: 0 Length: 21 Ignores Discretionary NL: false Idx: 17]
    [OPEN Inconsistent Length: 21 Idx: 18]
      [OPEN Inconsistent Length: 8 Idx: 19]
        [SYNTAX "_" Length: 1 Idx: 20]
        [BREAK Kind: continue Size: 1 Length: 3 Ignores Discretionary NL: false Idx: 21]
        [SYNTAX "x" Length: 1 Idx: 22]
        [SYNTAX ":" Length: 1 Idx: 23]
        [BREAK Kind: continue Size: 1 Length: 4 Ignores Discretionary NL: false Idx: 24]
        [SYNTAX "Int" Length: 3 Idx: 25]
      [CLOSE Idx: 26]
---v
      [BREAK Kind: close(mustBreak: true) Size: 0 Length: 0 Ignores Discretionary NL: false Idx: 28]
    [CLOSE Idx: 29]
    [OPEN Inconsistent Length: 0 Idx: 27]
---^
      [SYNTAX ")" Length: 1 Idx: 30]
      [BREAK Kind: continue Size: 1 Length: 7 Ignores Discretionary NL: false Idx: 31]
      [SYNTAX "throws" Length: 6 Idx: 32]
      [BREAK Kind: continue Size: 1 Length: 5 Ignores Discretionary NL: false Idx: 33]
      [SYNTAX "->" Length: 2 Idx: 34]
      [SPACE Size: 1 Flexible: false Length: 1 Idx: 35]
      [SYNTAX "R" Length: 1 Idx: 36]
    [CLOSE Idx: 37]
  [CLOSE Idx: 38]
...
```

Unfortunately, this is tricky because:

* We can't do both things here because this function executes entirely before the `ParameterClauseSyntax` is visited, and we see above that that gets the ordering wrong.
* We can't do anything in the `visitPost` methods (i.e., on our way back up the tree) because by then, the token has already been claimed and added to the stream. Modifying the before/after maps at that point will be ignored.
* We could split this so that the first part happens in `visit(ParameterClauseSyntax)` and the second happens here, but we only want to do the first part if the output clause is present, and we can't determine that based on the `ParameterClauseSyntax`.

...Actually, maybe that third point isn't an issue. An `.open/.close` group surrounding just a lone `)` would be pointless, but also harmless. Could you try that (and likewise in `SubscriptDecl`)?

My other idea was going to be that we could have a way to push information down the tree by having `laterBefore/laterAfter` operations that you would start at a higher node (to keep related operations together) but then actually pick off in a lower node (to get the ordering right). That would also work and would be generally useful, but it's a bigger architectural change that we may not need right now.

Sources/SwiftFormatConfiguration/Configuration.swift
p4checo: Oops. That's what you get by splitting commits 😅. Fixed

Sources/SwiftFormatConfiguration/Configuration.swift
p4checo: Completely missed that, thanks!

p4checo:
On a different topic regarding tooling and development process, how are you able to effectively work on the `master` branch using Xcode 11?

I couldn't find the development snapshot used for the swift-sintax version used in `master` (`swift-DEVELOPMENT-SNAPSHOT-2019-09-26-a`), as the oldest one on swift.org is `swift-5.1-DEVELOPMENT-SNAPSHOT-2019-10-13-a`. My first attempt was using the most recent one (`swift-5.1-DEVELOPMENT-SNAPSHOT-2019-11-05-a`), but had the same issue.

I've tried opening both the `Package.swift` directly in Xcode (native SPM integration), and using the project generated by `swift package generate-xcodeproj`, with no success 😫 

The project successfully builds but I'm always getting `Parsing failed with error: SwiftSyntax parser library isn't compatible` when running the tests.

My current solution is to work branching from the `swift-5.1-branch`  and then cherry pick into a branch from `master`, which is far from ideal.

Am I missing something? Thanks

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
p4checo: Makes sense 🤔 I confess I didn't expect the visitation order would impact when "creating" the stream, because I thought we only had to ensure we properly open/close stuff. However I now see that the ordering of the injected tokens is relevant and impacts the outcome.

> An .open/.close group surrounding just a lone ) would be pointless, but also harmless. Could you try that (and likewise in SubscriptDecl)?

Not sure if I understood correctly, but I made that change (I think) and it appears to also work, but is still not providing the final "ideal" structure. Is this what you expected? 👇 

```
    [SYNTAX "(" Length: 1 Idx: 16]
    [BREAK Kind: open(kind: SwiftFormatPrettyPrint.OpenBreakKind.block) Size: 0 Length: 21 Ignores Discretionary NL: false Idx: 17]
    [OPEN Inconsistent Length: 21 Idx: 18]
      [OPEN Inconsistent Length: 8 Idx: 19]
        [SYNTAX "_" Length: 1 Idx: 20]
        [BREAK Kind: continue Size: 1 Length: 3 Ignores Discretionary NL: false Idx: 21]
        [SYNTAX "x" Length: 1 Idx: 22]
        [SYNTAX ":" Length: 1 Idx: 23]
        [BREAK Kind: continue Size: 1 Length: 4 Ignores Discretionary NL: false Idx: 24]
        [SYNTAX "Int" Length: 3 Idx: 25]
      [CLOSE Idx: 26]
      [OPEN Inconsistent Length: 1 Idx: 27]
        [OPEN Inconsistent Length: 0 Idx: 28]
          [BREAK Kind: close(mustBreak: true) Size: 0 Length: 0 Ignores Discretionary NL: false Idx: 29]
        [CLOSE Idx: 30]
        [SYNTAX ")" Length: 1 Idx: 31]
      [CLOSE Idx: 32]
      [BREAK Kind: continue Size: 1 Length: 7 Ignores Discretionary NL: false Idx: 33]
      [SYNTAX "throws" Length: 6 Idx: 34]
      [BREAK Kind: continue Size: 1 Length: 5 Ignores Discretionary NL: false Idx: 35]
      [SYNTAX "->" Length: 2 Idx: 36]
      [SPACE Size: 1 Flexible: false Length: 1 Idx: 37]
      [SYNTAX "R" Length: 1 Idx: 38]
    [CLOSE Idx: 39]
```

Resulting from:
```swift
    // prioritize keeping ") throws -> <return_type>" together.
    if config.prioritizeKeepingFunctionOutputTogether,
      let returnTypeToken = node.signature.output?.returnType.lastToken
    {
      before(node.signature.input.rightParen, tokens: .open)
      after(node.signature.input.rightParen, tokens: .close)

      before(node.signature.input.rightParen, tokens: .open)
      after(returnTypeToken, tokens: .close)
    }
```

> My other idea was going to be that we could have a way to push information down the tree by having laterBefore/laterAfter operations that you would start at a higher node (to keep related operations together) but then actually pick off in a lower node (to get the ordering right). That would also work and would be generally useful, but it's a bigger architectural change that we may not need right now.

This sure seems like the best solution going forward. Just curious, do you mean passing additional  information on `SyntaxVisitorContinueKind.visitChildren` so that we can apply custom formatting on the next level children. or is it something different?

allevato: I was thinking that we'd split this operation into two of the `visit` methods, so that the tokens are added in the right order with respect to when they are visited. In this one (`visit(FunctionDeclSyntax)`), we would do:

```swift
    // prioritize keeping ") throws -> <return_type>" together.
    if config.prioritizeKeepingFunctionOutputTogether {
      after(node.signature.lastToken, tokens: .close)
    }
```

Then in `visit(ParameterClauseSyntax)`, we would add the open:

```
    if config.prioritizeKeepingFunctionOutputTogether {
      before(node.rightParen, tokens: .open)
    }
```

The part about the group just around the `)` was describing what would happen in the case where the function didn't have an output clause, because we wouldn't be able to detect that down in `visit(ParameterClauseSyntax)` (without looking up through parent nodes, which I try to avoid as much as possible as a design principle, though it does happen sometimes). But then I realized that it's probably OK if that happens—groups just prioritize keeping tokens together when there are breaks between them, so a group with a single token and no breaks is basically the same as the token by itself. In other words, it should be harmless.

I haven't completely thought through whether doing this split would have any other adverse side effects, but it's the first thing that comes to mind that might get the ordering right (aside from the deeper architectural change that I mentioned).

> Just curious, do you mean passing additional information on `SyntaxVisitorContinueKind.visitChildren` so that we can apply custom formatting on the next level children. or is it something different?

We wouldn't be able to pass information down directly because we're tied to the `SyntaxVisitor` API, which says we have to return those specific enum values, and each node is visited in isolation with no direct responsibility to manage its own recursion (unless it returns `.skipChildren` and hijacks the entire walk for that node's descendants, which I want to avoid because it would make the control flow harder to reason about).

Instead, I'm thinking we could have the higher nodes in the tree associate "pending formatting tokens" with one of their descendants in a map, and then a later visit method could ask for those tokens from the map and put them in the appropriate `before/after` call. But I haven't completely thought through a good design for that.

Sources/SwiftFormatConfiguration/Configuration.swift
p4checo: Hadn't pushed any changes yet. Sorry for the miscommunication. Should be by now 😄

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -67,14 +67,6 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
   func before(_ token: TokenSyntax?, tokens: [Token]) {
     guard let tok = token else { return }
-    for preToken in tokens {
-      if case .open = preToken {
-        openings += 1
-      } else if case .close = preToken {
-        assert(openings > 0)
-        openings -= 1
-      }
-    }
```
p4checo: Had to remove this check and the one on `after`. Do you see any alternative solution so that we can keep these checks? I thought of adding some "unsafe" variants for these special cases, but wanted your feedback first.

Now that I look at it, I'm not even sure how my previous solution didn't fire the assert (at least on the `after(returnTypeToken, tokens: .close)`, since it also didn't have a matching `.open` token)  🤔

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
p4checo: > The part about the group just around the ) was describing what would happen in the case where the function didn't have an output clause, because we wouldn't be able to detect that down in visit(ParameterClauseSyntax) (without looking up through parent nodes, which I try to avoid as much as possible as a design principle, though it does happen sometimes). But then I realized that it's probably OK if that happens—groups just prioritize keeping tokens together when there are breaks between them, so a group with a single token and no breaks is basically the same as the token by itself. In other words, it should be harmless.

Got it! Thanks for clearing that up! 

> I haven't completely thought through whether doing this split would have any other adverse side effects, but it's the first thing that comes to mind that might get the ordering right (aside from the deeper architectural change that I mentioned).

As mentioned on a separate comment I had to remove some checks on `before` and `after` calls (which is not ideal, but I'm unsure what's the best approach here). Besides that, it appears to be working great now. 🙌 

> We wouldn't be able to pass information down directly because we're tied to the SyntaxVisitor API, which says we have to return those specific enum values, and each node is visited in isolation with no direct responsibility to manage its own recursion (unless it returns .skipChildren and hijacks the entire walk for that node's descendants, which I want to avoid because it would make the control flow harder to reason about).

You're right, and please forgive all these silly questions from me 🤦‍♂ 

> Instead, I'm thinking we could have the higher nodes in the tree associate "pending formatting tokens" with one of their descendants in a map, and then a later visit method could ask for those tokens from the map and put them in the appropriate before/after call. But I haven't completely thought through a good design for that.

Makes sense, thanks for the explanation! 🙏 

So if I understood correctly, it would something along the lines of the already existing `pendingMultilineStringSegmentPrefixLengths = [TokenSyntax: Int]()`,
but the challenge is defining the key type for the dictionary, as we have different "families" of `Syntax` types which don't sit under a common `Hashable` parent. For the dictionary value type we could simply define an `enum` with the possible operations (position + tokens to inject), or even a closure that would contain whatever token operations we wanted (more prone to misuse). Sounds like an interesting problem!

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
allevato: Ouch! I guess this is fine, since this change is a legitimate place where the `.close` needs to be added to the token stream before the `.open` due to visitation order. If they're imbalanced, it would be detected later by the pretty-printer anyway. Catching that earlier and locally would be nicer, but that's not possible here.

I think the reason your previous solution didn't assert is because the ordering still just happened to keep the total count above zero at all times.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
allevato: Can you add a comment just above the call to `after` here (and in `SubscriptDeclSyntax`), something to the effect of:

```swift
// Due to visitation order, the matching .open break is added in ParameterClauseSyntax.
```

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -849,6 +843,12 @@ private final class TokenStreamCreator: SyntaxVisitor {
   func visit(_ node: ParameterClauseSyntax) -> SyntaxVisitorContinueKind {
     after(node.leftParen, tokens: .break(.open, size: 0), .open(argumentListConsistency()))
     before(node.rightParen, tokens: .break(.close, size: 0), .close)
+
+    // prioritize keeping ") throws -> <return_type>" together.
+    if config.prioritizeKeepingFunctionOutputTogether {
+      before(node.rightParen, tokens: .open)
```
allevato: Likewise, can you comment before the `before` call:

```swift
// Due to visitation order, this .open corresponds to a .close added in FunctionDeclSyntax
// or SubscriptDeclSyntax.
```

Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -129,10 +140,11 @@ public class Configuration: Codable {
       BlankLineBetweenMembersConfiguration.self, forKey: .blankLineBetweenMembers)
       ?? BlankLineBetweenMembersConfiguration()
     self.lineBreakBeforeControlFlowKeywords
-      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeControlFlowKeywords)
-      ?? true
+      = try container.decodeIfPresent(Bool.self, forKey: .lineBreakBeforeControlFlowKeywords) ?? false
```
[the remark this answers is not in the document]
p4checo: no worries, though so too!

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
p4checo: Great, thanks!

PR #75, opened 2019-11-06
Wrap after assignment operators, not before.

allevato:
This required some changes to the pretty-printer around newlines
that are adjacent to breaks that I'd like to revisit at some point
in the future; there are a lot of subtleties in situations involving
breaks immediately before a newline vs. immediately after a newline
that I think could be simplified and cleaned up.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
@@ -20,6 +20,9 @@ public class PrettyPrinter {
 
   /// Information about an open break that has not yet been closed during the printing stage.
   private struct ActiveOpenBreak {
+    /// The index of the open break.
+    let index: Int
+
     /// The kind of open break that created this scope.
     let kind: OpenBreakKind
 
@@ -258,7 +261,6 @@ public class PrettyPrinter {
 
       switch kind {
       case .open(let openKind):
-        wasLastBreakKindContinue = openKind == .continuation
         let lastOpenBreak = activeOpenBreaks.last
         let currentLineNumber = openCloseBreakCompensatingLineNumber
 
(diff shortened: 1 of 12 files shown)


allevato:
cc @dylansturg

Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
```diff
@@ -494,7 +517,13 @@ public class PrettyPrinter {
       // of any prior break tokens.
       case .newlines:
         if let index = delimIndexStack.last, case .break = tokens[index] {
-          lengths[index] += total
+          if index == i - 1 {
+            // A break immediately preceding a newline should have a length of zero, so that it
+            // doesn't fire.
+            lengths[index] = 0
```
dylansturg: Is this to avoid dealing with a newline after a break fires? I'm wondering if that could still be an issue in cases where the break is encountered after some printed tokens that **must** exceed the maximum line length (e.g. a really long function name in a function decl).

Tests/SwiftFormatPrettyPrintTests/TernaryExprTests.swift
```diff
@@ -41,11 +41,12 @@ public class TernaryExprTests: PrettyPrintTestCase {
       """
     let expected =
       """
-      let someLocalizedText = shouldUseTheFirstOption
-        ? stringFunc(name: "Name1", comment: "short comment")
-        : stringFunc(
-          name: "Name2", comment: "Some very, extremely long comment",
-          details: "Another comment")
+      let someLocalizedText =
+        shouldUseTheFirstOption
```
dylansturg: It seems like this should still fit on the previous line. Was this break a result of additional grouping?

Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
allevato: That seems to work ok; the following example breaks correctly, with the curly brace even being reset appropriately. Is this the kind of situation you were thinking of?

```swift
func test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test(foo: Int) {
  foo()
  bar()
}

// becomes

func
  test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test_test(
    foo: Int
  )
{
  foo()
  bar()
}
```

Tests/SwiftFormatPrettyPrintTests/TernaryExprTests.swift
allevato: The input text has a newline after the `=`, so it was "invalid" before and got removed. Now it's being preserved as a discretionary line break.

Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
dylansturg: Yup, that looks good. Thanks for checking!

Tests/SwiftFormatPrettyPrintTests/TernaryExprTests.swift
dylansturg: Doh. I should've seen that in the input. Thanks for clarifying.

PR #76, opened 2019-11-06
Fix collapsing of func param attribute

namolnad:
Currently the pretty printer does not add a break after a function parameter attribute, which will now be more common in SwiftUI due to the `@ViewBuilder` attribute. This results in the following behavior: 

**Before formatting:**
```
func foo<Content: View>(@ViewBuilder bar: () -> View) {
      bar()
}
```
**After formatting:**
```
func foo<Content: View>(@ViewBuilderbar: () -> View) {
      bar()
}
```

This PR adds a break in between and after the attribute tokens to ensure a space (or break) is maintained in these cases.

~_Note: I've been unable to get the tests to run locally due to a SwiftSyntax parser incompatibility error, so I haven't been able to confirm this works as expected._~ (Was able to test these against the `5.1` branch)

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -854,6 +854,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: FunctionParameterSyntax) -> SyntaxVisitorContinueKind {
     before(node.firstToken, tokens: .open)
+    arrangeAttributeList(node.attributes)
     after(node.colon, tokens: .break)
     before(node.secondName, tokens: .break)
 
--- a/Tests/SwiftFormatPrettyPrintTests/FunctionDeclTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/FunctionDeclTests.swift
@@ -700,4 +700,28 @@ public class FunctionDeclTests: PrettyPrintTestCase {
       """
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 21)
   }
+
+  func testDoesNotCollapseFunctionParameterAttributes() {
+    let input =
+    """
+    func foo<Content: View>(@ViewBuilder bar: () -> View) {
+      bar()
+    }
+
+    """
+
+    assertPrettyPrintEqual(input: input, expected: input, linelength: 60)
+  }
+
+  func testDoesNotCollapseStackedFunctionParameterAttributes() {
+    let input =
+    """
+    func foo<Content: View>(@FakeAttr @ViewBuilder bar: () -> View) {
+      bar()
+    }
+
+    """
+
+    assertPrettyPrintEqual(input: input, expected: input, linelength: 80)
+  }
 }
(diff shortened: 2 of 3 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -853,6 +853,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
   }
 
   func visit(_ node: FunctionParameterSyntax) -> SyntaxVisitorContinueKind {
+    after(node.attributes?.lastToken, tokens: .break)
```
allevato: Thinking about this further after my initial Twitter comment, I realize this won't handle stacked property wrappers—if you add a test like this, I bet it would mash them together:

```swift
func foo<Content: View>(@Foo @ViewBuilder bar: () -> View) {
  bar()
}
```

So instead of inserting this break, can you replace this line with a call to `arrangeAttributeList(node.attributes)`? That will do all the work of inserting breaks between the individual attributes, after the last one, and also adding some groups to keep them together when possible. You can see an example of this on line 939-ish in the `visit(EnumCaseDeclSyntax)` method.

(The reason this has to be an explicit call instead of happening automatically as part of the regular visitation of the syntax tree is due to the visitation order w.r.t. the way the before/after maps are built up.)

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -853,6 +853,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
     before(node.firstToken, tokens: .open)
```
allevato: I'd also recommend leaving this as the first line of the function, so that the relationship between the `.open` group before the first token and the rest of the content is clear to the reader, since this `.open` group should encompass everything else in the parameter node, including the attributes. (`arrangeAttributeList` also calls `before()` on the same node, so ordering here theoretically _does_ matter, even though _practically_ it doesn't in this specific case because the only thing that method inserts is another `.open`.)

Tests/SwiftFormatPrettyPrintTests/FunctionDeclTests.swift
```diff
@@ -700,4 +700,15 @@ public class FunctionDeclTests: PrettyPrintTestCase {
...
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 21)
   }
+
+  func testDoesNotCollapseFunctionParameterAttributes() {
+    let input =
+    """
+    func foo<Content: View>(@ViewBuilder bar: () -> View) {
+      bar()
+    }
+    """
```
allevato: Once you've made the change above, please add a test case with a parameter with two stacked property wrappers so we can verify that works as well.

allevato:
> Note: I've been unable to get the tests to run locally due to a SwiftSyntax parser incompatibility error, so I haven't been able to confirm this works as expected.

You'll need the trunk development snapshot that matches the swift-syntax dependency in the branch you're developing on, so for master that's 09-26, but unfortunately it looks like that can't be downloaded from swift.org anymore 😕   So we'll need to move to a new version of swift-syntax, but that'll also involve some refactoring...

Another option would be to develop against `swift-5.1-branch` (which should work with a stock Xcode 11 release) and then patch your change back into `master` for the PR. I'll test it locally before merging.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
namolnad: Sounds good, it did feel a bit odd adding this before the `.open` token, though I wasn't positive I fully understood how these tokens would come together. Will add this right after the `.open` token per your suggestion. Thanks!

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
namolnad: great call! just developed against the 5.1 branch and the stacked case was indeed broken. your suggested fix works like a charm!

namolnad:
Sounds good, will do momentarily!

allevato:
Thanks again for the quick fix! Merging now.

namolnad:
Of course—thanks for the quick review and merge!

PR #77, opened 2019-11-07
Fix formatting for vars with default values and accessors.

allevato:
We were losing the space before the equal sign when a pattern
binding had both an initial value and an accessor block. This
change makes the logic more robust, and even handles the
"syntactically valid" but compiler-rejected case of multiple
comma-delimited bindings that have accessor blocks.


allevato:
cc @dylansturg

PR #78, opened 2019-11-07
Correctly apply debug flags in lint mode.

allevato:


allevato:
cc @dylansturg

SDGGiesbrecht:
Wow! That was fast.

PR #80, opened 2019-11-08
Clarify SwiftPM Products

SDGGiesbrecht:
This includes the `SwiftFormatConfiguration` module in the `SwiftFormat` product, since its configuration type is necessary in order to instantiate any formatter or linter.

It also includes a standalone `SwiftFormatConfiguration` library product. It is separately useful for clients to be able to create, modify, encode, decode, load and dump configurations without incurring the cost of the SwiftPM or SwiftSyntax dependencies. (This will make even more of a difference once [SE‐0226](https://github.com/apple/swift-evolution/blob/master/proposals/0226-package-manager-target-based-dep-resolution.md) is available.)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,8 @@ let package = Package(
   name: "swift-format",
   products: [
     .executable(name: "swift-format", targets: ["swift-format"]),
-    .library(name: "SwiftFormat", targets: ["SwiftFormat"]),
+    .library(name: "SwiftFormat", targets: ["SwiftFormat", "SwiftFormatConfiguration"]),
+    .library(name: "SwiftFormatConfiguration", targets: ["SwiftFormatConfiguration"]),
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-package-manager.git", from: "0.4.0"),


SDGGiesbrecht:
Originates from [this forum thread](https://forums.swift.org/t/swiftformat-as-a-package-dependency/30531).

Package.swift
```diff
@@ -17,7 +17,8 @@ let package = Package(
   name: "swift-format",
   products: [
     .executable(name: "swift-format", targets: ["swift-format"]),
-    .library(name: "SwiftFormat", targets: ["SwiftFormat"]),
+    .library(name: "SwiftFormat", targets: ["SwiftFormat", "SwiftFormatConfiguration"]),
+    .library(name: "SwiftFormatConfiguration", targets: ["SwiftFormatConfiguration"]),
```
aciidb0mb3r: Does it make sense to ever want `SwiftFormatConfiguration` without `SwiftFormat`?

SDGGiesbrecht: My use case for that is a product that only needs to be able to dynamically assemble a configuration and convert it to JSON for export. The larger SwiftFormat library and its dependencies on SwiftSyntax and SwiftPM are massive overkill for this; they take a long time to fetch and build.

PR #81, opened 2019-11-13
Update BlankLineBetweenMembers rule to remove false positives.

dylansturg:
The rule was overly eager in adding blank lines around members that have comments in their leading trivia, because it didn't always correctly determine whether a given comment should trigger blank lines around the member.

I have refactored the rule using the following criteria for adding blank lines:
1. Any member whose leading trivia contains a non-blank-line-terminated comment must have a leading blank line (before the comment) and a trailing blank line (after the member).
2. Any comment that contains at least 1 blank line between the comment and the relevant syntax token is ignored entirely by the rule.

~This also includes formatting of the files that I touched; the changes were small enough that I think it's reasonable to include in a single PR. I'm fine with doing a separate pre-format PR though if the diff here is at all confusing.~ Reverted formatting changes.


dylansturg:
@allevato This PR is now updated to revert the formatting changes as we discussed.

PR #83, opened 2019-11-15
Migrate to swift-tools-support-core.

allevato:
This removes the dependency on the private libraries inside SwiftPM
that we've been using for argument parsing, file streams, and other
things, and uses the recently factored out swift-tools-support-core
repository instead which is intended to be publicly consumed.


allevato:
cc @dylansturg

PR #88, opened 2019-11-19
Fix formatting for `ObjectLiteralExpr`s.

allevato:
I believe we had this previously, when tuple expressions and function
call arguments were distinct nodes. But we didn't have explicit test
cases for it, and once those two nodes were merged and the argument
handling had to be hoisted up into the `FunctionCallExpr` node, we
didn't do the same for `ObjectLiteralExpr` and lost those.

This refactors the commonalities of function argument list formatting
a bit more so that it's easier to reuse, should it be needed in other
nodes in the future.


allevato:
cc @dylansturg

Tests/SwiftFormatPrettyPrintTests/ObjectLiteralExprTests.swift
```diff
@@ -0,0 +1,114 @@
+import SwiftFormatConfiguration
+
+public class ObjectLiteralExprTests: PrettyPrintTestCase {
```
dylansturg: Very thorough tests! 👏

PR #91, opened 2019-11-20
Fix scrunching of range operators.

allevato:
We can't unconditionally remove all whitespace around range operators
when they are preceded by a postfix operator or followed by a
postfix operator; the parser greedily treats the combined string
as a single operator.

We still remove whitespace around range operators when possible, but
rather than try to strictly enforce the lack of whitespace in all cases
through more clever and potentially disruptive transformations (like
surrounding the operands by parentheses), we just keep the whitespace
when we can't remove it.

Fixes SR-11798.


allevato:
cc @dylansturg

PR #92, opened 2019-11-20
Handle end-of-line comments correctly in BlankLineBetweenMembers.

allevato:
When we were doing the is-single-line check, we weren't considering that
the first comment in the trivia might precede any newlines in that trivia,
meaning it's an end-of-line comment for a previous line, not a line
comment that we should consider part of the current decl. Likewise, when
we insert a newline, we need to insert it after that initial EOL comment,
if it exists.

Complex logic like this should probably be factored into a more general
trivia transform utility at some point, so it can be localized.


allevato:
cc @dylansturg

PR #93, opened 2019-11-20
Allow the closing paren for function decls in protocols on the same l…

dylansturg:
…ine as the last parameter.

Breaking before this paren doesn't aid in the readability in the same way it does when the decl isn't part of a protocol, because there's no function body nor curly braces after the paren in this case. This break existed to aid readability by creating a boundary between the function's signature and its body. When the decl is part of a protocol, there is no body and the boundary isn't necessary.

In order to make this change, I had to hoist the grouping and breaking around parens in a function signature from the ParameterClauseSyntax visitor into parent visitor methods. The ParameterClauseSyntax is nested below the node which can determine if the decl has a body. Instead of traversing parent pointers, I hoisted the implementation into the parent visitors.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -262,6 +262,9 @@ private final class TokenStreamCreator: SyntaxVisitor {
       after(node.identifier.lastToken, tokens: .space)
     }
 
+    let mustBreak = node.body != nil || node.signature.output != nil
+    arrangeParameterClause(node.signature.input, forcesBreakBeforeRightParen: mustBreak)
+
     arrangeFunctionLikeDecl(
       node,
       attributes: node.attributes,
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -262,6 +262,11 @@ private final class TokenStreamCreator: SyntaxVisitor {
       after(node.identifier.lastToken, tokens: .space)
     }
 
+    let parameters = node.signature.input
+    let mustBreak = node.body != nil || node.signature.output != nil
+    after(parameters.leftParen, tokens: .break(.open, size: 0), .open(argumentListConsistency()))
```
allevato: Can you factor these two lines out into a private helper function like `arrangeParameterClause` that takes the `ParameterClauseSyntax` and `forcesBreakBeforeRightParen` as arguments, and use that throughout? That'll reduce some duplication.

I had to do something similar when I moved around the function call argument list formatting: https://github.com/apple/swift-format/pull/88/files#diff-1935f1a8f63b8de81e00cb07ae1f6ebaR703

But your version can be simpler because we don't need to pass the delimiters in separately (that was necessary because for the call, it could be either parens or brackets).

allevato: Thanks!

PR #94, opened 2019-11-22
Put freestanding `init` decls in a test into a type.

allevato:
Even though these aren't semantically valid Swift, these parse correctly
with the 9/26 snapshot we're currently on. At a more recent commit we
noticed that these `init`s seem to be hitting a code path that causes an
assertion in the parser:

    assertion failed in swift::Type swift::ValueDecl::getInterfaceType() const:
    ctx.getLegacyGlobalTypeChecker() && "The type checker must be installed to
    make semantic queries!"

We should try to pinpoint this crash, but for now I'm patching this so
that the test is more "correct" and it unblocks some of our internal work.

### DIFF
--- a/Tests/SwiftFormatRulesTests/ValidateDocumentationCommentsTests.swift
+++ b/Tests/SwiftFormatRulesTests/ValidateDocumentationCommentsTests.swift
@@ -187,28 +187,30 @@ public class ValidateDocumentationCommentsTests: DiagnosingTestCase {
   public func testInitializer() {
     let input =
     """
-    /// Brief summary.
-    ///
-    /// - Parameter command: The command to execute in the shell environment.
-    /// - Returns: Shouldn't be here.
-    init(label commando: String) {
-    // ...
-    }
+    struct SomeType {
+      /// Brief summary.
+      ///
+      /// - Parameter command: The command to execute in the shell environment.
+      /// - Returns: Shouldn't be here.
+      init(label commando: String) {
+      // ...
+      }
 
-    /// Brief summary.
-    ///
-    /// - Parameter command: The command to execute in the shell environment.
-    init(label command: String) {
-    // ...
-    }
+      /// Brief summary.
+      ///
+      /// - Parameter command: The command to execute in the shell environment.
+      init(label command: String) {
+      // ...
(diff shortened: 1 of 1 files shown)


akyrtzi:
\cc @nathawes could you take a look?

nathawes:
@allevato Are you still seeing this after the changes in https://github.com/apple/swift/pull/28226? That fixed several cases where we were hitting this assertion in SourceKit's syntactic requests so might fix this too.

allevato:
@nathawes Thanks for the link. I first wanted to grab the 2019-11-11 trunk snapshot (the snapshot that Swift for TensorFlow's branch was last rebased on, before your patch) to verify failure, but I wasn't able to reproduce my issue using either SwiftSyntax or `-frontend -emit-syntax` from the command line.

Digging a little deeper, I think this might be a result of a subtlety in Google's internal Bazel build for libswiftSyntaxParser; we're not defining `SWIFT_BUILD_ONLY_SYNTAXPARSERLIB` because we can't change configurations for the same build targets based on who depends on them, so the parser we build ends up with some real code paths in libAST instead of the `#if SWIFT_BUILD_ONLY_SYNTAXPARSERLIB`-guarded early-returns that are there. I'm wondering if one of those code paths was triggering the assertion.

Since this might be related to our weird setup, I'll have to wait until Swift for TensorFlow rebases on top of your change and see if the old `init`s still fail, and look into addressing the issue with our build targets so that we define the proper preprocessor symbols to avoid this again (if that is indeed the problem).

akyrtzi:
Actually the `SWIFT_BUILD_ONLY_SYNTAXPARSERLIB` is not defined in any of the builds that Swift CI produces, currently we only use that internally.

allevato:
So much for that theory then! Indeed, I just tried defining that symbol universally in our internal build as a quick test and it didn't make the issue go away, so I'm glad I'm wrong because it means I don't have to rearrange all those build targets.

Anyway, @nathawes' patch was small enough that I could apply it on top of our internal copy of the toolchain and I can verify that that **did** fix the issue; the tests passed again with the surrounding type removed from the initializers. Thanks!

PR #95, opened 2019-12-03
Grouping and continuation consistency improvements.

dylansturg:
Changes:
1. Add a group around the RHS of assignments (including pattern binding, e.g. let and var initializers) when the RHS expression is "compound".
2. Add a continuation open break before the RHS of assignments in a sequence expression to be consistent with pattern binding, which already has a continuation open break before the initializer expression.

Compound means an expression that consists of multiple subexpressions, including wrapper expressions, like try, by investigating the contents of the wrapped expression.

The additional groups around RHS of assignments better encapsulate the subexpression so that the RHS is visually separated from the LHS and assignment token.

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/AssignmentExprTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/AssignmentExprTests.swift
@@ -1,3 +1,5 @@
+import SwiftFormatConfiguration
+
 public class AssignmentExprTests: PrettyPrintTestCase {
   public func testBasicAssignmentExprs() {
     let input =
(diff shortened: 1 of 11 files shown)


dylansturg:
@allevato What do you think about these formatting changes for expressions? This better follows the conventions of some other formatters that we've discussed by breaking before complex expressions when assigning & initializing.

Tests/SwiftFormatPrettyPrintTests/BinaryOperatorExprTests.swift
```diff
@@ -16,8 +16,9 @@ public class BinaryOperatorExprTests: PrettyPrintTestCase {
 
     let expected10 =
       """
-      x = 1 + 8
-        - 9
+      x =
+        1 + 8
+          - 9
         ^*^ 5
```
allevato: Shouldn't these last two lines now also line up with the `- 9`?

dylansturg: Yeah, that's what I would expect. This is a result of the way our sequence folding works though. The "^*^" operator is unknown, so it's precedence is unknown (represented by `nil`) when this sequence expr is folded. When the unrecognized operator is encountered, the sequence containing the `=` is split there.

I would expect the sequence tree for this expr to look something like...
```
    =
  /   \
x     [RHS]  
```

But instead, it looks like...
```
        "\"
        /  \
       *   10
      / \
   "^*^" 4
   /  \
  =    5
 / \
x  [1 + 8 - 9]
```

Because of the shape of that tree of sequence exprs, the "^*^" operator and everything following it isn't included in the `break(open(.continuation))` on the RHS of the assignment operator. I think solving this would require folding the sequence such that everything to the right of the "=" is included in the right subtree sequence expr so that it can be included in that open break.

allevato: Riiight, got it. Ok, I'm somewhat less worried about it now. I may look at having unknown operators fold just below `AssignmentPrecedence` (or `TernaryPrecedence`) instead of where they do now—this doesn't risk divergence from the compiler because an unknown operator is an error there, but it would make the formatting results look better.

Tests/SwiftFormatPrettyPrintTests/AssignmentExprTests.swift
```diff
@@ -30,53 +32,164 @@ public class AssignmentExprTests: PrettyPrintTestCase {
...
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 30)
   }
 
-  public func testAssignmentFromFunctionCalls() {
+  public func testAssignmentOperatorFromSequenceWithFunctionCalls() {
+    let input =
+        """
+        result = firstOp + secondOp + someOpFetchingFunc(foo, bar: bar, baz: baz)
+        result = someOpFetchingFunc(foo, bar: bar, baz: baz)
+        result += someOpFetchingFunc(foo, bar: bar, baz: baz)
```
allevato: For completeness, can you add a `+=` case that has a compound expression on the RHS?

dylansturg: Done.

PR #96, opened 2019-12-05
Use continuation opens instead of block opens around conditions in if…

dylansturg:
… and guard stmts.

The block type of open creates 2x indents when the open break occurs on a line that is a continuation, because the relevant active-open-break can contribute a block indent and a continuation indent. The conditions in if and guard stmts aren't code blocks. These open breaks were added to support stacking a single continuation inside of each condition. The continuation type of open was added later and better models the behavior desired in this case.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -412,13 +412,12 @@ private final class TokenStreamCreator: SyntaxVisitor {
   func visit(_ node: IfStmtSyntax) -> SyntaxVisitorContinueKind {
     after(node.ifKeyword, tokens: .space)
 
-    // Add break groups around any conditions after the first so that those conditions have at least
-    // 1 non-continuation based indent from an indentation stack. Without that indent, all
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -418,7 +418,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
     // line because continuation breaks don't stack. There are no breaks around the first condition
     // because if-statements look better without a break between the "if" and the first condition.
     for condition in node.conditions.dropFirst() {
-      before(condition.firstToken, tokens: .break(.open, size: 0))
+      before(condition.firstToken, tokens: .break(.open(kind: .continuation), size: 0))
```
allevato: The comment above this loop probably needs to be updated now—in particular, the part about "so that those conditions have at least 1 _non-continuation based indent_ from an indentation stack".

dylansturg: Doh. Good call. I updated the similar comment for guard statements too.

PR #97, opened 2019-12-05
Various fixes for switch/case statements.

allevato:
- When combining integer fallthrough-only cases into ranges, we
  were dropping their where-clauses.
- It's not entirely safe to collapse integer cases into ranges,
  because it's possible to write a type that implements `~=` such
  that it only pattern matches against integers but not ranges.
  These are now collapsed into comma-delimited lists.
- It's not entirely safe to remove fallthrough-only cases before
  a default clause. Even though the default would match those, the
  patterns could have side effects. This is awful, but possible,
  and would break at runtime (not detected at compile-time).

The pretty-printer now also inserts newlines before case items
that have where clauses if they immediately follow a case item
that does not have a where clause, to avoid compiler warnings.


allevato:
cc @dylansturg

PR #98, opened 2019-12-05
Don't remove spaces around a range operator followed by a prefix dot.

allevato:
I *believe* this is the last remaining place where an "operator character"
would occur in an expression following a binary range operator where that
character would not already be identified itself as an operator token.

Fixes SR-11798.


allevato:
cc @dylansturg

PR #101, opened 2019-12-05
Support tab based indentation in the whitespace linter.

dylansturg:
The linter was crashing whenever it encountered a tab because it only expected spaces to be used for whitespace. The linter now supports a wider range of whitespace characters and allows using tabs for indentation. This includes some revisions to the error messages to add tabs are as a suggestion for adding/removing indentation. There is also a new error message used when tabs (or really any non-space) characters are used for spacing (i.e. the whitespace that isn't indentation).

SR-11211

### DIFF
--- a/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
+++ b/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SwiftFormatConfiguration
 import SwiftFormatCore
 import SwiftSyntax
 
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift
```diff
@@ -414,16 +427,124 @@ public class WhitespaceLinter {
...
+  /// The line's leading whitespace consists of multiple runs of different kinds of whitespace
+  /// characters.
+  case heterogeneous([Indent])
+}
+
+extension Indent {
+  /// Returns a string that describes the indentation in a human readable format, which is
+  /// appropriate for use in diagnostic messages.
+  var diagnosticDescription: String {
+    switch(self) {
```
allevato: Remove parens around the `switch` argument here, and two (I think) places below please 🙂  (We should have a rule!)

dylansturg: Done. The existing rule handled it; I just forgot to run the formatter. 😅

PR #105, opened 2019-12-12
Improve wrapping of parenthesized expressions.

allevato:
Since paren exprs are just represented as one-element tuples, they
were getting the same kind of wrapping as tuples, with breaks inside
the parens. This doesn't look great, because these are different kinds
of expressions. Now, we keep the parens inseparable from the tokens
that they are adjacent to, which lets them wrap more traditionally,
and we increase the continuation indentation inside them.

This required some changes to the way we stack indentation for
variable initialization and assignment to make multiple parenthesized
expressions line up nicely. For example, whereas we used to do this:

```swift
x =
  foo
    + bar
    + baz
```

Now we do

```swift
x =
  foo
  + bar
  + baz
```

...so that if those terms are paren exprs and they wrap inside, they
still shift in and out nicely and line up well:

```swift
x =
  (foo
    + foo2)
  + (bar
    + bar2)
  + (baz
    + baz2)
```

We find the loss of the original extra level of indentation to be
acceptable because the subsequent lines still start with a spaced
operator or some other token (like a prefix dot) that indicates
that the line is unambiguously a continuation line.


allevato:
cc @dylansturg

Tests/SwiftFormatPrettyPrintTests/TernaryExprTests.swift
```diff
@@ -43,10 +43,10 @@ public class TernaryExprTests: PrettyPrintTestCase {
       """
       let someLocalizedText =
         shouldUseTheFirstOption
-          ? stringFunc(name: "Name1", comment: "short comment")
-          : stringFunc(
-            name: "Name2", comment: "Some very, extremely long comment",
-            details: "Another comment")
+        ? stringFunc(name: "Name1", comment: "short comment")
+        : stringFunc(
```
dylansturg: Should the lines starting with "?" and ":" (the choices of the ternary) be indented +1 continuation from the condition here? Seems like that might be nice to show that it's a continuation of ternary expr.

If that's a change that we want, I think it's fine to fix in a later PR.

allevato: I thought about that for a bit too when I saw this change and I wasn't sure which way I wanted to go with it. I think I'll need to see what happens when I start fixing nested ternaries to see where things land. On the one hand, the continuation lines start with operators so it's clear they're continuations, but on the other hand a ternary has more "structure" than regular operators and showing that structure through indentation might also be good...

PR #106, opened 2019-12-12
Don't break inside empty parens of a no-argument func/init decl.

allevato:
This unfortunately required a bit more shuffling of code around in those
visitors to make sure that the after-tokens got inserted in the right
relative orders.

This ended up slightly changing the behavior for initializers with
generic arguments and attributes, but I like the new behavior better and
think it's actually more in the spirit of what we want; we now insert a
line break after the attributes because the rest of the function signature
is wrapped onto more than one line.


allevato:
cc @dylansturg

PR #107, opened 2019-12-12
Support ignoring AST nodes using a comment, for specific node types.

dylansturg:
Originally, I had hoped to support ignoring any type of node by simply including a comment before the node. That turned out to be infeasible due to an existing reliance on eventually visiting every TokenSyntax in the AST. For example, some nodes attach printing tokens before or after TokenSyntax instances that will be visited later. In the event that a TokenSyntax is ignored, those attached before/after tokens are also ignored leading to mismatches in the token stream.

Rather than to fix all such instances of that dependency, which may not itself be feasible, I've instead limited the ignore directive to 2 "top level" types of nodes: code block list items and member decl list items. This allows ignoring on an entire statement or member.

Verbatim token changes:
I had to add a new option on `Verbatim` to support only indenting the first line of the text contained in the token. Originally, verbatim always printed the indentation of the current context on every line of text. That breaks when verbatim is used to print text that's nested in other decls. Each time the formatter runs, the verbatim text is indented by its original indentation + the indentation of the current context.


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -2558,6 +2574,83 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
+
+  /// Returns whether the given trivia includes a directive to ignore formatting for the next node.
+  ///
+  /// - Parameter trivia: Leading trivia for a node that the formatter supports ignoring.
+  private func isFormatterIgnorePresent(inTrivia trivia: Trivia) -> Bool {
+     for piece in trivia {
+       switch piece {
+       case .lineComment(let text), .blockComment(let text),
+         .docLineComment(let text), .docBlockComment(let text):
+         if text.contains("swift-format-ignore") { return true }
```
allevato: Can we be more precise here, perhaps with a prefix check (I believe we'll have to strip off the comment leading characters, though)? That way we allow this:

```swift
// swift-format-ignore
```

but not this:

```swift
// Do disable the formatter, write "// swift-format-ignore".
```

Also, let's remove the support for `docLineComment` and `docBlockComment`; we don't want people to put directives in their documentation. If they want to disable formatting on a documented decl, we should require this:

```swift
// swift-format-ignore
/// Blah blah blah
func foo() {
  poorlyFormattedCode()
}
```

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -2558,6 +2574,83 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
+    // Regardless of the level of nesting, if the ignore directive is present on the first token
+    // contained within the node then the entire node is eligible for ignoring.
+    if let firstTrivia = node.firstToken?.leadingTrivia {
+      return isFormatterIgnorePresent(inTrivia: firstTrivia)
+    }
+
+    // Some nodes consist of multiple tokens where the user might place the ignore directive. For
+    // example, the comment could appear after a list of modifiers on a decl. Search for the ignore
+    // directive on a top level token, but don't search inside of any other children.
+    for token in node.children where token is TokenSyntax {
```
allevato: Can you refresh my memory of a specific case where the `firstToken` is non-nil but where we might want to support the directive on other immediate children? And then can you add that example to the comment so that I don't forget it again? 🙂

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -981,6 +981,14 @@ private final class TokenStreamCreator: SyntaxVisitor {
   }
 
   func visit(_ node: MemberDeclListItemSyntax) -> SyntaxVisitorContinueKind {
+    if shouldFormatterIgnore(node: node) {
+      appendFormatterIgnored(node: node)
+
+      // Add this break so that trivia parsing will allow discretionary newlines after the node.
+      appendToken(.break(.same, size: 0))
```
allevato: Can this be moved into `appendFormatterIgnored(node:)` to reduce a little bit of the call sites' duplication?

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: I've dropped doc comments and updated to specifically check for exactly "swift-format-ignore" (ignoring whitespace).

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: This was support situations where the ignore comment is placed between some attributes and/or modifiers on decl. For example:
```
public  // swift-format-ignore
  struct Foo {
}
```

As we discussed, I've reverted this part of the change so that the comment must be leading trivia of the first token in the decl.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: Good call. Done.

dylansturg:
> Once this is in, are we also going to follow up with a change to migrate the phase 1 rule support over to this model?

That's the plan, but I was sending it out in stages to keep the changes manageable.

PR #109, opened 2019-12-14
Replace enable/disable directives with ignore for formatter rules.

dylansturg:
Previously, the rules used an "active" enable/disable model where rules were explicitly enabled or disabled for a number of lines by using paired enable & disable comments. This model makes sense for rules, but it proved to be challenging to implement for the pretty printing portion of the formatter. Instead of having diverging patterns, I've reimplented the rules logic to use ignore next node style comments like were previously implemented in the pretty printer.

In addition, I've added support for comma delimited lists of rules since I was rewriting most of code anyway.

Authors may no longer use `// swift-format-[enable|disable]: [rule-name]`. Instead, rules can only be "ignored" by adding `// swift-format-ignore[: rule-name, rule-name, rule-name]`.

## Considerations
I opted to use "activate" to replace the behavior offered starting with "swift-format-enable". This functionality allows authors to create off-by-default rules that are selectively enabled on sections of code. The baseline swift-format package doesn't have any off-by-default rules, but I didn't want to break the ability to add those rules.

I'm not thrilled with the name; I tried to find a more passive verb (like ignore) because "activate" implies there could later be a corresponding "deactivate". I explicitly didn't use "enable" to avoid any perceived conflict with existing enable/disable comments.

I'm open to renaming "activate" to something else if there are suggestions.

### DIFF
--- a/Sources/SwiftFormatCore/Context.swift
+++ b/Sources/SwiftFormatCore/Context.swift
@@ -75,14 +75,11 @@ public class Context {
   /// location or not.
   public func isRuleEnabled(_ ruleName: String, node: Syntax) -> Bool {
     let loc = node.startLocation(converter: self.sourceLocationConverter)
-    guard let line = loc.line else { return false }
-    switch ruleMask.ruleState(ruleName, atLine: line) {
+    switch ruleMask.ruleState(ruleName, at: loc) {
     case .default:
       return configuration.rules[ruleName] ?? false
     case .disabled:
       return false
-    case .enabled:
-      return true
     }
   }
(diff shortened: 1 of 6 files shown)


dylansturg:
I agree that "activate" is more trouble than it's worth. It doesn't create a pattern that I think we should incentivize anyone to use. 

I've removed the activate capability and simplified the code where possible. This also allowed me to apply an optimization where we stop visiting children once a rule-less ignore is encountered, since all rules are disabled in all subsequent children.

Please take a look and let me know what you think.

PR #111, opened 2019-12-18
Fix a number of bugs in UseShorthandTypeNames.

allevato:
This started as a fix for a bug where nested generic argument
clauses were being dropped, but I kept peeling back layers of
the onion and uncovering other cases that were missed or
places where we were doing the wrong thing. I took the
opportunity to rewrite the rule from the ground up, especially
the part that transforms types into equivalent "expressions",
making it truly recursive so that it handles more situations
correctly.

The tests have also been cleaned up, to make it clearer what
each test case is actually meant to be testing.


allevato:
cc @dylansturg

PR #113, opened 2019-12-20
Support IfConfigDecl clauses in UseEnumForNamespacing

klaaspieter:
Anything inside a IfConfigDecl was ignored when determining if a struct
could be an enum. This looks for IfConfigDeclSyntax in the
MemberDeclListSyntax and only turns the struct into an enum when all of
the clauses have namespace compatible members.

Fixes: https://bugs.swift.org/browse/SR-11830

### DIFF
--- a/Sources/SwiftFormatRules/UseEnumForNamespacing.swift
+++ b/Sources/SwiftFormatRules/UseEnumForNamespacing.swift
@@ -76,6 +76,18 @@ public final class UseEnumForNamespacing: SyntaxFormatRule {
         }
         // Do not append private initializer.
 
+      case let decl as IfConfigDeclSyntax:
+        let membersToKeep: [MemberDeclListSyntax] = decl.clauses
+          .compactMap { clause in
+            (clause.elements as? MemberDeclListSyntax).flatMap(membersToKeepIfUsedAsNamespace(_:))
+          }
+
+        if membersToKeep.count < decl.clauses.count {
+          return nil
+        } else {
+          declList.append(member)
+        }
+
       default:
         declList.append(member)
       }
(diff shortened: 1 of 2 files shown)


allevato:
It looks like the commit history here has commits from the 5.1 branch included in it, which may be the case if you were doing work there. Can you clean that up?

klaaspieter:
@allevato Yeah. I'm in the middle of adding those extra tests. I wanted to actually try and test against master so I've been trying to get the `swift-DEVELOPMENT-SNAPSHOT-2019-09-26-a` toolchain.

Is the only way to get it to build it from source?

klaaspieter:
Should be fixed now. Although I still haven't been able to test this against master 😞

dylansturg:
> @allevato Yeah. I'm in the middle of adding those extra tests. I wanted to actually try and test against master so I've been trying to get the `swift-DEVELOPMENT-SNAPSHOT-2019-09-26-a` toolchain.
> 
> Is the only way to get it to build it from source?

You can still download that snapshot from Swift.org, it's just hard to find since the link has long-since been removed from the UI. You can just change the date in the download URL to match any snapshot. For example, here's the link for 2019-09-26-a snapshot for use in Xcode:
https://swift.org/builds/development/xcode/swift-DEVELOPMENT-SNAPSHOT-2019-09-26-a/swift-DEVELOPMENT-SNAPSHOT-2019-09-26-a-osx.pkg

klaaspieter:
@dylansturg worth updating the README with that link?

dylansturg:
> @dylansturg is it worth updating the README with that link?

I think we want to get away from that old snapshot and onto a newer one that supports the latest Swift-Syntax soonish™️ so it may not be relevant for too much longer.

PR #117, opened 2020-01-15
Multithread processing source files

keith:
Previously every source file was formatted / linted one by one. On our
codebase this took a full project format from ~17 minutes to ~5 minutes.

### DIFF
--- a/Sources/swift-format/Frontend/Frontend.swift
+++ b/Sources/swift-format/Frontend/Frontend.swift
@@ -92,7 +92,7 @@ class Frontend {
     if paths.isEmpty {
       processStandardInput()
     } else {
-      processPaths(paths)
+      processPaths(paths, parallel: lintFormatOptions.parallel)
     }
   }
 
@@ -124,29 +124,46 @@ class Frontend {
   }
 
   /// Processes source content from a list of files and/or directories provided as paths.
-  private func processPaths(_ paths: [String]) {
+  private func processPaths(_ paths: [String], parallel: Bool) {
     precondition(
       !paths.isEmpty,
       "processPaths(_:) should only be called when paths is non-empty.")
 
-    for path in FileIterator(paths: paths) {
-      guard let sourceFile = FileHandle(forReadingAtPath: path) else {
-        diagnosticEngine.diagnose(Diagnostic.Message(.error, "Unable to open \(path)"))
-        continue
+    let lock = NSLock()
+    if parallel {
+      let allFilePaths = Array(FileIterator(paths: paths))
+      DispatchQueue.concurrentPerform(iterations: allFilePaths.count) { index in
+        let path = allFilePaths[index]
+        openAndProcess(path, lock: lock)
       }
-
-      guard let configuration = configuration(
-        atPath: lintFormatOptions.configurationPath, orInferredFromSwiftFileAtPath: path)
-      else {
-        // Already diagnosed in the called method.
-        continue
+    } else {
+      for path in FileIterator(paths: paths) {
+        openAndProcess(path, lock: nil)
       }
-
-      let fileToProcess = FileToProcess(
-        fileHandle: sourceFile, path: path, configuration: configuration)
-      processFile(fileToProcess)
     }
   }
+
+  /// Read and process the given path, optionally synchronizing diagnostic output with the given lock.
+  private func openAndProcess(_ path: String, lock: NSLock?) -> Void {
+    guard let sourceFile = FileHandle(forReadingAtPath: path) else {
+      lock?.lock()
+      defer { lock?.unlock() }
+      diagnosticEngine.diagnose(Diagnostic.Message(.error, "Unable to open \(path)"))
+      return
+    }
+
+    guard let configuration = configuration(
+      atPath: lintFormatOptions.configurationPath, orInferredFromSwiftFileAtPath: path)
+    else {
+      // Already diagnosed in the called method.
+      return
+    }
+
+    let fileToProcess = FileToProcess(
+      fileHandle: sourceFile, path: path, configuration: configuration)
+    processFile(fileToProcess)
+  }
+
 
   /// Returns the configuration that applies to the given `.swift` source file, when an explicit
   /// configuration path is also perhaps provided.
--- a/Sources/swift-format/Subcommands/LintFormatOptions.swift
+++ b/Sources/swift-format/Subcommands/LintFormatOptions.swift
@@ -49,6 +49,9 @@ struct LintFormatOptions: ParsableArguments {
     """)
   var ignoreUnparsableFiles: Bool
 
+  @Flag(help: "Whether or not to run formatting or linting in parallel consuming all possible resources")
+  var parallel: Bool
+
   /// The list of paths to Swift source files that should be formatted or linted.
   @Argument(help: "Zero or more input filenames.")
   var paths: [String]


keith:
This is meant as a RFC for how we can achieve this if this is something we want to do and if it's safe to do.

harlanhaskins:
Could you use `DispatchQueue.concurrentPerform`?

keith:
@harlanhaskins I was planning on doing that, but the file discovery right now is lazy (which I assume was very intentional) so we don't have a count up front to use

Sources/swift-format/main.swift
```diff
@@ -69,18 +69,36 @@ fileprivate func main(_ arguments: [String]) -> Int32 {
...
 ) -> Int32 {
   var result = 0
+  let concurrentQueue = DispatchQueue(label: "com.queue.Concurrent", qos: .userInitiated, attributes: .concurrent)
+  let group = DispatchGroup()
+  let lock = NSLock()
   for path in FileIterator(paths: paths) {
-    guard let sourceFile = FileHandle(forReadingAtPath: path) else {
-      stderrStream.write("Unable to create a file handle for source from \(path).\n")
-      stderrStream.flush()
-      return 1
```
kastiglione: Previously this aborted the whole set of files. Now it seems like it would have "keep-going" behavior.

allevato:
This is definitely something we've been wanting to look into, so thanks for kicking it off for us.

At a very high-level, this feels like it should be safe because the closure being called (`formatMain` or `lintMain`) is effectively `main` after the command line options have been type-safely-parsed, so there's not really any globally mutable state that we should be concerned about. (Famous last words.)

But I also feel like my Dispatch skills have atrophied enough after years of mostly writing command-line Swift and subsets-of-Python-build-rules that I'm not the best person to make such pronouncements 😛   @dylansturg , care to take a look?

One possibility would be to add a `--parallel` flag to control this at first, if we have any concerns. Since command line parsing is done right before the `processSources` call, it would be trivial to pipe that in and make a decision.

keith:
Definitely happy to add that flag if that's something folks want!

dylansturg:
I patched this change in a couple ways, and the formatter is **consistently crashing** on:
- A single simple file (e.g. just an empty function)
- A directory containing 1 simple file
- A directory containing many thousands of files

The first configuration was patching this in `master`, using the 2019-09-26 toolchain snapshot. That consistently crashes a few levels deep in SwiftSyntax's `doVisit`. LLDB also crashes when I have the debugger connected, so I don't have much detail.

The second configuration that I tried was patching in `swift-5.1-branch` with Xcode 11.0 and it's packaged toolchain. This also consistently crashes in SwiftSyntax, a few levels deep in the visitor methods. LLDB doesn't crash, so I was able to get an error message:
>  error: Trying to put the stack in unreadable memory at: 0x700005880850.

@allevato Mentioned that he previously encountered an issue where SwiftSyntax's recursive visitor implementation regularly exceeded the stack size on non-main threads. I wonder if that's what is happening here? Could also be a red herring.

Finally, what configuration did you use for testing? I'd like to understand what's different, and why I'm seeing this crash consistently but you didn't.

jpsim:
It might be worth running this with TSan enabled to see if it catches anything.

akyrtzi:
> @allevato Mentioned that he previously encountered an issue where SwiftSyntax's recursive visitor implementation regularly exceeded the stack size on non-main threads. I wonder if that's what is happening here? Could also be a red herring.

See the JIRA bug (https://bugs.swift.org/browse/SR-11170) for the workaround.

allevato:
Thanks @akyrtzi , I wasn't able to find that JIRA issue for some reason. So the good news is, that specific problem will at least be mitigated once we migrate `master` from the 09-26 snapshot to all the newer SwiftSyntax APIs that landed since then. (If that's indeed the problem causing the crashes @dylansturg is seeing, which I'm unsure about given the test case of a file with just an empty function.)

ahoppen:
Are you seeing the stack issues in both debug and release builds or only debug builds? The issue I encountered (and fixed in https://github.com/apple/swift-syntax/pull/147) only occurred in debug builds.

dylansturg:
@ahoppen I only tried debug originally. I tried again on release, and I'm not seeing any crash in release. Sounds like it's the issue that you fixed?

ahoppen:
@dylansturg Yes, the issue should be fixed in `master` then.

Sources/swift-format/main.swift
```diff
@@ -69,18 +69,36 @@ fileprivate func main(_ arguments: [String]) -> Int32 {
...
+    group.enter()
+    concurrentQueue.async {
```
p4checo: small 💅 improvement: you can pass in the `group` directly to `DispatchQueue.async()`, avoiding the `.enter()` and `.leave()` calls 😉

akyrtzi:
> I think we should wait until we update the formatter to the latest version of swift-syntax

Formatter has been updated.🙂

keith:
After rebasing I believe I'm still seeing the stack issue that was mentioned with a debug build, I can try release tomrorow

akyrtzi:
Use a `Thread` with a custom stack size, as I mention in https://bugs.swift.org/browse/SR-11170

ahoppen:
@keith, I have just gotten a chance to take a look at the stack overflow issue. It seems like the problem I had, resurfaced. I just opened https://github.com/apple/swift-syntax/pull/205 which should solve the issue. Could you try your changes in combination with my patch?

keith:
With that change it does work with debug builds

ahoppen:
@keith Great. I have just merged the changes. Once a tag has been created for the new version, could you also adjust the version of SwiftSyntax that `swift-format` depends on in `Package.swift`?

Sources/swift-format/main.swift
keith: +1, given that I assume this error should be an edge case hopefully this behavior change is ok

keith:
I've updated swift-format to include that change, unfortunately I had to use a commit instead of a tag here https://forums.swift.org/t/no-new-master-snapshots-since-2-21/34200

Sources/swift-format/main.swift
```diff
@@ -69,18 +69,36 @@ fileprivate func main(_ arguments: [String]) -> Int32 {
...
+        stderrStream.write("Unable to create a file handle for source from \(path).\n")
+        stderrStream.flush()
+        lock.lock()
+        result |= 1
+        lock.unlock()
+        return
+      }
+
+      let configuration = loadConfiguration(forSwiftFile: path, configFilePath: configurationPath)
+      let transformResult = transform(sourceFile, path, configuration)
```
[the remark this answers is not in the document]
keith: It looks like there is a way to make it write atomically https://github.com/apple/swift-tools-support-core/blob/e11ef6f91b3bfcca36bda9b0f0cc38876328db78/Sources/TSCBasic/FileSystem.swift#L378-L395 although right now I believe this is broken since it appears `moveItem` doesn't actually allow overwriting:

```
Error Domain=NSCocoaErrorDomain Code=516 "“foo.txt” couldn’t be moved to “swift-format” because an item with the same name already exists." UserInfo={NSSourceFilePathErrorKey=/Users/ksmiley/dev/oss-swift/swift-format/foo.txt, NSUserStringVariant=(
    Move
), NSDestinationFilePath=/Users/ksmiley/dev/oss-swift/swift-format/bar.txt, NSFilePath=/Users/ksmiley/dev/oss-swift/swift-format/foo.txt, NSUnderlyingError=0x7fd475d19f20 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}
```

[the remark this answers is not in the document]
keith: I submitted https://github.com/apple/swift-tools-support-core/pull/56 to fix this

Sources/swift-format/Utilities/Helpers.swift
```diff
@@ -42,16 +42,33 @@ func readSource(from fileHandle: FileHandle) -> String? {
...
+        lock.unlock()
+      }
     }
-    let configuration = try loadConfiguration(forSwiftFile: path, configFilePath: configurationPath)
-    transform(sourceFile, path, configuration)
+  }
+
+  group.wait()
+  if let error = lastError {
+    throw error
```
keith, line 71: Up for suggestions on different error handling options here

keith:
I've rebased this again and dropped my Package.swift changes since those are no longer required. I would love to know what we can do to get this merged!

Sources/swift-format/Utilities/Helpers.swift
allevato: Thanks for continuing with this in the face of constant refactoring 😅 (Another one of which is coming down the pipe: https://github.com/apple/swift-format/pull/178)

Taking a closer look at the original code (before your PR), its intended behavior was unclear to me (because it's an edge case that we've never hit). If we failed to open a source file, we'd emit a diagnostic and then return. If an error is thrown trying to read the config file for a particular file, we let it get thrown passed us, letting the argument parser's logic catch it. In either case, an error at any time stops linting/formatting for the whole file set.

Spot-checking what some other formatters do in similar situations:

* Prettier [logs an error message and continues on](https://github.com/prettier/prettier/blob/master/src/cli/util.js#L456-L464) if it can't open the file.
* Clang-format [logs an error message and continues if it fails to open a source file](https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/ClangFormat.cpp#L350-L353) or if there's a [problem getting the style](https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/ClangFormat.cpp#L382-L385), among other conditions.

So our current code was making this a lot harder than it needed to be, and https://github.com/apple/swift-format/pull/178 cleans all that up (among other things) so now instead of worrying about early exit or throwing conditions, a bad file/configuration just causes a diagnostic to be emitted and we move on, and you don't have to think about the error conditions at all.

---

Aside from that, I was thinking that it would be nice to control this with a `-j <N>/--jobs=<N>` flag so that we can try rolling it out off by default at first and then flip it on later once we've put it through some stress tests. But AFAIK Dispatch makes it kind of hard to control that. Would it make sense to use `OperationQueue` here instead?

soffes: Personally, I much prefer a `--parallel` (or equivalent name) flag instead of `--jobs=N` because I'm never sure how many to pick. I think it would be great if it just let `Dispatch` decide how many jobs instead of letting the user control it :)

allevato: That would work too. I don't have too strong an opinion here about the specifics, as long as it's off by default (for now) and doesn't cause excessive thread creation when it's enabled.

Maybe each format/lint job is short-running enough that we don't need to worry about finer grained control, or we can start with a binary `--parallel` flag now and add job count control later if we find we need it.

Sources/swift-format/Utilities/Helpers.swift
```diff
@@ -42,16 +42,33 @@ func readSource(from fileHandle: FileHandle) -> String? {
...
+  let concurrentQueue = DispatchQueue(label: "com.queue.Concurrent", qos: .userInitiated, attributes: .concurrent)
+  let group = DispatchGroup()
+  let lock = NSLock()
+  var lastError: Error?
   for path in FileIterator(paths: paths) {
-    guard let sourceFile = FileHandle(forReadingAtPath: path) else {
-      diagnosticEngine.diagnose(
-        Diagnostic.Message(.error, "Unable to create a file handle for source from \(path)."))
-      return
+    concurrentQueue.async(group: group) {
```
akyrtzi, line 52: You should use `DispatchQueue.concurrentPerform(iterations:execute:)`, it will simplify the code a bit and is strongly recommended over passing `async` blocks to a concurrent queue without any upper bound check for how many run concurrently. The latter can lead to thread exhaustion.

keith: I don't see a specific answer about this, but there was previous conversation about it https://github.com/apple/swift-format/pull/117#issuecomment-574898189 I think the file discovery is intentionally lazy so we don't know the number of iterations here

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -49,6 +49,9 @@ struct LintFormatOptions: ParsableArguments {
     """)
   var ignoreUnparsableFiles: Bool
 
+  @Flag(help: "Whether or not to run formatting or linting in parallel consuming all possible resources")
```
keith, line 52: I opted not to put a docstring here since it would say the same thing, happy to add one if we'd like to keep them on everything

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -124,27 +124,42 @@ class Frontend {
...
-    for path in FileIterator(paths: paths) {
+    let concurrentQueue = DispatchQueue(
+        label: "swift-format-queue", qos: .userInitiated, attributes: .concurrent)
+    let lock = NSLock()
+    let openAndProcess: (String) -> Void = { path in
       guard let sourceFile = FileHandle(forReadingAtPath: path) else {
-        diagnosticEngine.diagnose(Diagnostic.Message(.error, "Unable to open \(path)"))
-        continue
+        lock.lock()
+        defer { lock.unlock() }
```
keith, line 138: Note this is on both paths but I assume that's ok since this is an unexpected case anyways

Sources/swift-format/Utilities/Helpers.swift
akyrtzi: @allevato is there a reason that the file discovery cannot be eager?

Sources/swift-format/Utilities/Helpers.swift
keith: Added the `parallel` flag!

Sources/swift-format/Frontend/Frontend.swift
allevato: There's this diagnostic, but then there's all the others. Have you tried parallel mode with the `lint` command yet, especially on a large set of files that have a large number of warnings?

SwiftSyntax's `DiagnosticEngine` and `PrintingDiagnosticConsumer` don't have any synchronization that I can see, and in particular, `PrintingDiagnosticConsumer.handle` makes [multiple calls](https://github.com/apple/swift-syntax/blob/master/Sources/SwiftSyntax/PrintingDiagnosticConsumer.swift#L39-L54) to `FileHandle.write`, so unless I've missed something, I'd be surprised if this doesn't interleave diagnostics.

If we're going to support a parallel mode, I think we either need to provide our own diagnostic consumer that synchronizes its `handle` method, or make SwiftSyntax's `PrintingDiagnosticConsumer` do it by default. @akyrtzi

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -124,27 +124,42 @@ class Frontend {
...
+    }
+
+    for path in FileIterator(paths: paths) {
+      if parallel {
+        concurrentQueue.async {
+          openAndProcess(path)
+        }
+      } else {
+        openAndProcess(path)
+      }
```
allevato, line 162: You'll still need a `DispatchGroup` to wait on here, right? Otherwise the process could exit after enqueuing but not necessarily exiting all of the work.

Sources/swift-format/Subcommands/LintFormatOptions.swift
allevato: That's fine, both are probably redundant and the `help` string clearly takes priority since it's user-facing.

Sources/swift-format/Utilities/Helpers.swift
allevato: @akyrtzi Just that at the time it was written, there was no reason for it to be, so we could avoid the startup delay of collecting everything up front if someone ran the tool with `--recursive` on a large directory structure.

Is `DispatchQueue.concurrentPerform` preferred over `OperationQueue`, which also provides control over the number of concurrent jobs?

Sources/swift-format/Frontend/Frontend.swift
keith: Added the lock to the args of this and only passed it in the concurrent case

Sources/swift-format/Utilities/Helpers.swift
keith: I pushed the `concurrentPerform` option so everyone can see what that would look like, happy to change it to whatever we want here

akyrtzi:
> If we're going to support a parallel mode, I think we either need to provide our own diagnostic consumer that synchronizes its `handle` method, or make SwiftSyntax's `PrintingDiagnosticConsumer` do it by default. @akyrtzi

Yes, it makes sense to have `PrintingDiagnosticConsume` synchronize its writes to `stderr` so that it writes out a full diagnostic uninterrupted, but I don't think it is enough, you'd want to avoid interposing diagnostics from different files.

I would propose that diagnostics should be collected for each file and then printed as a group for each file in a sensible manner at the end, instead of printing to stderr as soon as something comes up.
That would also give you the opportunity to improve presentation as well later on, maybe optionally put out an html report or something, instead of printing to stderr.

harlanhaskins:
You could have one diagnostic engine and one consumer per file, and then they can replay their diagnostics out to the outside engine

akyrtzi:
> @akyrtzi Just that at the time it was written, there was no reason for it to be, so we could avoid the startup delay of collecting everything up front if someone ran the tool with --recursive on a large directory structure.

Does this startup delay really matter, the total time find+process files will still be the same at the end. And if you go with collecting the diagnostics to avoid having diagnostics printed to stderr interposed from different files, then you'll still not see some diagnostic until all the files are processed.

> Is DispatchQueue.concurrentPerform preferred over OperationQueue, which also provides control over the number of concurrent jobs?

The benefit is that you'd just let Dispatch decide how many threads to use, instead of having to manually chose. But either way would be better than just continuously doing `async` on a concurrent queue.

allevato:
> And if you go with collecting the diagnostics to avoid having diagnostics printed to stderr interposed from different files, then you'll still not see some diagnostic until all the files are processed.

I agree that we don't want the diagnostics from different files to be interposed, but I still think we can do better than deferring the output of **all** diagnostics until **all** files are processed.

If we have to write our own diagnostic consumer that collects diagnostics and groups them by file, then we can just have a method on that consumer that says to flush the diagnostics for that file, call it when processing of that file is complete, and synchronize around that. (For diagnostics with `unknown` files, we could just emit those immediately, also synchronized.) Or as @harlanhaskins suggested above, use separate engines for each file, but that one worries me a bit because in the future we might want to have a JSON output mode that keeps all the diagnostics in a single output file and separate engines would make that harder since we'd have to merge them all afterwards.

> The benefit is that you'd just let Dispatch decide how many threads to use, instead of having to manually chose. But either way would be better than just continuously doing async on a concurrent queue.

`OperationQueue` also has `defaultMaxConcurrentOperationCount` which lets the system decide how many jobs to run concurrently, but I don't know the details of how it selects that number compared to `DispatchQueue.performConcurrent`.

However, as I mentioned in one of my replies to Keith above, I think it would be fine to have `--parallel` mode eagerly evaluate the iterator into an array while leaving non-parallel mode to do it lazily, and doesn't require any changes to the way we traverse the file system.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -124,27 +124,42 @@ class Frontend {
...
       }
 
       let fileToProcess = FileToProcess(
         fileHandle: sourceFile, path: path, configuration: configuration)
-      processFile(fileToProcess)
+      self.processFile(fileToProcess)
+    }
+
+    let lock = NSLock()
+    let allFilePaths = Array(FileIterator(paths: paths))
```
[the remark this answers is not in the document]
keith: Should we go back to the other option then? Not sure I see a big win from using `concurrentPerform` at the moment especially if it requires this extra branching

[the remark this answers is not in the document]
keith: (Pushed this change in the meantime)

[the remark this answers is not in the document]
allevato: That's why I was curious about `OperationQueue`, because it seems like it still provides a way to let the system choose the parallelism while also letting us walk the file hierarchy lazily, by just calling `addOperation` repeatedly instead of requiring the iteration count up front. 🤷‍♂ 

But if that's not advisable for some reason, I don't consider pre-computing the file list only in parallel mode to be extra branching—rather, it's choosing the best distribution of the work given the requirements of the APIs we're using (or not using).

keith:
At this point with debug or release mode I see crashes in SwiftSyntax:

```
Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called
swift-format(12906,0x7000042ba000) malloc: Incorrect checksum for freed object 0x7ff84ac8b7a8: probably modified after being freed.
Corrupt value: 0x3000080184b7284f
 

Thread 0:: Dispatch queue: com.apple.root.user-interactive-qos
0   swift-format                  	0x0000000103c5027c partial apply + 108
1   swift-format                  	0x0000000103c4d2ca SyntaxRewriter.visit(_:) + 330 (SyntaxRewriter.swift:4534)
2   swift-format                  	0x0000000103c4dc80 SyntaxRewriter.visitChildren<A>(_:) + 2224 (SyntaxRewriter.swift:5052)
3   swift-format                  	0x0000000103b9bc1e SyntaxRewriter.visit(_:) + 206 (SyntaxRewriter.swift:324)
4   swift-format                  	0x0000000103bcf19d SyntaxRewriter.visitImplBooleanLiteralExprSyntax(_:) + 1661 (SyntaxRewriter.swift:2147)
5   swift-format                  	0x0000000103c52428 partial apply + 104
```

harlanhaskins:
What does TSan say? These structures should be immutable...

keith:
So with tsan I first hit many other issues. Looks like we're hitting some non-thread safe code elsewhere, like the use of DiagnosticEngine in rules:

https://github.com/apple/swift-format/blob/5786e26754c100f5e4e7d8df9e75ab50be9a9ce7/Sources/SwiftFormatWhitespaceLinter/WhitespaceLinter.swift#L344-L354

shahmishal:

keith:
If someone is interested in picking up this change that would be great! Last I remember I believe the next blocker is that the diagnostics reporting types do not support multithreading

mattt:
I'm facing the same threading issues with my use of the SwiftFormatter API for [swift-doc](https://github.com/swiftdocorg/swift-doc). Running with TSAN, it found a data race here:

https://github.com/apple/swift-format/blob/d4bba6e22891ff1813e8267e36f2b00307684366/Sources/SwiftFormatCore/Rule.swift#L34

```
SUMMARY: ThreadSanitizer: data race Rule.swift:34 in static Rule.ruleName.getter
==================
==================
WARNING: ThreadSanitizer: data race (pid=13716)
  Read of size 8 at 0x0001115c6888 by thread T4:
    #0 static Rule.ruleName.getter Rule.swift:34 (swift-doc:x86_64+0x1003c1c50)
    #1 SyntaxFormatRule.visitAny(_:) SyntaxFormatRule.swift:34 (swift-doc:x86_64+0x1003ce72f)
    #2 SyntaxRewriter.visitImplSourceFileSyntax(_:) SyntaxRewriter.swift:2778 (swift-doc:x86_64+0x100f27e82)
```

allevato:
Thanks for catching that one; it's not one of the original data races we dealt with in earlier iterations of this PR (the linked code was added later; it wasn't a race before because we inadvertently were never updating the cache 😬) but we'll need to go back and synchronize that now too.

@keith is correct about the original blocking issue being that `DiagnosticEngine` in SwiftSyntax not being thread-safe. Last I chatted with @akyrtzi he wasn't opposed to just synchronizing inside that class by default, so I can try to put together a PR for that soon unless someone else beats me to it.

keith:
Here's a change to make DiagnosticEngine thread safe: https://github.com/apple/swift-syntax/pull/243

keith:
Here's a change to fix the issue Mattt mentioned that I also hit once DiagnosticEngine was usable from multiple threads: https://github.com/apple/swift-format/pull/242

keith:
Here's a new PR for this change https://github.com/apple/swift-format/pull/243

PR #120, opened 2020-01-21
Replaces the `MultiLineTrailingCommas` phase 1 rule with Pretty Printer tokens.

dylansturg:
The phase 1 rule for adding trailing commas was incapable of being 100% correct, because a phase 1 cannot accurately determine whether an array or dictionary literal requires multiple lines. The only part of the formatter that has the necessary information is the PrettyPrinter.

There is a new pair of tokens: `commaDelimitedRegionStart` and `commaDelimitedRegionEnd`. A comma is inserted at the end token if and only if the end token is on a different line than the start token. The PrettyPrinter uses a stack of line numbers for the start token to implement this.


Sources/SwiftFormat/Pipelines+Generated.swift
```diff
@@ -243,8 +233,12 @@ extension LintPipeline {
     return .visitChildren
   }
 
-  func visit(_ node: SwitchStmtSyntax) -> SyntaxVisitorContinueKind {
+  func visit(_ node: SwitchCaseListSyntax) -> SyntaxVisitorContinueKind {
```
allevato: Oh, crap, I never regenerated this, did I 😞 

At some point we should add a test that verifies that this file is up-to-date.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -47,6 +47,10 @@ private final class TokenStreamCreator: SyntaxVisitor {
   /// stream.
   private var pendingMultilineStringSegmentPrefixLengths = [TokenSyntax: Int]()
 
+  /// Lists tokens that shouldn't be appended to the token stream as `syntax` tokens. They will be
+  /// printed conditionally using a different type of token.
+  private var ignoredTokens =  [TokenSyntax]()
```
allevato: In practice this is probably minor, but since `TokenSyntax`es are `Hashable`, you can use a `Set<>` instead of an array here to make the lookup closer to constant-time.

Sources/SwiftFormat/Pipelines+Generated.swift
dylansturg: Maybe a pre-commit hook? We might want to do the same for formatting the formatter soon too.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift

Sources/SwiftFormat/Pipelines+Generated.swift
dylansturg: Maybe a pre-commit hook? We might want to do the same for running the formatter on the formatter soon too.

allevato: We could make a test that invokes `generate-pipeline` and then verifies that the output is the same as the file in the source tree. Then once we're on CI, failure to regenerate would catch it.

A pre-commit hook that just auto-generates it could be smoother since it would prevent failure, but I'm a little wary of executing that much logic in a pre-commit hook (I'm not sure if there's ever a time where we _wouldn't_ want that thing to run automatically).

PR #122, opened 2020-01-23
Diagnostics for trailing comma changes.

dylansturg:
As part of the migration from a phase 1 rule for trailing commas to using the PrettyPrinter, we still need to emit diagnostics when the source's use of trailing commas is incorrect. These diagnostics are similar to those from the original phase 1 rule, except they don't specify the type of collection (instead saying "collection literal"). I think context of the line and column number + the general term "collection" are sufficient for the user to understand the intent.

I added support to PrettyPrintTestCase for collecting diagnostics so that they can be tested.

Finally, the WhitespaceLinter is only able to handle changes in whitespace. By adding/removing trailing commas, the whitespace linter would fail. Instead of checking the whitespace linter to support skipping comma changes, I've created a "mode" in the PrettyPrinter to print the text of the source exactly as it was while only changing whitespace. If we ever introduce other types of text edits in the PrettyPrinter, this should work for that too.

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/PrettyPrintTestCase.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/PrettyPrintTestCase.swift
@@ -6,20 +6,69 @@ import XCTest
 @testable import SwiftFormatPrettyPrint
 
 public class PrettyPrintTestCase: XCTestCase {
+  /// A helper that will keep track of which diagnostics have been emitted, and their locations.
+  private var consumer: DiagnosticTrackingConsumer? = nil
+
+  /// Implements `DiagnosticConsumer` to keep track which diagnostics have been raised and their
+  /// locations.
+  private class DiagnosticTrackingConsumer: DiagnosticConsumer {
+    var registeredDiagnostics = [(String, line: Int?, column: Int?)]()
+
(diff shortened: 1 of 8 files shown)


Tests/SwiftFormatPrettyPrintTests/PrettyPrintTestCase.swift
```diff
@@ -6,37 +6,90 @@ import XCTest
 @testable import SwiftFormatPrettyPrint
 
 public class PrettyPrintTestCase: XCTestCase {
+  /// A helper that will keep track of which diagnostics have been emitted, and their locations.
+  private var consumer: DiagnosticTrackingConsumer? = nil
+
+  /// Implements `DiagnosticConsumer` to keep track which diagnostics have been raised and their
+  /// locations.
+  private class DiagnosticTrackingConsumer: DiagnosticConsumer {
+    var registeredDiagnostics = [(String, line: Int?, column: Int?)]()
```
allevato: Dude, I was just about to make a similar change to the phase 1 rule tests so I could more accurately check the line/column numbers of the comment diagnostics I added. _Spooky_ 👻 

Later on we should merge things to have a common `XCTestCase` subclass for both rule "phases" that has this `XCTAssertDiagnosed` behavior in it.

PR #123, opened 2020-01-23
Make the `NoBlockComments` rule lint-only.

allevato:
After some real world use, this rule feels a bit too aggressive
as a format rule. The opportunities where the automatic fix are
known to be harmless are fairly limited—e.g., end-of-line or
single-line block comments.

When block comments spanned multiple lines, or started at the
end of a line and continued on multiple lines, determining the
appropriate amount of leading indentation to preserve for the
subsequent lines is not always trivial (consider the case of a
multi-line comment where some lines are indented less than the
first).

In other situations, block comments might be used to temporarily
comment out large blocks of code, and transforming these to
line comments is impolite, to say the least.


allevato:
cc @dylansturg 

This change also adds something I've wanted for a while; an easy way to compute the `AbsolutePosition`/`SourceLocation` of a specific `TriviaPiece` of a node (right now, just leading trivia). I should upstream this to SwiftSyntax eventually.

PR #125, opened 2020-01-23
No trailing commas in array/dict initializers.

dylansturg:
swift-syntax doesn't differentiate the array/dict expr syntax used for initializing an array/dict (where the elements are type identifiers) from an array/dict expr syntax of a collection literal (where the elements are the contents of the collection). There isn't a straightforward way to determine which is which when visiting the array or dict, since it's based on the parent being a function call.

A simple solution is to never add a trailing comma to array/dict when there is only 1 element. This impacts multiline array/dict literals with only 1 element, but that's probably uncommon and it's reasonable to skip the trailing comma for only 1 element.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -707,7 +707,12 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: ArrayElementListSyntax) -> SyntaxVisitorContinueKind {
     insertTokens(.break(.same), betweenElementsOf: node)
-    if let firstElement = node.first, let lastElement = node.last {
+
+    // The syntax library can't distinguish an array initializer (where the elements are types) from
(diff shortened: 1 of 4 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -707,7 +707,7 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   func visit(_ node: ArrayElementListSyntax) -> SyntaxVisitorContinueKind {
     insertTokens(.break(.same), betweenElementsOf: node)
-    if let firstElement = node.first, let lastElement = node.last {
+    if let firstElement = node.first, let lastElement = node.last, firstElement != lastElement {
```
allevato: Can you comment before this line (and in the dictionary function) why the `first != last` check is important so that future contributors (and future _us_) have that information? It's a little less discoverable below the fold of a commit message in the git history.

dylansturg: Good idea. This isn't the most obvious problem and solution.

PR #127, opened 2020-01-28
Preserve garbage text trivia when pretty printing.

allevato:
Garbage text typically refers to unrecognized characters in the input,
which we haven't typically cared about. However, there are also some
uses of garbage text that don't represent "garbage" per se, but content
that the lexer/parser doesn't care about but which we should still
preserve during formatting:

* a hashbang line at the beginning of the file
* a Unicode byte-order marker at the beginning of the file
* source control conflict markers

When we see these (or any garbage), we treat them as verbatim content
so that we can handle it as nondestructively as possible.

This change also makes some small fixes to verbatim handling, and
removes a convenience constructor for verbatim tokens so that we force
the caller to specify the indentation behavior instead of relying on an
arbitrary default.


allevato:
cc @dylansturg

Tests/SwiftFormatPrettyPrintTests/GarbageTextTests.swift
```diff
@@ -0,0 +1,147 @@
...
+      """
+
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 21)
+  }
+
+  func testUnknownCodePointAsTrailingTriviaAreGluedToPreviousToken() {
+    // Note: The third line here is a bit of a weird case. Despite the operator being separated from
+    // the left-hand-side by a space, the intervening garbage text appears to put the parser in a
+    // mode where it treats the operator as a postfix operator, which then causes `secondTerm` to be
+    // parsed as its own distinct statement and `CodeBlockItem`---hence the lack of indentation.
```
dylansturg: Wow, what a weird case. Is that worth a bug for the parser?  I'm not sure if it can be fixed, but I would expect any text identified as `garbageText` to not impact parsing.

Tests/SwiftFormatPrettyPrintTests/GarbageTextTests.swift
```diff
@@ -0,0 +1,147 @@
...
+      print("!!")
+      }
+      """
+
+    let expected =
+      """
+      func greet() {
+        print("Hello")
+      <<<<<<< hash:filename
+      print("werld")
```
dylansturg: Should this line and the next non-garbage line be indented to match the scope? Or are these lines also treated as `garbageText`?

allevato: Everything from the opening conflict marker to the closing marker is lexed as part of the garbage text, so we don't have a way to format what's inside either branch.

Tests/SwiftFormatPrettyPrintTests/GarbageTextTests.swift
allevato: Yeah, I don't know if garbage text affects the trivia in a way that causes it to ignore _all_ of the other nearby trivia for the purposes of associating the operator with the thing before it, or what's going on 🤷‍♂

PR #129, opened 2020-01-29
Update swift-format to new SwiftSyntax API

ahoppen:
Update swift-format to the updated SwiftSyntax API.

Some rough measurements on my local machine show that this speeds formatting up by ~2x (measured on formatting the `swift-format` source code).


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1832,25 +1824,25 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
     return .visitChildren
   }
 
-  func visit(_ node: AccessPathComponentSyntax) -> SyntaxVisitorContinueKind {
+  override func visit(_ node: AccessPathComponentSyntax) -> SyntaxVisitorContinueKind {
     return .visitChildren
   }
 
   // FIXME: Remove once the changes around `GenericRequirementSyntax` and its children have settled.
   #if !HAS_UNCONSOLIDATED_GENERIC_REQUIREMENTS
```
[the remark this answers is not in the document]
ahoppen: Oh, yes these aren’t necessary anymore. I removed the checks and/or their blocks. I also checked for more `#if` but seems like there aren't any more of these checks.

Package.swift
```diff
@@ -21,10 +21,7 @@ let package = Package(
     .library(name: "SwiftFormatConfiguration", targets: ["SwiftFormatConfiguration"]),
   ],
   dependencies: [
-    .package(
-      url: "https://github.com/apple/swift-syntax",
-      .revision("swift-DEVELOPMENT-SNAPSHOT-2019-09-26-m")
-    ),
+    .package(url: "https://github.com/apple/swift-syntax", .branch("master")),
```
akyrtzi: If swift-format depends on master we won't know if we break it inadvertently after some SwiftSyntax change that gets in master, because swift-format does not participate in Swift CI currently. Maybe we should stick to using tags but switch to the most recent one?
@allevato what do you think?

allevato: It looks like the most recent tag of swift-syntax is the one we're currently using, `swift-DEVELOPMENT-SNAPSHOT-2019-09-26-m`. Can you tag the current state of `master` for the latest dev snapshot and we can switch to that for the time being? Then once we have the rest of the work done to get swift-format on CI, we can flip it over to `master`.

akyrtzi: @shahmishal could you enable auto-tagging in the SwiftSyntax repo please? 🙏

shahmishal: Updated, we should see tag on next nightly release.

allevato: Thanks @shahmishal! It looks like `swift-DEVELOPMENT-SNAPSHOT-2020-01-29-a` was tagged last night, so let's switch to that one for now.

Could you also update the table in [README.md](https://github.com/apple/swift-format/blob/master/README.md) to refer to the new snapshot?

ahoppen: Sounds good to me. `swift-format` now uses the `swift-DEVELOPMENT-SNAPSHOT-2020-01-29-a`. 

In the readme, I just replaced `swift-DEVELOPMENT-SNAPSHOT-2019-09-26-m` with `swift-DEVELOPMENT-SNAPSHOT-2020-01-29-a` since there is no branch that would refer to the old snapshot.

allevato:
Alright, with those changes in I think we're good to go. Thanks!

PR #132, opened 2020-01-30
Wrap closure params in open/close breaks to indent inside parens.

dylansturg:
Closures have an `input` clause which can either be parenthesized or not. When the clause is parenthesized, it's a different type of syntax. We want the formatter to create extra indentation inside of the parens in that case, just like it would for function call argument lists. To do this, I have added an open and a close break inside of the parens to create a block scope around the closure parameters.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -889,13 +889,6 @@ private final class TokenStreamCreator: SyntaxVisitor {
   }
 
   override func visit(_ node: ClosureSignatureSyntax) -> SyntaxVisitorContinueKind {
-    let consistency: GroupBreakStyle
-    if let input = node.input, input.is(ClosureParamListSyntax.self) {
-      consistency = argumentListConsistency()
-    } else {
-      consistency = .inconsistent
-    }
-
     before(node.firstToken, tokens: .open)
 
     if let input = node.input {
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -889,22 +889,33 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
       // after the capture list's right bracket if there are arguments following it or we'll end up
       // with an extra space if the line doesn't wrap.
       after(node.capture?.rightSquare, tokens: .break(.same))
-      before(input.firstToken, tokens: .open(consistency))
-      after(input.lastToken, tokens: .close)
+
+      // When it's parenthesized, the input is a `ParameterClauseSyntax`. Otherwise, it's a
+      // `ClosureParamListSyntax`. The parenthesized version is wrapped in open/close breaks so that
+      // the parens create an extra level of indentation.
+      if let parameterClause = input.as(ParameterClauseSyntax.self) {
```
allevato: I just realized there are a couple things we need to be careful about here. When we added the `prioritizeKeepingFunctionOutputTogether` setting, we had to split where the open/close tokens were added due to visitation order. So if a `ParameterClause` can occur here and that setting is enabled, we won't compensate for that here.

Can you add a few tests with `prioritizeKeepingFunctionOutputTogether` that cover various combinations of signatures with throws/return types and fix up what needs to be handled for that? A test or two with an empty argument list (`{ () in ... }`) would be good too for completeness.

I don't know if we care about changing the behavior in the closure case for that setting. I guess we _could_ choose between this (`prioritizeKeepingFunctionOutputTogether = false`):

```swift
let x = {
  (foo: Int, bar: Int) throws
    -> Int in
  statements
}
```

and this (`prioritizeKeepingFunctionOutputTogether = true`):

```swift
let x = {
  (
    foo: Int,
    bar: Int
  ) throws -> Int in
  statements
}
```

Or we could always go with the second one. I'm not sure off the top of my head what we do today.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -889,22 +889,33 @@ private final class TokenStreamCreator: SyntaxVisitor {
...
+        // Group outside of the parens, so that the entire argument list stays together if possible.
+        before(input.firstToken, tokens: .open)
+        after(input.lastToken, tokens: .close)
+
+        // Group and break inside of the parens to create a scope for the argument list, similar to
+        // a function call argument list.
+        let consistency = argumentListConsistency()
+        after(parameterClause.leftParen, tokens: .break(.open, size: 0), .open(consistency))
+        before(parameterClause.rightParen, tokens: .break(.close, size: 0), .close)
```
allevato: I think these three lines can be replaced by a call to `arrangeParameterClause` with `forcesBreakBeforeRightParen: true`.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
dylansturg: So it turns that today using `prioritizeKeepingFunctionOutputTogether = true` crashes for closures with parenthesized parameter lists (i.e. `ParameterClauseSyntax`). Since it is very "function-like", I think respecting that configuration makes sense.

I had to rework the groups around the parameter clause, since the group around the output can't coexist with a group around the parens.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift

PR #134, opened 2020-01-31
Format differentiation attributes.

allevato:
This change adds correct formatting for `@differentiable`,
`@derivative`, and `@transpose`.

Fixes SR-12002.


allevato:
cc @dylansturg @dan-zheng

dan-zheng:
Orthogonal question for @allevato: I recall we discussed fallback logic for parsing `UnknownSyntax`. This could help `swift-format` handle unknown or newly-added syntax kinds, e.g. `DifferentiableAttributeArgumentsSyntax`.

I wonder if that's still a good idea?

allevato:
> Orthogonal question for @allevato: I recall we discussed fallback logic for parsing `UnknownSyntax`. This could help `swift-format` handle unknown or newly-added syntax kinds, e.g. `DifferentiableAttributeArgumentsSyntax`.
> 
> I wonder if that's still a good idea?

Yeah, I'm working on a follow-up PR to flesh out the rest of our `Unknown*` handling.

PR #135, opened 2020-01-31
Flesh out unknown node handling.

allevato:
We previously only covered a subset of the unknown nodes. Treat them
all as verbatim text so that malformed or unrecognized syntax does
not get completely clobbered. (Note that in some cases, depending on
how breaks/spaces are inserted into the token stream relative to tokens
in the unknown nodes, leading spaces may still get compressed).


allevato:
cc @dylansturg

PR #136, opened 2020-02-02
Various visibility (and other) cleanup.

allevato:
This change cleans up visibility throughout the code, with the main objectives being:

- Make things not `public` if they don't need to be (e.g., tests).
- Make some `internal` things `private` or `fileprivate` if they don't need to be accessed outside the scope/file.
- Use `fileprivate` at file scope instead of `private`.
- Remove all `@testable` imports, making some `internal` things `public` to support that change (needed to add swift-format to CI).

This also deletes some unused code, mostly helpers that were written to work with nodes/trivia in the phase 1 rules that are no longer needed.


allevato:
cc @dylansturg

PR #137, opened 2020-02-03
Add `FileprivateAtFileScope` rule.

allevato:
This rule checks file-scoped declarations and diagnoses them if
they are declared `private` instead of `fileprivate` (and fixes
them in format mode).


allevato:
cc @dylansturg

PR #139, opened 2020-02-04
Group the closing paren of exprs into the enclosed open/close breaks.

dylansturg:
The previous behavior kept line breaks before the right paren, but discarded line breaks after the left paren. This caused the formatter to create "unbalanced" parenthesized expressions where the left paren was glued to the first token in an expression but there could be a line break before the right paren.

Instead, both the left and right parens are now glued to tokens in the parenthesized expression.


allevato:
What happens if parenthesized expressions are nested so that multiple closing parens occur together? Does this keep them all glued together, or does it only handle the one immediately following?

dylansturg:
That's a good point. Nested paren expressions didn't work very well. I have updated the change to walk up consecutive right parens and absorb all of the consecutive right parens.

PR #140, opened 2020-02-06
Expose finding ".swift-format" file

Trzyipolkostkicukru:
Expose finding ".swift-format" file, and creating Configuration with that file, so that it could be used with sourcekit-lsp

I hope it's okay if there are no tests for this functionality in here, and instead it's tested indirectly by the sourcekit-lsp with an awesome Tibs that makes testing multi file projects easy

### DIFF
--- a/Sources/SwiftFormatConfiguration/Configuration.swift
+++ b/Sources/SwiftFormatConfiguration/Configuration.swift
@@ -120,6 +120,12 @@ public struct Configuration: Codable, Equatable {
     self.version = highestSupportedConfigurationVersion
   }
 
+  /// Constructs a Configuration using values from a JSON config file
+  public init(configFile url: URL) throws {
+    let data = try Data(contentsOf: url)
+    self = try JSONDecoder().decode(Configuration.self, from: data)
+  }
+
   public init(from decoder: Decoder) throws {
     let container = try decoder.container(keyedBy: CodingKeys.self)
 
@@ -188,6 +194,25 @@ public struct Configuration: Codable, Equatable {
       lineBreakAroundMultilineExpressionChainComponents,
       forKey: .lineBreakAroundMultilineExpressionChainComponents)
     try container.encode(rules, forKey: .rules)
+  }
+
+  /// Configuration file associated with a swift file.
+  /// 
+  /// Looks for a ".swift-format" file in the same directory as the swift file, or its nearest parent.
+  /// If one is not found, returns "nil".
+  public static func configurationFile(forSwiftFile url: URL) -> URL? {
+    var path = url.absoluteURL
+    let configFilename = ".swift-format"
+
+    repeat {
+      path = path.deletingLastPathComponent()
+      let candidateFile = path.appendingPathComponent(configFilename)
+      if FileManager.default.isReadableFile(atPath: candidateFile.path) {
+        return candidateFile
+      }
+    } while path.path != "/"
+
+    return nil
   }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -120,6 +120,12 @@ public struct Configuration: Codable, Equatable {
     self.version = highestSupportedConfigurationVersion
   }
 
+  /// Constructs a Configuration using values from a JSON config file
```
allevato: Let's leave the format of the file unspecified in the doc comment, so that we have the ability to change it in the future (while still retaining support for older formats):

```
/// Constructs a Configuration by loading it from a configuration file.
```

Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -120,6 +120,12 @@ public struct Configuration: Codable, Equatable {
...
+  public init(configFile url: URL) throws {
```
allevato: To match the naming convention of types like `Data` that load URLs, let's call this `init(contentsOf url: URL) throws`.

Sources/SwiftFormatConfiguration/Configuration.swift
```diff
@@ -189,6 +195,25 @@ public struct Configuration: Codable, Equatable {
       forKey: .lineBreakAroundMultilineExpressionChainComponents)
     try container.encode(rules, forKey: .rules)
   }
+
+  /// Configuration file associated with a swift file.
+  /// 
+  /// Looks for a ".swift-format" file in the same directory as the swift file, or its nearest parent.
+  /// If one is not found, returns "nil".
+  public static func configurationFile(forSwiftFile url: URL) -> URL? {
```
allevato: If we're generalizing this so that other people can call it, we don't have to require that the input URL point specifically to a Swift file. For example, we might want to allow someone to pass the URL of a directory to ask "what configuration file applies to files in this directory?"

With that in mind, we could name this `func url(forConfigurationFileApplyingTo url: URL) -> URL?`, and update the doc comment accordingly, such as

```
/// Returns the URL of the configuration file that applies to the given file or directory.
```

I think we can be vague about the exact search behavior in the doc comment because it will give us the flexibility to adjust it in the future as well.

But we'll also need to update the implementation below: before the loop, we'll need to check to see if the incoming URL is a file (rather than a directory), and if it is, delete the file component from the path. Then we'll also need to adjust the first line of the loop and the condition slightly to make sure that we still go up to the parent at the right time and exit correctly when we hit the root.

Sources/swift-format/main.swift
```diff
@@ -89,53 +89,28 @@ fileprivate func loadConfiguration(
...
-  let configFilename = ".swift-format"
-
-  repeat {
-    path = path.deletingLastPathComponent()
-    let testPath = path.appendingPathComponent(configFilename).path
-    if FileManager.default.isReadableFile(atPath: testPath) {
-      return testPath
-    }
-  } while path.path != "/"
+  if let swiftFileUrl = swiftFilePath.map(URL.init(fileURLWithPath:)),
```
allevato: Please use `swiftFileURL` and `configFileURL` throughout.

Sources/SwiftFormatConfiguration/Configuration.swift
Trzyipolkostkicukru: I added a placeholder component for files instead of adjusting the loop condition, because otherwise it would be hard to write it in a way that doesn't skip the root directory

PR #142, opened 2020-02-07
Factor out common DiagnosingTestCase support.

allevato:
Over time we've built up three slightly different versions of the
same `DiagnostingTestCase` class in different test targets. This
change factors out all the commonalities into a separate test
support target and has the other tests now subclass it.

On platforms that support it (macOS 10.15+, and Linux), formatting
tests now use `CollectionDifference` to show a line-based diff of
the expected/actual output, which is a significant readability
improvement. For example:

```
failed - Pretty-printed result was not what was expected - Actual output (+) differed from expected output (-):
 func name<R>(
   _ x: Int
 ) throws -> R
-where
-  Fooooooo == Barrrrr
+  where
+    Fooooooo == Barrrrr
```


allevato:
cc @dylansturg

allevato:
To clarify from the original description, when it says "formatting tests  now use `CollectionDifference`", that's both pretty printer tests *and* phase 1 rule tests—everything goes through the same comparison logic now.

Tests/SwiftFormatWhitespaceLinterTests/WhitespaceTestCase.swift
```diff
@@ -1,54 +1,23 @@
...
-  }
-
-  override func tearDown() {
-    guard shouldCheckForUnassertedDiagnostics else { return }
-
-    for diag in consumer.registeredDiagnostics {
-      XCTFail("unexpected diagnostic '\(diag)' emitted")
-    }
+class WhitespaceTestCase: DiagnosingTestCase {
+  override func setUp() {
```
dylansturg: Should this call `super.setUp`? Any chance of `DiagnosingTestCase.setUp` containing some code that needs to be executed?

allevato: Yes! Even though all we do here is change the value of `shouldCheckForUnassertedDiagnostics`, if we ever add more logic in the base class we'll want to make sure it's not skipped.

Sources/SwiftFormatTestSupport/DiagnosingTestCase.swift
```diff
@@ -0,0 +1,170 @@
+import SwiftFormatConfiguration
+import SwiftFormatCore
+import SwiftSyntax
+import XCTest
+
+/// DiagnosingTestCase is an XCTestCase subclass meant to inject diagnostic-specific testing
+/// routines into specific formatting test cases.
+open class DiagnosingTestCase: XCTestCase {
+  /// Set during lint tests to indicate that we should check for any unasserted diagnostics when the
```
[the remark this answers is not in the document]
allevato: Done.

PR #143, opened 2020-02-08
Ignore discretionary newlines in a number of new locations.

allevato:
- After the colon in...
  - ...a type annotation (`x: Int`)
  - ...a tuple type (`(a: A, b: B) -> Void`)
  - ...a tuple expression element (`(x: 10, y: 20)`)
  - ...a dictionary type (`[Key: Value]`)
  - ...a dictionary element (`["key": "value"]`)
  - ...a function call argument (`foo(x: 10, y: 20)`)
  - ...a function declaration parameter (`func foo(x: Int)`)
  - ...a generic parameter (`struct A<T: P> {}`)
  - ...an availability attribute (`@attribute(renamed: "foo")`)
- Between the first and second name of a parameter in a function
  declaration (`func foo(firstName secondName: Type)`)
- After the specifier in an attributed type (`inout Foo`)

We still allow wrapping in all of these locations, but only if the
formatter determines that it is necessary to make the contents fit.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -510,7 +510,9 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
     after(node.inKeyword, tokens: .space)
 
     if let typeAnnotation = node.typeAnnotation {
-      after(typeAnnotation.colon, tokens: .break(.open(kind: .continuation)))
+      after(
+        typeAnnotation.colon,
+        tokens: .break(.open(kind: .continuation), newlines: .elective(ignoresDiscretionary: true)))
       after(typeAnnotation.lastToken, tokens: .break(.close(mustBreak: false), size: 0))
     }
 
@@ -695,7 +697,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   /// - Parameter node: The tuple expression element to be arranged.
   private func arrangeAsTupleExprElement(_ node: TupleExprElementSyntax) {
     before(node.firstToken, tokens: .open)
-    after(node.colon, tokens: .break)
+    after(node.colon, tokens: .break(.continue, newlines: .elective(ignoresDiscretionary: true)))
     after(node.lastToken, tokens: .close)
   }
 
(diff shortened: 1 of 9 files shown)


allevato:
cc @dylansturg

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1240,9 +1248,10 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
   override func visit(_ node: TupleTypeElementSyntax) -> SyntaxVisitorContinueKind {
     before(node.firstToken, tokens: .open)
-    after(node.colon, tokens: .break)
-    after(node.inOut, tokens: .break)
```
dylansturg: Do we need this break after the `inOut` token?

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1303,7 +1312,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
   override func visit(_ node: AvailabilityLabeledArgumentSyntax) -> SyntaxVisitorContinueKind {
     before(node.label, tokens: .open)
-    after(node.colon, tokens: .break(.continue, size: 1))
+    after(node.colon, tokens: .break(.continue, newlines: .elective(ignoresDiscretionary: true)))
```
dylansturg: This break had a `size` of 1 before, to create a space. Do we still want that space when it doesn't fire?

allevato: The default size for a break is 1 (https://github.com/apple/swift-format/blob/master/Sources/SwiftFormatPrettyPrint/Token.swift#L210), so I was cleaning up the unnecessary specification while editing here.

We probably ought to do that throughout at some point. I also think we ought to make the break kind be explicit instead of defaulting to `.continue`.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
allevato: The `inOut` token here is legacy, I believe, from when the attribute used to come before the variable name instead of the type (e.g., `inout foo: Bar` instead of `foo: inout Bar`). The parser doesn't accept that syntax anymore (I tried it in the syntax explorer), so I don't think this will ever be non-nil. When using the new syntax (`foo: inout Bar`), the type comes across as an `AttributedTypeSyntax`, and we add a break after the `specifier` (which might be `inout`, `__shared`, etc.).

PR #148, opened 2020-02-24
Don't apply `FileprivateAtFileScope` rule to extensions.

allevato:
This is a subtlety of SE-0169. For most decls, `private` at file
scope is equivalent to `fileprivate`. But for extensions, since that
visibility is distributed throughout the extension's members,
changing `private` to `fileprivate` would increase the visibility of
those members. When `private`, those members would only be visible
within the same extension scope and within other extensions to the
same type and to the main type decl itself *that occur within the
same file.* If they were made `fileprivate`, then this would make
them visible to *all* declarations within the same file.


allevato:
cc @dylansturg

PR #151, opened 2020-02-25
Absorb trailing comma of function call args in grouped exprs.

dylansturg:
Previously, these trailing commas could be placed on a new line after the expression because the close break after the expression would fire if the comma landed exactly 1 over the column limit. Now the comma is moved inside of the group and breaks around the expr so that a break inside of the expr fires instead of the close break.


allevato:
Are there other trailing commas that pose similar problems that we should knock out while we're in here? For example, array/dictionary/tuple literals with element expressions ending in a close break?

dylansturg:
> Are there other trailing commas that pose similar problems that we should knock out while we're in here? For example, array/dictionary/tuple literals with element expressions ending in a close break?

Good point - I reproduced the issue in arrays, dictionaries, tuples and pattern binding. I updated `TokenStreamCreator` to handle those trailing commas too.

PR #152, opened 2020-02-26
Make the `FileprivateAtFileScope` rule configurable.

allevato:
The rule has now been named `FileScopedDeclarationPrivacy` and supports
making effectively-private file-scoped declarations *either* `fileprivate`
*or* `private`, instead of forcing `fileprivate`. This is controlled using
a configuration key (which now defaults to `private`, based on user
feedback).


allevato:
cc @dylansturg

PR #154, opened 2020-02-27
Adopt ArgumentParser for the command-line tool

natecook1000:
This adopts the new `ArgumentParser` library for parsing command-line arguments.


Package.swift
```diff
@@ -23,9 +23,10 @@ let package = Package(
   dependencies: [
     .package(
       url: "https://github.com/apple/swift-syntax",
-      .revision("swift-DEVELOPMENT-SNAPSHOT-2020-01-29-a")
+      .branch("master")
```
fernandodev: better to use a specific branch, tag or commit

fernandodev: such as `swift-5.2-branch`

Sources/swift-format/CommandLineOptions.swift
```diff
@@ -10,191 +10,115 @@
...
-    try binder.fill(parseResult: args, into: &opts)
-
-    if opts.inPlace && (ToolMode.format != opts.mode || opts.paths.isEmpty) {
-      throw ArgumentParserError.unexpectedArgument("--in-place, -i")
+  mutating func validate() throws {
+    if version {
+      throw CleanExit.message("0.0.1")
+    }
+    
+    if inPlace && (mode == .format || paths.isEmpty) {
```
allevato: It looks like part of this got flipped; it should be `mode != .format`.

Sources/swift-format/Run.swift
```diff
@@ -29,38 +29,34 @@ import TSCBasic
...
+  let diagnosticEngine = makeDiagnosticEngine()
   let linter = SwiftLinter(configuration: configuration, diagnosticEngine: diagnosticEngine)
   linter.debugOptions = debugOptions
   let assumingFileURL = URL(fileURLWithPath: assumingFilename ?? "<stdin>")
 
   guard let source = readSource(from: sourceFile) else {
-    diagnosticEngine.diagnose(
-      Diagnostic.Message(.error, "Unable to read source for linting from \(assumingFileURL.path)."))
-    return 1
+    throw FormatError.readSource(path: assumingFileURL.path)
```
allevato: In https://github.com/apple/swift-format/pull/144, @dylansturg refactored `lintMain` and `formatMain` a bit so that the error messages here were sent through the diagnostic engine (instead of being written directly to `stderr`)—this unified all possible error messages that the tool might emit after argument parsing. For example, we could add a flag later to replace the diagnostic consumer with one that serializes to JSON instead of printing them to `stderr`, and file I/O errors would be handled seamlessly.

Changing these to throw errors here would revert that back to having two different kinds of behavior for different kinds of errors. Do we need this change?

Sources/swift-format/main.swift
```diff
@@ -16,58 +16,46 @@ import SwiftFormatConfiguration
...
-  let diagnosticEngine = makeDiagnosticEngine()
-  switch options.mode {
-  case .format:
-    if options.paths.isEmpty {
-      let configuration = loadConfiguration(
-        forSwiftFile: nil, configFilePath: options.configurationPath)
-      return Int32(
-        formatMain(
+extension SwiftFormatCommand {
+  func run() throws {
```
allevato: Since `ArgumentParser` has subcommand support, it would make sense to have `format`, `lint`, and `dump-configuration` be subcommands with `format` as the default subcommand. It looks like that would also improve the command line validation because options could be restricted to just the commands that they apply to. **But** we also have internal tooling that depends on the current command line interface, so we wouldn't want to do that breaking change here yet.

As a possible follow-up PR, does `ArgumentParser`'s design let us easily create those subcommands and then have the old `-m/--mode` flag delegate to them as a migration path so that we could have both interfaces implemented simultaneously, and then once we have our other tooling no longer depending on the old interface, we could remove it from `swift-format`?

Sources/swift-format/Run.swift
natecook1000: I'll definitely revert that — it looked like I was replicating the current usage, but of course coalescing errors to look forward at using other outputs makes tons of sense.

Sources/swift-format/main.swift
natecook1000: We should totally be able to have those parallel interfaces. I have an old version of `swift-format` migrated to a subcommand interface, so after this is ready to merge I'll update that and open another PR.

Sources/swift-format/Run.swift
```diff
@@ -76,7 +75,7 @@ func lintMain(
 func formatMain(
   configuration: Configuration, sourceFile: FileHandle, assumingFilename: String?, inPlace: Bool,
   debugOptions: DebugOptions, diagnosticEngine: DiagnosticEngine
-) -> Int {
+) {
   let formatter = SwiftFormatter(configuration: configuration, diagnosticEngine: diagnosticEngine)
```
dylansturg: I recently changed this to pass nil, so that `DiagnosticEngine` is only used for fatal messages when formatting, in #155. It looks like this is causing a merge conflict - can you update and resolve the conflict?

natecook1000:
> @dylansturg:
> Is there a way to have a nonzero exit status and not print that extra "Error"? 

Good point. This looks like something we'll want to add to `ArgumentParser`.

allevato:
> Good point. This looks like something we'll want to add to `ArgumentParser`.

Is this something you'd add in the near future or later? If it's soon we could just wait for it, but if it's later, I don't want to block the rest of this change on it. @dylansturg suggested offline that we could change the `FormatError` message to something like "Exiting due to previous errors." The extra message isn't ideal but at least it would be temporary, and you could tag the message with a `TODO` comment to replace it once we have a way to silently exit with a non-zero status.

natecook1000:
I'll add it today, so it will be in the `0.0.2` release. We can just leave out the `Error: ` prefix if there's nothing to show afterward.

natecook1000:
Updated this to use the new `ArgumentParser` functionality that allows exiting with a failure code without adding to the stderr output.

PR #160, opened 2020-03-10
Add back the short forms of `--mode` (`-m`) and `--version` (`-v`).

allevato:
Eventually we'll want to transition these flags to be subcommands
instead, but for the time being, these short names were accidentally
dropped in the migration to `ArgumentParser`, so put them back for
now.


allevato:
No worries, I should have caught it! But I mostly run `swift-format` in default mode and rarely use the flag. Clearly we need better end-to-end tests 🙂

PR #161, opened 2020-03-17
Migrate modes to subcommands.

allevato:
This change refactors the command line interface's
modes from using a flag (`-m/--mode`) to using
swift-argument-parser subcommands instead. This
provides a much better separation of responsbilities
because subcommands only define arguments that they
support, rather than having a single command that
supports the union of all possible arguments.

The old mode flag is still supported for backward
compability (see `LegacyMainCommand`) in order to
give clients who may be using it in automation an
opportunity to migrate away from it, but it
will be removed in the future.


allevato:
@natecook1000 I've factored out the `format`, `lint`, `dump-configuration`, and `version` "modes" into separate subcommands and I really like the structure a lot better than the way we used to have it.

I tried adding `LegacyMainCommand` as a temporary default subcommand to keep the `-m` flag working for a brief time. But I'm seeing some weird behavior:

```shell
# works
$ .build/debug/swift-format --mode=dump-configuration
{
  "fileScopedDeclarationPrivacy" : {
  ...

# doesn't work
$ .build/debug/swift-format --mode dump-configuration
Error: Unknown option '--mode'
Usage: swift-format dump-configuration

# doesn't work
$ .build/debug/swift-format -m dump-configuration
Error: Unknown option '-m'
Usage: swift-format dump-configuration
```

It looks like the value of the `-m/--mode` flag is being interpreted as a subcommand name and then it's trying to parse `-m/--mode` as a flag for that other subcommand (which is why it only works when it's joined via `=`), when I would expect that the subcommand would be defaulted if it wasn't the first argument following the executable.

I guess this is meant to support more widely applicable flags between the executable name and the subcommand name, then? Is there a way to restructure my commands to make this work?

natecook1000:
Awesome — looks really good so far!

Sounds like there’s a bug in how we handle default subcommands. I’ll see if I can fix this before the `0.0.3` release. There should be a workaround: put the legacy options directly on the root command, and throw a `CleanExit.helpRequest` if there’s no mode specified. Unfortunately, there’s a different bug ([already fixed](https://github.com/apple/swift-argument-parser/pull/78) in unreleased `master`) that will cause its own problems with that approach.

allevato:
Thanks! In that case, this isn't particularly urgent—I'll file an issue to track this on the swift-argument-parser repository and then pick this back up once it's fixed.

natecook1000:
@allevato `0.0.3` is now tagged with a fix for this issue!

allevato:
Thanks! That did the trick and this is ready for review now. Can you take a look and make sure I'm using subcommands and common arguments the way they're intended?

allevato:
I'm happy with both of those suggestions! After removing the `Command` suffix I also nested the subcommand types inside the main `SwiftFormatCommand` type; it felt redundant doing that before, but now it makes more sense with the bare names. (I left `SwiftFormatCommand` with the suffix to avoid a type and an imported module with the same name.)

Factoring out the `--version` flag into a completely separate `ParsableArguments` type that encapsulates the logic is another nice idea so that I don't have to handle it in the root command. I was trying to think of an elegant way of doing that if it wasn't a subcommand, and that works really well.

Thanks!

PR #164, opened 2020-03-26
Speed up WhitespaceLinter by ~200x.

allevato:
This PR is split into 4 commits to make the individual changes easier to follow:

1. Adds a performance test. On my iMac Pro, this averaged a runtime of about **0.867s** in the original implementation (for the 300 line source sample). That includes time in SwiftSyntax to parse, then to run the pretty printer, and finally to run the whitespace linter. Most of the time was in the linter.
2. Update the linter to convert the user and formatted text into arrays of UTF-8 code units and iterate over those, instead of iterating over characters or using any string APIs. This by itself sped things up by about 40x.
3. Remove our custom code for computing the line/column based on UTF-8 offset. This was accidentally starting from the beginning of the string every time it needed to be computed; now we use the `SourceLocationConverter` that SwiftSyntax already computed.
4. Removed computationally expensive uses of `Collection.split` (causing a lot of heap allocations) by lazily computing whitespace runs. This required merging a lot of the logic in the individual check functions into `compareWhitespace` so it's a bit less easy to follow, but I tried commenting to make up for that.

Overall, this reduced the runtime of the performance test on average to about **0.004s**.


allevato:
cc @dylansturg

SDGGiesbrecht:
🎉

PR #168, opened 2020-04-01
Remove swift-tools-support-core dependency.

allevato:
TSC was mostly used for its argument parser. Now that
we've migrated to swift-argument-parser, the remaining
file system and path APIs in TSC are only used by the
Format command and can be fairly easily replaced by
Foundation APIs instead.


allevato:
cc @dylansturg

PR #170, opened 2020-04-01
Move conditional compilation directive inside tests.

allevato:
This ensures that the test function always exists, even if the
symbol is not defined, so that `--generate-linuxmain` contains
the symbol in both cases. When the symbol is undefined, the
empty test will vacuously succeed.


allevato:
cc @dylansturg

PR #173, opened 2020-04-02
Remove calls to `FileHandle.synchronizeFile()`.

allevato:
Standard output does not support `fsync`, which is the system
call that this boils down to on Linux (and presumably on Darwin
as well). In both cases, `fsync` sets `errno` to `EINVAL`, but
unfortunately Foundation's behavior differs per platform: on
Darwin the error is silently swallowed, but on Linux it traps.


allevato:
cc: @dylansturg

PR #174, opened 2020-04-02
Fix `@differentiable` attribute formatting.

allevato:
- Don't insert a space before `where` if it's the first thing
  in the attribute.
- Apply formatting to comma-delimited parameter "tuples"
  following the `wrt:` clause.

Fixes SR-12414.


allevato:
cc @dylansturg

PR #178, opened 2020-04-06
Refactor main executable into a frontend architecture.

allevato:
Instead of a bunch of free functions that pass their arguments
around, this change now refactors the core format and lint file
handling functionality into a set of "frontend" classes so that
the diagnostic engine and related state can be created in one
place and shared.

This also significantly improves error handling. Whereas we
were previously killing the process on the first sign of a
source file or configuration file being invalid, we now emit
an error and skip the file, allowing any remaining inputs to
still be processed.

Lastly, this introduces a configuration cache so that we're
not re-reading the configuration from the file system for
every source file we process. In most cases, when we're
processing a recursive directory structure, they'll all
share a configuration file at a common root, so we cache
that based on its file URL and return it when requested.


allevato:
cc @dylansturg

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -0,0 +1,202 @@
...
+      path: lintFormatOptions.assumeFilename ?? "<stdin>",
+      configuration: configuration)
+    processFile(fileToProcess)
+  }
+
+  /// Processes source content from a list of files and/or directories provided as paths.
+  private func processPaths(_ paths: [String]) {
+    precondition(
+      !paths.isEmpty,
+      "processFiles() should only be called when paths is non-empty.")
```
dylansturg, line 130: Should this be `processPaths`?

allevato: It should be `processPaths(_:)` 😛  That's what I get for trying to have a helpful sanity check.

PR #180, opened 2020-04-07
Recursively apply NoEmptyTrailingClosureParentheses rule.

dylansturg:
This rule wasn't being applied recursively, which meant it wouldn't be applied to the trailing closure of a function call where parens where removed.


Sources/SwiftFormatRules/NoEmptyTrailingClosureParentheses.swift
```diff
@@ -37,7 +39,9 @@ public final class NoEmptyTrailingClosureParentheses: SyntaxFormatRule {
       on: node.calledExpression,
```
allevato, line 39: I wonder if we need to recursively visit `node.calledExpression` as well.

I think this test would catch it:

```swift
foo(bar() { baz })() { blah }
```

`node.calledExpression` should be `foo(bar() { baz })`, so if the parens after `bar` don't get dropped, then we need to call it there too.

dylansturg: Good call. Added.

PR #181, opened 2020-04-07
Rearrange breaks so class decls won't overflow line length.

dylansturg:
The inheritance list for type and extension decls is wrapped with open/close tokens, but the relevant open-break token was nested inside of a different group. The previous token layout created the correct indentation of the tokens in the inheritance list, but meant the break between the colon and first token in the inheritance list could never fire because it was followed by a close token.

This resulted in type and extension decls that overflow the column limit when the type name + `:` + first token in the inhertance list together was longer than the column limit.

I fixed this by moving the open break inside of the open/close tokens surrounding the inheritance list. Normally, the open break comes before the open but that would force a break before the first token in the inheritance list if the entire list doesn't fit. That behavior wouldn't be consistent with existing behavior. Instead, placing the open break inside of the open only breaks if the first token in the inheritance list is too long.

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/ExtensionDeclTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/ExtensionDeclTests.swift
@@ -316,8 +316,8 @@ final class ExtensionDeclTests: PrettyPrintTestCase {
     let expected =
 
       """
-      public extension MyContainer: MyContainerProtocolOne,
-        MyContainerProtocolTwo,
+      public extension MyContainer:
+        MyContainerProtocolOne, MyContainerProtocolTwo,
         SomeoneElsesContainerProtocol,
         SomeFrameworkContainerProtocol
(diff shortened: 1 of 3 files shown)


Tests/SwiftFormatPrettyPrintTests/ExtensionDeclTests.swift
```diff
@@ -316,8 +316,8 @@ final class ExtensionDeclTests: PrettyPrintTestCase {
     let expected =
 
       """
-      public extension MyContainer: MyContainerProtocolOne,
-        MyContainerProtocolTwo,
+      public extension MyContainer:
```
allevato, line 319: Not in this PR, but we should have an internal sanity check that none of the expected output lines are longer than the given line length (with an override for comments/string literals, I guess), since that would have caught this mistake—53 columns but line length of 50.

PR #183, opened 2020-04-08
Minor updates to `@differentiable`.

allevato:
https://github.com/apple/swift/pull/30001 removed the logic
from the parser to handle the `jvp:` and `vjp:` arguments of
the attribute (but left the syntax definitions in place for
the time being). Since this causes an attribute with those
arguments to fail to parse, I've removed them from the tests
so that those tests continue to pass under the new behavior.
(The arguments were always optional, so they pass under the
old behavior as well.)

This also caught and fixed a bug where an attribute with
only a `wrt:` and a `where` clause wasn't getting formatted
correctly.


allevato:
cc @dylansturg

PR #184, opened 2020-04-08
Add breaks around conditions of while-stmt.

dylansturg:
This aligns the behavior of while-stmt and guard-stmt conditions. The while-stmt conditions didn't have open-breaks, which meant there was no additional indentation nor would continuation breaks be able to stack.

This means there's always a break before the first token of the condition element list when the while-stmt's condition is multiple lines. That is likely to cause some churn in existing code. An alternative would be to treat this like if-stmt, where there is no break after the while token.

## Issues
- This is likely to cause churn. I think we could minimize that by aligning with the if-stmt behavior, but I think it makes more sense to allow a break after `while`, like `guard`.


allevato:
To avoid churn, can we align with `if` for now, then once we have region-restricted formatting in the future, change it to have the break after the `while`?

dylansturg:
> To avoid churn, can we align with `if` for now, then once we have region-restricted formatting in the future, change it to have the break after the `while`?

I think that's fair. Made the change.

PR #185, opened 2020-04-09
Add a consistent group around if-stmts.

dylansturg:
The consistent group forces the bodies of if-stmts with else clauses to have breaks around all braces when the stmt spans multiple lines. Previously. the breaks around the braces only fired if the body didn't fit on the line. That behavior lead to hard to read if-stmts because if was unclear where conditions stopped and the body started.

### DIFF
--- a/Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
+++ b/Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
@@ -381,7 +381,9 @@ final class IfStmtTests: PrettyPrintTestCase {
             SomeVeryLongTypeNameThatDefinitelyBreaks,
           baz: Baz
         ) = foo(a, b, c, d)
-      { return nil }
+      {
+        return nil
+      }
 
       """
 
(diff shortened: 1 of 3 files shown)


Tests/SwiftFormatPrettyPrintTests/IfStmtTests.swift
```diff
@@ -515,4 +515,49 @@ final class IfStmtTests: PrettyPrintTestCase {
...
+      if let foo = getmyfoo(), let bar = getmybar(),
+        foo.baz && bar.baz && someOtherCondition
+      {
+        foo()
+      } else {
+        bar()
+      }
+      if let foo = getmyfoo(), let bar = getmybar(),
+        foo.baz && bar.baz && someOtherCondition
+      { foo() }
```
allevato, line 557: I think it would be nice to have these breaks fire too when the condition wraps, if possible—that way, the body is only condensed to one line if the *whole statement* fits on one line. I'm surprised that doesn't happen already, since I'd expect the reset break that did fire to do so consistently with the breaks inside the braces. What's going on there, and would it be a lot harder to make it do that?

dylansturg: I intentionally only added the consistent group when there is an else clause, since the problem of inconsistent breaks around the clauses don't exist when there's only 1 clause. It's a simple change to unconditionally add the consistent group though.

PR #193, opened 2020-04-21
Updates for recent syntax/toolchain changes (2020-04-19).

allevato:
- Remove differentiation nodes that no longer exist.
- Add support for multiple `catch` clauses (SE-0276).
- Remove a never-used configuration helper type that
  incorrectly used a `let`-with-initial-value, which
  recent toolchain snapshots now issue a warning for
  in `Decodable` types.

This PR moves the `master` branch up to require the
`swift-DEVELOPMENT-SNAPSHOT-2020-04-19-a` development
snapshot toolchain.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -14,9 +14,9 @@
         "package": "SwiftSyntax",
         "repositoryURL": "https://github.com/apple/swift-syntax",
         "state": {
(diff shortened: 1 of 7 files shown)


allevato:
cc @dylansturg

PR #199, opened 2020-06-09
Peak memory use optimizations.

dylansturg:
Reduce peak memory usage of swift-format by discarding collections once they are no longer needed.
- Remove `Token` arrays from `beforeMap` and `afterMap` while processing the associated token.
- Empty `preVisitedExprs` after visitation finishes for the root expression of the tree.

# `beforeMap` and `afterMap`

By the end of a file, these maps can accumulate a meaningful number of tokens such that they significantly impact the peak memory usage.

This admittedly replaces a O(1) lookup operation with a O(N) lookup-and-remove operation, by replacing `subscript` with `removeValue(forKey:)`. In my testing, I found that this had no impact on the runtime. My hypothesis is that removing from these dictionaries keeps them small enough that O(N) runtime isn't a problem.

# `preVisitedExprs`

This set previously contained `ExprSyntax` objects and was maintained for the entire file. Once the visitor has finished with all children of the node that initiated inserting the contextual breaks, it will never visit any of the children again. This set can be emptied at that point to keep it from growing until the visitor finishes with the whole file.

# Results

The runtime is unchanged, but this reduces peak memory usage especially for large files.

- Large file (30k LOC): 108.3 MB peak to 72.1 MB peak
- Medium file (10k LOC) : 35.9 MB peak to 24.4 MB peak
- swift-protobuf library (many small files): 77.6 MB peak to 75.6 MB peak

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -43,7 +43,11 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
   /// Lists the expressions that have been visited, from the outermost expression, where contextual
   /// breaks and start/end contextual breaking tokens have been inserted.
-  private var preVisitedExprs = Set<ExprSyntax>()
+  private var preVisitedExprs = Set<SyntaxIdentifier>()
+
+  /// Tracks the "root" exprs where previsiting for contextual breaks started so that
+  /// `preVisitedExprs` can be emptied after exiting an expr tree.
+  private var rootExprs = Set<SyntaxIdentifier>()
 
   /// Lists the tokens that are the closing or final delimiter of a node that shouldn't be split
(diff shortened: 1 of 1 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -850,6 +854,13 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
     return .visitChildren
   }
 
+  override func visitPost(_ node: MemberAccessExprSyntax) {
+    let nodeID = node.id
+    if rootExprs.remove(nodeID) != nil {
+      preVisitedExprs.removeAll()
+    }
```
allevato, line 861: Can you pull this into a private helper function that takes a generic argument `<Node: ExprSyntaxProtocol>`? Then each of these `visitPost`s will be a one-liner.

dylansturg: Done

PR #202, opened 2020-06-12
Correct version string for `--version` flag (master)

MPLew-is:
This is going to be a pain to keep bumping for every release, but this change at least gets the correct version in for now.

Feel free to close this or #201 if you don't feel both are necessary, but since it seems like you'd cherry pick this commit anyway between the branches, I figured I'd save you some time and make them separate PRs.


MPLew-is:
I wholeheartedly agree it would be useful to have some way of accessing some versioning information from Swift Package Manager, but I'm not sure what easy mechanism there would be for embedding that information into the end binary. Would love to see a discussion on the subject though.

PR #203, opened 2020-06-12
Correct version string for `--version` flag (swift-5.1-branch)

MPLew-is:
This is going to be a pain to keep bumping for every release, but this change at least gets the correct version in for now.

I'm guessing at a version here based on the 5.2-related versioning scheme, but I feel like it's important to have something beyond `0.0.1` to indicate what version it is for testing purposes. Feel free to close though if you think this isn't necessary.

### DIFF
--- a/Sources/swift-format/main.swift
+++ b/Sources/swift-format/main.swift
@@ -56,7 +56,7 @@ fileprivate func main(_ arguments: [String]) -> Int32 {
     dumpDefaultConfiguration()
     return 0
   case .version:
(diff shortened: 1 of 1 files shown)


MPLew-is:
Completely understandable, I mostly wanted to suggest it to see if there might be some value in it (`0.0.1` is a little ambiguous, and [I got bit by a versioning issue](https://bugs.swift.org/browse/SR-12948) recently that this would have helped prevent). But no worries, I'll close this out.

PR #204, opened 2020-06-13
Allow indenting case statements relative to the containing switch block

MPLew-is:
Adds an `indentSwitchCaseLabels` configuration option to match Xcode's verbiage
Adds various tests to ensure this option does not affect other switch/case indentation
Fixes: https://bugs.swift.org/browse/SR-12947

I can't get the test suite to run to completion on `master` (`Fatal error: 'try!' expression unexpectedly raised an error: SwiftSyntax parser library isn't compatible: file /Users/mplewis/Projects/swift-format/Tests/SwiftFormatCoreTests/RuleMaskTests.swift, line 14`, likely due to me using the production swift release), so that's why this is into `swift-5.2-branch`. If you'd prefer it into `master`, I can go get a development snapshot running to test out my changes and re-target this PR.


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -627,10 +627,28 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+
+    // If switch/case labels were configured to be indented, insert an extra close break after the
+    // case body.
+    // The `after` call after this `if` block will insert its tokens before this one, matching up
+    // with the `open` break inserted above, which is before all other tokens.
+    if config.indentSwitchCaseLabels {
+       after(node.lastToken, tokens: .break(.close, size: 0))
+    }
     after(node.lastToken, tokens: .break(.close, size: 0), .close)
+
```
allevato, line 651: Instead of dealing with the order-inversion property of multiple calls to `after()` for the same node, you can collect the tokens in a local array and then pass them to a single call to `after` so that there's no confusion about the order. We do something like this elsewhere:

```swift
var afterLastTokenTokens: [Token] = [.break(.close, size: 0), .close]
if config.indentSwitchCaseLabels {
  afterLastTokenTokens.append(.break(.close, size: 0))
}
after(node.lastToken, tokens: afterLastTokenTokens)
```

MPLew-is: Done in 6a91b91

MPLew-is:
I have re-targeted and -tested this on `master` and everything seems to check out; please let me know if you have any other feedback!

And one thing to note is that it is fairly cumbersome to have to deal with development snapshots (especially if you're developing outside Xcode, like I do most of the time), so I might suggest having `master` (or whatever branch people should base development off of and submit pull requests to) be based on the current production version and have a development branch for the upcoming one. Just my two cents though, I know I'm not the one that has to deal with development on this on a daily basis.

PR #205, opened 2020-06-16
Remove @testable imports, upgrading symbols to public visibility.

dylansturg:


Sources/SwiftFormat/SyntaxValidatingVisitor.swift
```diff
@@ -68,7 +68,7 @@ fileprivate class SyntaxValidatingVisitor: SyntaxVisitor {
 /// syntax is valid.
 ///
 /// - Parameter syntax: The root of a tree of syntax nodes to check for compatibility.
-func firstInvalidSyntaxPosition(in syntax: Syntax) -> AbsolutePosition? {
+public func firstInvalidSyntaxPosition(in syntax: Syntax) -> AbsolutePosition? {
```
allevato, line 71: Since we have to sacrifice access control now, please add a leading underscore to this function name so that it doesn't leak (as obviously) into the `SwiftFormat` public API. (IIRC, SourceKit may have some logic that excludes or deprioritizes declarations with leading underscores from code complete suggestions.)

akyrtzi: >  SourceKit may have some logic that excludes or deprioritizes declarations with leading underscores from code complete suggestions

I think this is only for system frameworks but @rintaro will know for sure.

dylansturg: I think Argyrios is right - code complete is still suggesting `_firstInvalidSyntaxPosition` as the first option when I type something similar with other alternatives. I tried inside-of this module and in other modules (specifically the relevant test case).

I think a leading underscore still provides a hint to avoid using this function.

Alternatively, I could use an enum to add a namespace around this function (e.g. `SwiftFormatInternal`). Let me know if you think we should try harder to hide it.

rintaro: Yeah, hiding `_` prefixed symbols are for system frameworks only.

allevato: Ah, bummer. Well, this is fine for now—the leading underscore should be enough of an indicator that it should be hands-off. We can consider cleaning it up later if we feel there's a larger need.

Thanks for clarifying, @akyrtzi and @rintaro !

PR #206, opened 2020-06-20
Cherry-pick switch/case indentation PR

MPLew-is:
Cherry-picks and squashes #204 into the Swift 5.2 branch so it can be used against projects running the production Swift version.

Feel free to close if this you want to do this differently, just thought I'd save you some trouble from having to do it manually.


allevato:
This apparently slipped under my radar. My apologies!

While not the original request, this change did go into the 5.3-compatible release.

PR #207, opened 2020-06-24
Support building swift-format as part of Swift

dylansturg:
There are 3 changes to support building swift-format as part of the Swift project:

- Migrate other Swift project dependencies (swift-syntax and swift-argument-parser) to depend upon the master branches of those projects, so that they can be built as part of a single checkout.
- Create a build script that can be invoked by the Swift project's build system to build and test swift-format.
- Conditionally support building with local checkouts of Swift project dependencies (swift-syntax and swift-argument-parser).

When developing just for swift-format, it's still possible to use a recent DEV snapshot of the Swift toolchain to build and test without using local checkouts of the dependencies. That is the default behavior used by the Swift PM package (see the dependencies added with and without the `SWIFTCI_USE_LOCAL_DEPS` env var. 

I ran into some trouble when using `multiroot-data-file` for Swift PM when combining swift-format with the other swift-syntax dependent projects, because the other projects rely on test discovery. Test discovery seems to work fine for swift-format, so I've enabled it when using the build script. The `LinuxMain` testing support file is preserved because other existing workflows rely on that file to test on Linux when test discovery isn't supported.

This follows the instructions for incorporating swift-format into the Swift project's CI system:
https://forums.swift.org/t/testing-swift-format-as-part-of-swift-s-ci-infrastructure/30619

# Open Question

I'm not sure what is the best way to document the fact that `master` branch now depends on the latest build of Swift as opposed to a specific snapshot. I updated the README to note that the latest snapshot is necessary, but let me know if you think of a better way to document that.


dylansturg:
> I wonder if that table cell should just say "Swift at master".

That sounds like a good idea to me. I updated it.

akyrtzi:
@benlangmuir could you take a look at the build-script changes?
Also /cc @ahoppen for fyi

Package.resolved
```diff
@@ -5,17 +5,17 @@
         "package": "swift-argument-parser",
         "repositoryURL": "https://github.com/apple/swift-argument-parser.git",
         "state": {
-          "branch": null,
-          "revision": "8d31a0905c346a45c87773ad50862b5b3df8dff6",
-          "version": "0.0.4"
+          "branch": "master",
+          "revision": "eb51f949cdd0c9d88abba9ce79d37eb7ea1231d0",
+          "version": null
```
benlangmuir, line 10: Now that your dependencies are on branches rather than tags/versions, does it still make sense to have Package.resolved file checked in? Our CI system will build this against ToT for each dependency, but if a user checks it out or uses it as a dependency they will get whatever version is stamped here.  That seems like it will cause mismatches.

build-script-helper.py
```diff
@@ -0,0 +1,180 @@
...
+
+def printerr(message):
+  print(message, file=sys.stderr)
+
+def main(argv_prefix = []):
+  args = parse_args(argv_prefix + sys.argv[1:])
+  run(args)
+
+def parse_args(args):
+  parser = argparse.ArgumentParser(prog='BUILD-SCRIPT-HELPER.PY')
```
benlangmuir, line 30: What's up with all-caps here?

build-script-helper.py
```diff
@@ -0,0 +1,180 @@
...
+
+  parser.add_argument('--package-dir', default='')
```
benlangmuir, line 32: Please use `--package-path` to match `swift build` as well as other build-script-helper scripts such as sourcekit-lsp.

build-script-helper.py
```diff
@@ -0,0 +1,180 @@
...
+  parser.add_argument('-v', '--verbose', action='store_true', help='log executed commands')
+  parser.add_argument('--prefix', help='install path')
+  parser.add_argument('--config', default='debug')
+  parser.add_argument('--build-dir', default='.build')
```
benlangmuir, line 36: `--build-path`

build-script-helper.py
```diff
@@ -0,0 +1,180 @@
...
+  args = parse_args(argv_prefix + sys.argv[1:])
+  run(args)
+
+def parse_args(args):
+  parser = argparse.ArgumentParser(prog='BUILD-SCRIPT-HELPER.PY')
+
+  parser.add_argument('--package-dir', default='')
+  parser.add_argument('-v', '--verbose', action='store_true', help='log executed commands')
+  parser.add_argument('--prefix', help='install path')
+  parser.add_argument('--config', default='debug')
```
benlangmuir, line 35: `--configuration`

build-script-helper.py
```diff
@@ -0,0 +1,180 @@
...
+
+  parsed.swift_exec = os.path.join(parsed.toolchain, 'bin', 'swift')
+
+  # Convert package_dir to absolute path, relative to root of repo.
+  repo_path = os.path.dirname(__file__)
+  parsed.package_dir = os.path.realpath(
+                        os.path.join(repo_path, parsed.package_dir))
+
+  # Convert build_dir to absolute path, relative to package_dir.
+  parsed.build_dir = os.path.join(parsed.package_dir, parsed.build_dir)
```
benlangmuir, line 53: This is unintuitive to me.  In CI, this should always be absolute, and it should never be inside your source directory.  If I run this script manually, I would expect it to be relative to my working directory not the source tree.  If you want to make the default value be $SRC/.build to match swiftpm, you should only apply that to the default, not to all paths.  I.e.

* change default to `None`
* If path is `None`, explicitly set `build_dir = os.path.join(parsed.package_dir, '.build')`

Package.resolved
dylansturg: @allevato Is it safe to delete this?

allevato: Yeah, it should be deleted *and* added to `.gitignore`. Swift Package Manager will regenerate it when it does the checkout, but if it exists already it will be used, so we don't want wrong revisions to leak to users.

build-script-helper.py
dylansturg: I've renamed this param and the others that you flagged.

Just FYI - these names were taken from an [existing build-script-helper.py](https://github.com/apple/swift-stress-tester/blob/master/build-script-helper.py), so you or someone else may want to rename the params on that script for consistency.

akyrtzi: > Just FYI - these names were taken from an existing build-script-helper.py, so you or someone else may want to rename the params on that script for consistency.

/cc @nathawes

Package.resolved
allevato: Please make sure to add Package.resolved to `.gitignore` as well.

dylansturg: Overlooked that, thanks.

PR #208, opened 2020-06-29
Rewrite and fix `OneVariableDeclarationPerLine`.

allevato:
The old rule implementation assumed that if there were
multiple bindings in a single var decl, the last one
contained the type that should be applied to all of
them. The rules are actually more complex than that,
because they don't apply to bindings that have
initializer clauses (from which the type is inferred),
and they also allow multiple sequences of bindings in
the same decl (e.g., `var a, b: Int, c, d: String`).

This change also ensures proper handling of leading
comments (does not duplicate them) and that nested
declarations are transformed correctly.

Fixes SR-13051.


allevato:
cc @dylansturg

PR #209, opened 2020-06-30
Add portable Darwin build

dduan:
I use this script for distributing [DrString][], which also depends on `lib_InternalSwiftSyntaxParser.dylib`.

Summary of approach: 

1. distribute a copy of `lib_InternalSwiftSyntaxParser.dylib`
2. modify the main executable's rpath to include the copy's relative path. 

This ensures the .dylib is always compatible with the executable. The resulting .tar.gz can be unpacked into, say, the Homebrew installation folders. And users no longer need the "right" Xcode release to run `swift-format`.

[DrString]: https://github.com/dduan/DrString/blob/master/Scripts/package-darwin.sh

### DIFF
--- a/Scripts/package-darwin.sh
+++ b/Scripts/package-darwin.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+BUILD_PATH=.build/release
+LIB_PATH=$BUILD_PATH/lib
+BIN_PATH=$BUILD_PATH/bin/swift-format
+SWIFT_PATH=$(xcrun --find swift)
+
+swift build --configuration release --disable-sandbox
+rm -rf $LIB_PATH
+rm -rf $BIN_PATH
+mkdir -p $LIB_PATH
+mkdir -p $BUILD_PATH/bin
+cp .build/release/swift-format $BIN_PATH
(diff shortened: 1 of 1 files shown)


shepting:
What do you think about adding a CI test for this script to show how it works, and to verify that something doesn't break it?

dduan:
I'm not sure which CI system is preferred. If we could use GitHub Actions it'd be pretty trivial to implement.

allevato:
We're working on getting swift-format set up to run on the same swift-ci as apple/swift and apple/swift-syntax, since they're closely intertwined (changes to the compiler can result in changes to swift-syntax which can result in changes to swift-format).

Once we have that set up, that should give us some other benefits like automated tagging. I'm not sure what we can get there in the way of binary releases, but until we have that figured out I don't want to commit to supporting a particular script or distribution method just yet.

dduan:
That's understandable. This script is more appropriate for packagers (realistically, Homebrew package maintainers). I could see benefits of source package providing convenience for packagers but that's entirely optional.

From a very selfish POV this PR has been sitting here for too long for my PR-zero chasing desires. So I'd like to close it for now.

PR #212, opened 2020-07-29
Add support for Throws doc comments

frankus:
Adding a `- Throws:` declaration to a doc comment on a method with multiple parameters is causing a linting error. 

This PR adds support to parse throws documentation and will ensure that it is present on throwing functions/initializers and absent on non-throwing or rethrowing ones. 

I'm a little unsure of the comment declaration parsing code so some extra scrutiny is especially welcome there.

### DIFF
--- a/Sources/SwiftFormatRules/ValidateDocumentationComments.swift
+++ b/Sources/SwiftFormatRules/ValidateDocumentationComments.swift
@@ -24,12 +24,12 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
 
   public override func visit(_ node: InitializerDeclSyntax) -> SyntaxVisitorContinueKind {
     return checkFunctionLikeDocumentation(
-      DeclSyntax(node), name: "init", parameters: node.parameters.parameterList)
+      DeclSyntax(node), name: "init", parameters: node.parameters.parameterList, throwsOrRethrowsKeyword: node.throwsOrRethrowsKeyword)
   }
 
   public override func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
     return checkFunctionLikeDocumentation(
-      DeclSyntax(node), name: node.identifier.text, parameters: node.signature.input.parameterList,
+      DeclSyntax(node), name: node.identifier.text, parameters: node.signature.input.parameterList, throwsOrRethrowsKeyword: node.signature.throwsOrRethrowsKeyword,
       returnClause: node.signature.output)
   }
 
@@ -37,6 +37,7 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
     _ node: DeclSyntax,
     name: String,
     parameters: FunctionParameterListSyntax,
+    throwsOrRethrowsKeyword: TokenSyntax?,
     returnClause: ReturnClauseSyntax? = nil
   ) -> SyntaxVisitorContinueKind {
     guard let declComment = node.docComment else { return .skipChildren }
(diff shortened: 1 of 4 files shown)


Sources/SwiftFormatRules/ValidateDocumentationComments.swift
```diff
@@ -95,6 +97,22 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
...
   }
+
+  /// Ensures the function has throws documentation if it may actually throw
+  /// an error.
+  private func validateThrows(
+    _ throwsOrRethrowsKeyword: TokenSyntax?,
+    name: String,
+    throwsDesc: String?
+  ) {
+    let needsThrowsDesc = throwsOrRethrowsKeyword != nil && throwsOrRethrowsKeyword!.tokenKind == .throwsKeyword
```
allevato, line 108: This could be written more simply as `let needsThrowsDesc = throwsOrRethrowsKeyword?.tokenKind == .throwsKeyword`; if `throwsOrRethrowsKeyword` is nil, then that equality will also fail so there's no need to check it separately.

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
```diff
@@ -153,4 +171,15 @@ extension Diagnostic.Message {
     "replace the singular inline form of 'Parameter' tag with a plural 'Parameters' tag "
       + "and group each parameter as a nested list"
   )
+
+  public static func removeThrowsComment(funcName: String) -> Diagnostic.Message {
+    return Diagnostic.Message(
+      .warning,
+      "remove the throws comment of \(funcName), it doesn't throw an error"
```
allevato, line 178: Suggested message: `"remove the 'Throws' tag for non-throwing function \(funcName)"`

(I realize that yours is similar to the `return` message, but it should be improved at some point as well; we can do that in a separate change.)

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
```diff
@@ -153,4 +171,15 @@ extension Diagnostic.Message {
...
+    )
+  }
+
+  public static func documentErrorsThrown(funcName: String) -> Diagnostic.Message {
+    return Diagnostic.Message(.warning, "document the errors thrown by \(funcName)")
```
allevato, line 183: Suggested message: `"add a 'Throws' tag describing the errors thrown by \(funcName)"`

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
```diff
@@ -95,6 +97,22 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
       diagnose(.documentReturnValue(funcName: name), on: returnClause)
     }
   }
+
+  /// Ensures the function has throws documentation if it may actually throw
+  /// an error.
+  private func validateThrows(
```
allevato, line 103: It's probably worth an additional brief paragraph of documentation here to say that it's an explicit choice to **not** consider functions that use `rethrows` as needing `Throws` tags.

I wasn't sure if there was an established practice on this, but at least in the standard library, none of the functions that use `rethrows` document their errors; the only functions that have `Throws` tags are in `Result` and the various `Codable`-related types, which are all `throws`. So following that precedent (as you're already doing) makes sense.

frankus: I was going off of the default behavior in Xcode, which in the release version is to crash 😜 and in the beta is to *not*  stub out a `- Throws` comment for functions that rethrow. 

But yeah, good call on calling that out explicitly.

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
frankus: Fixed.

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
frankus: Fixed.

Sources/SwiftFormatRules/ValidateDocumentationComments.swift
frankus: Fixed.

frankus:
I'm open to reverting the last commit, but it seemed like a single enum handles things better than three separate `Bool`s. At the same time, this introduces even more changes to part of the code you mentioned that you're working on rewriting.

PR #215, opened 2020-07-31
Visit child nodes in NoParensAroundConditions rule.

dylansturg:
Without visiting the child nodes, the rule wasn't applied to nested nodes meaning parens wouldn't be removed.

### DIFF
--- a/Sources/SwiftFormatRules/NoParensAroundConditions.swift
+++ b/Sources/SwiftFormatRules/NoParensAroundConditions.swift
@@ -40,18 +40,28 @@ public final class NoParensAroundConditions: SyntaxFormatRule {
       $0.highlight(expr.sourceRange(converter: self.context.sourceLocationConverter))
     }
 
+    guard
+      let visitedTuple = visit(tuple).as(TupleExprSyntax.self),
+      let visitedExpr = visitedTuple.elementList.first?.expression
+    else {
+      return expr
+    }
     return replaceTrivia(
-      on: expr,
-      token: expr.lastToken,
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatRules/NoParensAroundConditions.swift
```diff
@@ -50,8 +50,12 @@ public final class NoParensAroundConditions: SyntaxFormatRule {
...
     let conditions = visit(node.conditions).as(ConditionElementListSyntax.self)!
-    let result = node.withIfKeyword(node.ifKeyword.withOneTrailingSpace())
+    var result = node.withIfKeyword(node.ifKeyword.withOneTrailingSpace())
       .withConditions(conditions)
+      .withBody(CodeBlockSyntax(visit(node.body)))
+    if let elseBody = node.elseBody {
+      result = result.withElseBody(visit(elseBody))
+    }
     return StmtSyntax(result)
   }
```
allevato, line 61: This is the closest GitHub will let me leave a comment to something that was collapsed...

It looks like we don't recurse into `tup` in `visit(_: ConditionElementSyntax)` starting on line 62, so we might still be missing some (admittedly disgusting) cases like the inner parentheses in this one:

```swift
if foo.someCall({ if (x) {} }) {}
```

Can you test that too?

dylansturg: That's a good point. This also made me realize that we need to visit the child nodes in the early exit path for many other node types here. I fixed those as well, and added relevant tests.

PR #217, opened 2020-08-11
Add build-only support for the `async` keyword.

allevato:
Since the parser's behavior is gated on the `-enable-experimental-concurrency`
compiler flag, we unfortunately can't write any tests for this yet as the
in-memory parser will not have the concurrency support enabled. This change
only adds minimal support for `async` in the pretty printer and fixes a
build breakage in `master` caused by the addition of the new keyword.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1401,6 +1401,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   override func visit(_ node: FunctionTypeSyntax) -> SyntaxVisitorContinueKind {
     after(node.leftParen, tokens: .break(.open, size: 0), .open)
     before(node.rightParen, tokens: .break(.close, size: 0), .close)
+    before(node.asyncKeyword, tokens: .break)
     before(node.throwsOrRethrowsKeyword, tokens: .break)
     before(node.arrow, tokens: .break)
     before(node.returnType.firstToken, tokens: .break)
@@ -1603,6 +1604,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 
   override func visit(_ node: FunctionSignatureSyntax) -> SyntaxVisitorContinueKind {
+    before(node.asyncKeyword, tokens: .break)
     before(node.throwsOrRethrowsKeyword, tokens: .break)
     before(node.output?.firstToken, tokens: .break)
     return .visitChildren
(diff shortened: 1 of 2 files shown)


allevato:
@akyrtzi @DougGregor This is an interesting situation. Since the parser's behavior around the `async` keyword is controlled by `-enable-experimental-concurrency`, it's not currently possible for swift-format (or any SwiftSyntax client) to do the right thing at runtime with code that uses `async` because SwiftSyntax won't set that flag and the source won't be parsed correctly. This is kind of unfortunate, since it means anyone using the experimental feature will have broken output if they try to use swift-format, even using a version built with a version of the parser that "supported" it.

Since it's probably still desired to have the compiler itself fail at parse time if a keyword is used without the flag being provided, can we ensure that the _parser library's_ behavior is never conditional on compiler flags? For example, we could define a constant like this:

```cpp
#if SWIFT_BUILD_ONLY_SYNTAXPARSERLIB
#define SWIFT_ALLOW_EXPERIMENTAL_PARSE true
#else
#define SWIFT_ALLOW_EXPERIMENTAL_PARSE false
#endif
```

And then, anywhere that we currently check `Context.LangOpts.EnableExperimentalConcurrency`, we could check `(SWIFT_ALLOW_EXPERIMENTAL_PARSE || Context.LangOpts.EnableExperimentalConcurrency)` instead.

WDYT?

DougGregor:
I think that's a reasonable solution, yes!

allevato:
Thanks! I'll take a stab at a PR to add that.

DougGregor:
Hmm. This will get us into some trouble if we actually do introduce any source breaks. I've been using `__await <expression>` rather than `await <expression>` to avoid breaking source while we experiment, but anything source-breaking we do at the syntactic level will potentially cause breakage here.

akyrtzi:
`SWIFT_BUILD_ONLY_SYNTAXPARSERLIB` is not used from the public CI currently, a better way is to allow this if `LangOpts.ParseForSyntaxTreeOnly` is true.

allevato:
> This will get us into some trouble if we actually do introduce any source breaks.

For source-breaking changes in general, would they always need to be guarded by a `-swift-version` bump? I don't think that's a situation we've had to deal with yet, so could that mean SwiftSyntax might need to take the language version as an optional argument to ensure the right code paths are used?

DougGregor:
> For source-breaking changes in general, would they always need to be guarded by a `-swift-version` bump? I don't think that's a situation we've had to deal with yet, so could that mean SwiftSyntax might need to take the language version as an optional argument to ensure the right code paths are used?

Source-breaking changes would need a `-swift-version` bump, yes. And we haven't really dealt with one of this since SwiftSyntax has existed, because we've only done additive things. It would be better if we didn't have to pass a `-swift-version` through at all, so syntax-based tools don't have to care which language dialect we're working with. I don't know if that's an achievable goal; it might work in practice for the `await` case I'm talking about because the ambiguity is more semantic than syntactic.

PR #219, opened 2020-08-12
Support opt-in rules using an `isOptIn` class property.

dylansturg:
The codegen pipeline evaluates the new property for rule types that are discovered and uses it to write the correct default value into `RuleRegistry`. No rules are converted to opt-in status as part of this change.

The new property needs to be support overriding in subclasses, so I had to define it as an open class property in the 2 superclasses for rules (`SyntaxFormatRule` and `SyntaxLintRule`).

### DIFF
--- a/Sources/generate-pipeline/RuleCollector.swift
+++ b/Sources/generate-pipeline/RuleCollector.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import SwiftFormatCore
 import SwiftSyntax
 
 // These rules will not be added to the pipeline.
@@ -19,7 +20,7 @@ let suppressRules = ["UseEarlyExits", "UseWhereClausesInForLoops"]
 /// Collects information about rules in the formatter code base.
 final class RuleCollector {
   /// Information about a detected rule.
-  private struct DetectedRule {
+  struct DetectedRule: Hashable {
     /// The type name of the rule.
     let typeName: String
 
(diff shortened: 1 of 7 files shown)


Sources/generate-pipeline/RuleCollector.swift
```diff
@@ -132,7 +136,10 @@ final class RuleCollector {
       /// Ignore it if it doesn't have any; there's no point in putting no-op rules in the pipeline.
       /// Otherwise, return it (we don't need to look at the rest of the inheritances).
       guard !visitedNodes.isEmpty else { return nil }
-      return DetectedRule(typeName: typeName, visitedNodes: visitedNodes, canFormat: canFormat)
+      let isOptIn =
+        (_typeByName("SwiftFormatRules.\(typeName)") as? Rule.Type).map { $0.isOptIn } ?? false
```
allevato, line 140: This doesn't seem like it should work since `generate-pipeline` doesn't depend the on `SwiftFormatRules` module. Are you sure `_typeByName` isn't just returning `nil` here and it's ending up `false` by default, yielding the same effective outcome?

If we add the dependency to `SwiftFormatRules`, I think we should make it a stronger precondition failure if `_typeByName` returns nil because that shouldn't happen for any rules that we discover through our syntax walk, and then we should be safer.

dylansturg: Good point. Fixed that.

PR #221, opened 2020-08-17
Add an exception to AlwaysUseLowerCamelCase for test names.

dylansturg:
When using descriptive names, e.g. explaining the context and expected outcome in the name, the test names can be quite long and using underscores in the names to organize phrases becomes beneficial. In other situations (i.e. not test cases), it's better to use shorter names that don't require such organization.

This change adds an exception specifically for test cases so that the linter won't complain about underscores in those names.

### DIFF
--- a/Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
+++ b/Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
@@ -19,35 +19,81 @@ import SwiftSyntax
 /// Lint: If an identifier contains underscores or begins with a capital letter, a lint error is
 ///       raised.
 public final class AlwaysUseLowerCamelCase: SyntaxLintRule {
+  /// Stores function decls that are test cases.
+  private var testCaseFuncs = Set<FunctionDeclSyntax>()
+
+  public override func visit(_ node: SourceFileSyntax) -> SyntaxVisitorContinueKind {
+    // Tracks whether "XCTest" is imported in the source file before processing individual nodes.
+    setImportsXCTest(context: context, sourceFile: node)
+    return .visitChildren
+  }
+
+  public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+    // Check if this class is an `XCTestCase`, otherwise it cannot contain any test cases.
+    guard context.importsXCTest == .importsXCTest else { return .visitChildren }
+
+    // Identify and store all of the function decls that are test cases.
+    let testCases = node.members.members.compactMap {
+      $0.decl.as(FunctionDeclSyntax.self)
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
```diff
@@ -19,31 +19,72 @@ import SwiftSyntax
...
+  private var testCaseFuncs = Set<FunctionDeclSyntax>()
+
+  public override func visit(_ node: SourceFileSyntax) -> SyntaxVisitorContinueKind {
+    // Tracks whether "XCTest" is imported in the source file before processing individual nodes.
+    setImportsXCTest(context: context, sourceFile: node)
+    return .visitChildren
+  }
+
+  public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+    // Check if this class is an `XCTestCase`, otherwise it cannot contain any test cases.
```
allevato, line 32: Since we don't have type information, this would fail to capture anyone who uses an intermediate subclass of `XCTestCase` as the base class for their tests.

Since we're checking whether `XCTest` is imported, should we just assume that that + `func test*()` is a good enough heuristic and not worry about this one?

Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
```diff
@@ -19,31 +19,72 @@ import SwiftSyntax
...
+
+    // Identify and store all of the function decls that are test cases.
+    let testCases = node.members.members.compactMap {
+      $0.decl.as(FunctionDeclSyntax.self)
+    }.filter {
+      // Filter out non-test methods using the same heuristics as XCTest to identify tests.
+      // Test methods are methods that start with "test", have no arguments, and void return type.
+      $0.identifier.text.starts(with: "test")
+        && $0.signature.input.parameterList.isEmpty
+        && $0.signature.output == nil
```
allevato, line 51: In recent versions of XCTest, test methods are allowed to be declared `throws`, so we need to handle that possibility as well.

Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
dylansturg: That seems reasonable to me. I removed the superclass check.

Sources/SwiftFormatRules/AlwaysUseLowerCamelCase.swift
dylansturg: The `throws` keyword will be alright, and I added a test for it. But this did make me realize I wasn't handling `-> Void` or `-> ()` correctly.

PR #224, opened 2020-08-25
Apply AlwaysUseLowerCamelCase in more places.

dylansturg:
The AlwaysUseLowerCamelCase rule now applies to the following additional nodes:
- Closure paramters, single param and parameter lists
- Function parameters
- Optional binding conditions (e.g. bindings in if and guard stmts)
- Recursively in nodes whose parent nodes were previously skipping children


allevato:
@swift-ci please test

PR #226, opened 2020-08-28
Allocate 1 instance of each rule per LintPipeline.

dylansturg:
The previous implementation allocated a new instance of each relevant rule at each node in the `LintPipeline`. This meant that rules couldn't share state between nodes and it meant that potentially many instances of each rule were allocated per linted file.

This change has a modest impact improving performance (15-20% in some testing).

### DIFF
--- a/Sources/SwiftFormat/LintPipeline.swift
+++ b/Sources/SwiftFormat/LintPipeline.swift
@@ -27,10 +27,10 @@ extension LintPipeline {
   ///   - node: The syntax node on which the rule will be applied. This lets us check whether the
   ///     rule is enabled for the particular source range where the node occurs.
   func visitIfEnabled<Rule: SyntaxLintRule, Node: SyntaxProtocol>(
-    _ visitor: (Rule) -> (Node) -> SyntaxVisitorContinueKind, in context: Context, for node: Node
+    _ visitor: (Rule) -> (Node) -> SyntaxVisitorContinueKind, for node: Node
   ) {
     guard context.isRuleEnabled(Rule.self.ruleName, node: Syntax(node)) else { return }
-    let rule = Rule(context: context)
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormat/LintPipeline.swift
```diff
@@ -20,38 +20,35 @@ extension LintPipeline {
...
-  ///   - context: The formatter context that contains information about which rules are enabled or
-  ///     disabled.
+  ///   - visitor: A reference to the `visit` method on an instance of a `SyntaxLintRule` subclass.
   ///   - node: The syntax node on which the rule will be applied. This lets us check whether the
   ///     rule is enabled for the particular source range where the node occurs.
-  func visitIfEnabled<Rule: SyntaxLintRule, Node: SyntaxProtocol>(
-    _ visitor: (Rule) -> (Node) -> SyntaxVisitorContinueKind, in context: Context, for node: Node
+  ///   - name: The name of the rule from the `SyntaxLintRule` subclass. Used to determine if the
+  ///     rule should be applied to the current node.
+  func visitIfEnabled<Node: SyntaxProtocol>(
```
allevato, line 28: Is it possible to keep the signature of these methods the same to avoid the repetition and churn to the generator?

Since we have `Rule` as a generic argument that's inferred from the `visit` method that gets passed in, it seems like we should be able to replace the `Rule(context:)` initialization in this function with a call to the `rule` helper function to get the cached one, but leave everything else more or less the same.

dylansturg: Good point. I mostly reverted the signatures of these methods, but I still deleted the `context` param because it's not used and it's already available as a property of `LintPipeline` anyway.

allevato: Even better!

PR #228, opened 2020-09-01
Handle existing commas in single element collections.

dylansturg:
The `whitespaceOnly` mode in the PrettyPrinter handle existing commas in single element collections incorrectly by always removing the comma. This happened because an earlier fix stopped placing `commaDelimitedRegion[Start|End]` tokens around single element collections, while still ignoring any existing trailing comma token. Rather than move the `whitespaceOnly` decision deep into the TokenStreamCreator, I've revised that previous fix so that all logic to decide when to keep and remove commas is localized inside the PrettyPrinter.

This was causing crashes when linting code that had a trailing comma in a single element collection, because `whitespaceOnly` mode was incorrectly causing non-whitespace changes to occur.

Fixes SR-12782

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/Token.swift
+++ b/Sources/SwiftFormatPrettyPrint/Token.swift
@@ -185,8 +185,8 @@ enum Token {
   case commaDelimitedRegionStart
 
   /// Marks the end of a comma delimited collection, where a trailing comma should be inserted
-  /// if and only if the collection spans multiple lines.
-  case commaDelimitedRegionEnd(hasTrailingComma: Bool)
+  /// if and only if the collection spans multiple lines and has multiple elements.
+  case commaDelimitedRegionEnd(hasTrailingComma: Bool, isSingleElement: Bool)
 
   /// Starts a scope where `contextual` breaks have consistent behavior.
   case contextualBreakingStart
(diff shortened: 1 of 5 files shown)


Sources/SwiftFormatPrettyPrint/Token.swift
```diff
@@ -185,8 +185,8 @@ enum Token {
   case commaDelimitedRegionStart
 
   /// Marks the end of a comma delimited collection, where a trailing comma should be inserted
-  /// if and only if the collection spans multiple lines.
-  case commaDelimitedRegionEnd(hasTrailingComma: Bool)
+  /// if and only if the collection spans multiple lines, and the collection had multiple elements.
```
jrtibbetts, line 188: `if and only if the collection spans multiple lines and has multiple elements.`

dylansturg:
> Nice! It looks like this may fix [SR-12782](https://bugs.swift.org/browse/SR-12782)?

I didn't realize we already had a bug for this. I tagged it in the PR's description.

Sources/SwiftFormatPrettyPrint/Token.swift
dylansturg: Thanks. Updated the text.

PR #230, opened 2020-09-10
Format multiple trailing closures.

allevato:
When a function call has multiple trailing closures, every closure is forced to wrap.

---

There are two potentially surprising things about the behavior here:

1. Unlike an `if/else` statement, which can be formatted like this if it would fit entirely on one line:

    ```swift
    if foo { bar } else { baz }
    ```

   ...we do not allow the same thing to occur for multiple trailing closures at this time:

    ```swift
    // this
    foo(x) { bar() } label: { baz() }
    // will always become
    foo(x) {
      bar()
    } label: {
      baz()
    }
    ```

2. The forced break occurs after the closure signature `{ arg, arg, arg in‸` if it's present; otherwise, it occurs after the open brace. This lets the closure signature stay on the same line as the open brace if it fits, but it will still wrap (via the normal break that occurs here `{‸arg, arg, arg in`) if it wouldn't fit.

   However, because the break after `in‸` is _forced_, that means that if the break after `{‸` _also_ fires, then even a single-statement closure will have a line break after the signature, when regular closures otherwise wouldn't. In other words,

    ```swift
    // possible
    reallyLongFunctionName(x) {
      x in bar(x)
    }

    // NOT possible
    reallyLongFunctionName(x) {
      x in bar(x)
    } secondLabelThatIsAlsoLong: {
      y in baz(y)
    }

    // ...it becomes this:
    reallyLongFunctionName(x) {
      x in
      bar(x)
    } secondLabelThatIsAlsoLong: {
      y in
      baz(y)
    }
    ```

I can live with (1); I don't think it's a *huge* problem because I imagine the number of function calls with multiple trailing closures that would fit entirely on one line are exceedingly rare to begin with, and I think in the case of multiple trailing closures it's more readable to have them wrapped anyway so that the label isn't hidden in the middle of the line.

I'm less happy about (2), since it's a weird inconsistency, but I'm not sure our current model is flexible enough to say "given `{₁arg, arg, arg in₂`, force break 2 only if we didn't already break at 1". It's almost like a reset, but not quite the same because we're not conditioning it on continuation line state.

@dylansturg Any ideas here?

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -54,6 +54,10 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   /// moved past these tokens.
   private var closingDelimiterTokens = Set<TokenSyntax>()
 
+  /// Tracks closures that are never allowed to be laid out entirely on one line (e.g., closures
+  /// in a function call containing multiple trailing closures).
+  private var forcedBreakingClosures = Set<SyntaxIdentifier>()
+
   init(configuration: Configuration, operatorContext: OperatorContext) {
     self.config = configuration
     self.operatorContext = operatorContext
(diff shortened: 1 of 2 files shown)


dylansturg (approved):

allevato:
Going to go ahead and merge this to get it into the 5.3 branch; I'll revisit improvements to the newline behavior above later.

PR #231, opened 2020-09-10
Use regular property initialization for argument defaults.

allevato:
This gets rid of warnings from deprecated swift-argument-parser APIs.


allevato:
cc @dylansturg

dylansturg (approved):

PR #243, opened 2020-10-19
Multithread processing source files

keith:
Previously every source file was formatted / linted one by one. On our
codebase this took a full project format from ~17 minutes to ~5 minutes.

This is an updated version of https://github.com/apple/swift-format/pull/117

Depends on:

- https://github.com/apple/swift-format/pull/242
- https://github.com/apple/swift-syntax/pull/243

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,9 @@ as well as the following command line options:
 * `-i/--in-place`: Overwrites the input files when formatting instead of
   printing the results to standard output.
 
+* `-p/--parallel`: Process files in parallel, simultaneously across
+  multiple cores.
+
 * `-r/--recursive`: If specified, then the tool will process `.swift` source
   files in any directories listed on the command line and their descendants.
   Without this flag, it is an error to list a directory on the command line.
--- a/Sources/swift-format/Frontend/Frontend.swift
+++ b/Sources/swift-format/Frontend/Frontend.swift
@@ -92,7 +92,7 @@ class Frontend {
     if paths.isEmpty {
       processStandardInput()
     } else {
-      processPaths(paths)
+      processPaths(paths, parallel: lintFormatOptions.parallel)
     }
   }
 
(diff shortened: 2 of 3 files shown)


Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -49,6 +49,9 @@ struct LintFormatOptions: ParsableArguments {
     """)
   var ignoreUnparsableFiles: Bool = false
 
+  @Flag(help: "Whether or not to run formatting or linting in parallel consuming all possible resources")
```
mattt, line 52:
```suggestion
  /// Whether or not to run the formatter/linter in parallel.
  @Flag(help: "Process files in parallel, simultaneously across multiple cores.")
```

mattt: Suggested edit to match existing documentation and help text convention.

As far as the note about "consuming all possible resources", I wasn't able to find any definitive, on-the-record language from Apple to support that. The most they've committed to publicly (that I can find, at least) was from [WWDC 2017 Session 706](https://asciiwwdc.com/2017/sessions/706?q=parallel).

mattt (changes requested):

mattt:
@swift-ci Please test

mattt:
@allevato I just noticed that this project isn't setup to require CI checks to pass before merging. Is this something you'd consider implementing for the swift-format repository?

allevato:
We've submitted changes to checkout `swift-format` as part of the CI builds and added build scripts for that purpose (see https://forums.swift.org/t/testing-swift-format-as-part-of-swift-s-ci-infrastructure/30619).

@shahmishal , can you update us on the current status of swift-format on Swift CI? Is there anything else we need to do on our end or does your team need to enable something for this to work?

Jeehut:
This looks amazing, thank you @keith. I think the new `parallel` option should be documented in the `README.md` though, would you mind adding that to be explained alongside `--in-place` and `--recursive`?

keith:
Done!

mattt (approved):

README.md
```diff
@@ -82,6 +82,9 @@ as well as the following command line options:
 * `-i/--in-place`: Overwrites the input files when formatting instead of
   printing the results to standard output.
 
+* `-p/--parallel`: Whether or not to run formatting or linting in parallel
+  consuming all possible resources printing the results to standard output.
```
allevato, line 86: Can you update this the same way you updated the flag text in the code to sound less dire? 😄 

Something like "[...] to run formatting or linting in parallel across multiple cores".

Also, we don't need to say anything about stdout, right? If this option is paired with `-i`, the files will still be updated in-place?

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -124,28 +124,41 @@ class Frontend {
...
     precondition(
       !paths.isEmpty,
       "processPaths(_:) should only be called when paths is non-empty.")
 
-    for path in FileIterator(paths: paths) {
-      guard let sourceFile = FileHandle(forReadingAtPath: path) else {
-        diagnosticEngine.diagnose(Diagnostic.Message(.error, "Unable to open \(path)"))
-        continue
+    if parallel {
+      let allFilePaths = Array(FileIterator(paths: paths))
```
allevato, line 133: _Rambling:_ I still have some concerns about the upfront flattening of the `FileIterator` being slow if the formatter is run in parallel on something like a networked file system or in recursive mode on the root of a large monorepo (who would do such a thing), but I should probably test my concerns against real data before going to the trouble of adding more work.

If we need to change this in the future, I was concerned about having to do our own thread pooling, but maybe there's an easy alternative here: instead of flattening the whole iterator at once, just pull off a both sufficiently large and sufficiently small number of files at a time (say, 200, or a nice round number like 256) and loop over the batches while processing each batch with `concurrentPerform`. That would probably still get us good parallelism in the batches while avoiding eagerly flattening the whole array.

You don't need to make this change now (I might be worried about the performance for nothing), but I just wanted to write it out here so I don't forget it later if we do need it.

README.md
keith: Yep sorry, bad copy paste

Sources/swift-format/Frontend/Frontend.swift
keith: +1 good thoughts, I would definitely be interested to hear if you see issues with this given the large number of files.

allevato:
Thanks for your work on this, and your patience while we sorted out various other related issues. Let's merge it finally 🎉

keith:
Thanks!

PR #244, opened 2020-10-26
Improve pattern recognition in `UseLetInEveryBoundCaseVariable`.

allevato:
This rule was previously flagging uses of `let/var` that don't
apply (such as `let x as Y`). This commit also includes
improvements to drill through optional patterns (e.g.,
`case let .foo(x)?`) and now handles tuple patterns as well
(e.g., `case let (x, y)`).

Fixes SR-11773.


allevato:
cc @dylansturg

dylansturg (approved):

PR #249, opened 2021-01-20
Apply private and fileprivate rules to UseSynthesizedInitializer.

dylansturg:
The linter rule didn't handle the cases when the stored properties of a struct are private or fileprivate. A synthesized memberwise initializer has access level according to the stored properties. If a type had a private stored property and a memberwise initializer without an access level modifier, then the rule was suggesting to remove the initializer even though the synthesized initializer would have private visibility.

The rules for `private` vs `fileprivate` vs `internal` for synthesized, or default, memberwise initializers is taken from:
[The Swift Programming Language - Access Control](https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#ID21).


allevato (approved):
Nice!

PR #251, opened 2021-03-13
Test Pull Request

shahmishal:


shahmishal:
@swift-ci test

shahmishal:
```
swift-syntax/Sources/SwiftSyntax/RawSyntax.swift:120:45: error: value of type 'swiftparse_token_data_t' has no member 'range'
```
@allevato is this known issue with main branch?

allevato:
@shahmishal Thanks for getting this wired up!

We are aware of some upstream changes that we haven't picked up yet in swift-format, but I would expect most of them to be around syntax node changes that we haven't adopted (like places where `actor` can appear now). I wouldn't expect to see errors when compiling the swift-syntax _dependency_... maybe @ahoppen or @akyrtzi has some insight? Could our build script be incorrect?

ahoppen:
In https://github.com/apple/swift/pull/36249 and https://github.com/apple/swift/pull/36283 I changed the C API through which SwiftSyntax communicates with the compiler. It looks like the CI bot is not using the latest toolchain to compiler `swift-format`. 

@shahmishal I don’t see any messages regarding the download or build of the latest toolchain in the CI logs. Could it be that we’re missing that step?

Also, I think we actually need to build a `main` compiler to test `swift-format` because `swift-format` depends on the `main` branch of SwiftSyntax which in turn depends on the `main` branch of `swift`. I also don’t think we can get away with just using the latest open source toolchain because whenever the Python syntax files [change](https://github.com/apple/swift/commits/main/utils/gyb_syntax_support), we need the corresponding PR to be merged in both `swift` and `swift-syntax`, otherwise SwiftSyntax will crash on launch. And these changes are fairly frequent (rough guess from the linked changelog 1 change/month). I know that’s a bit unfortunate for `swift-format` testing times 😬

shahmishal:
@swift-ci test

allevato:
@swift-ci please test

(just want to get some fresh logs)

ahoppen:
@swift-ci Please test

PR #255, opened 2021-04-30
Fix some tests that are broken against the latest toolchains.

allevato:
These are broken either because of parser crashes (due to malformed
input, which we can change easily enough) or due to potential
behavioral changes.

See SR-14552 and SR-14553.


allevato:
Manually tested with `swift test`:

```
Test Suite 'All tests' passed at 2021-04-30 12:45:47.279.
	 Executed 564 tests, with 0 failures (0 unexpected) in 11.137 (11.200) seconds
```

PR #263, opened 2021-09-01
Remove the hardcoded `suppressRules` list in `generate-pipeline`.

allevato:
Now that we have the `isOptIn` property for rules, this should be used to disable rules by default instead of removing them from pipeline generation entirely. This also unblocks https://github.com/apple/swift-format/pull/261, since excluded rules wouldn't show up in the generated name cache, causing their tests to crash.


allevato:
@dylansturg

dylansturg (approved):

PR #267, opened 2021-09-25
Fix crash when running Lint from LegacyMain

mtgto:
Fix crash bug in LegacyMain lint mode. I test with Xcode 13 and [tag 0.50500.0](https://github.com/mtgto/swift-format/releases/tag/0.50500.0) on macOS 11.6 Intel.
This crash does not happen in using SubCommand lint.

In addition, adds a description of `strict` flag to README.

```console
$ swift run swift-format --mode lint Package.swift
[0/0] Build complete!
/Users/user/work/cocoa/swift-format/Package.swift:15:1: warning: [OrderedImports]: sort import statements lexicographically
/Users/user/work/cocoa/swift-format/Package.swift:83:22: warning: remove trailing comma from the last element in single line collection literal
/Users/user/work/cocoa/swift-format/Package.swift:24:18: warning: [RemoveLine]: remove line break
/Users/user/work/cocoa/swift-format/Package.swift:140:2: warning: [RemoveLine]: remove line break
/Users/user/work/cocoa/swift-format/Package.swift:147:1: warning: [LineLength]: line is too long
ArgumentParser/Flag.swift:103: Fatal error:
--------------------------------------------------------------------
Can't read a value from a parsable argument definition.

This error indicates that a property declared with an `@Argument`,
`@Option`, `@Flag`, or `@OptionGroup` property wrapper was neither
initialized to a value nor decoded from command-line arguments.

To get a valid value, either call one of the static parsing methods
(`parse`, `parseAsRoot`, or `main`) or define an initializer that
initializes _every_ property of your parsable type.
--------------------------------------------------------------------

zsh: illegal hardware instruction  swift run swift-format --mode lint Package.swift
```
----

#252 introduces `--strict`, but this flag should be defined in LintFormatOptions.swift, not Lint.swift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ as well as the following command line options:
   files in any directories listed on the command line and their descendants.
   Without this flag, it is an error to list a directory on the command line.
 
+* `-s/--strict`: If specified, exits non-zero code if one or more warning are found.
+  This option only works in lint mode.
+
 ### Configuration
 
 For any source file being checked or formatted, `swift-format` looks for a
--- a/Sources/swift-format/Subcommands/Lint.swift
+++ b/Sources/swift-format/Subcommands/Lint.swift
@@ -21,18 +21,12 @@ extension SwiftFormatCommand {
 
     @OptionGroup()
     var lintOptions: LintFormatOptions
-    
-    @Flag(
-      name: .shortAndLong,
-      help: "Fail on warnings."
-    )
-    var strict: Bool = false
 
     func run() throws {
       let frontend = LintFrontend(lintFormatOptions: lintOptions)
       frontend.run()
       if frontend.errorsWereEmitted { throw ExitCode.failure }
-      if strict,
+      if lintOptions.strict,
          frontend.diagnosticEngine.diagnostics
             .contains(where: { $0.message.severity == .warning }) {
         throw ExitCode.failure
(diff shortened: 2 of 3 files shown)


kylebrowning:
I meant to say I tested your PR and it didn't crash

mtgto:
😄

allevato:
Since this only happens with the legacy `--mode` flag, we should just remove that entirely (I've been meaning to anyway, but it's never bubbled up in my priority list). There's no compelling reason to keep the legacy mode anymore—it was only provided for users to transition to the subcommand syntax, which they've had plenty of time to do.

allevato:
The legacy entry point and `--mode` flag were removed in https://github.com/apple/swift-format/pull/275, so AFAICT this is no longer needed. (Please re-open if that isn't the case!)

Thanks for bringing it up regardless—it was the motivation I needed to kick out the old compatibility mode. 🙂

mtgto:
@allevato 

> The legacy entry point and --mode flag were removed in #275, so AFAICT this is no longer needed. (Please re-open if that isn't the case!)

README.md has description about options `--mode`, and lack of `--strict`.
Can you fix it? (I will not create a PR to fix it)

allevato:
Thanks for the reminder! https://github.com/apple/swift-format/pull/276 updates the docs.

PR #270, opened 2021-11-04
Refactor "diagnostics" into "findings".

allevato:
This change addresses https://github.com/apple/swift-syntax/pull/309 and https://github.com/apple/swift-syntax/pull/325, which split `SwiftSyntaxParser` into its own module and then removed the `DiagnosticEngine` APIs (so only the `Diagnostic` type remains as the way diagnostics from the parser are sent back to the caller).

Rather than remain coupled to swift-syntax's `Diagnostic` type, this change creates a similar `Finding` type (and related types) to describe the "lint findings" that tree-based rules, the pretty-printer, and the whitespace linter encounter during
formatting/linting.

This `Finding` type is the currency type for these kinds of findings/diagnostics emitted by the formatter API layer. The
`swift-format` frontend tool now adopts `DiagnosticsEngine` from https://github.com/apple/swift-tools-support-core to manage and print these, but other clients using the API could use something different entirely. Since the usage of `TSC` is limited to the executable, API users don't have to take on a large and mostly unrelated dependency.

There are still some opportunities for renaming here -- mainly replacing instances of `diagnose` in the internal code with `emitFinding` or something similar -- but I'll leave that as future cleanup.


allevato:
Tested against `swift-DEVELOPMENT-SNAPSHOT-2021-01-28-a`:

```
Test Suite 'All tests' passed at 2021-11-04 15:53:21.029.
	 Executed 578 tests, with 0 failures (0 unexpected) in 11.055 (11.084) seconds
```

cc @dylansturg

Sources/swift-format/Utilities/DiagnosticPrinting.swift
```diff
@@ -0,0 +1,37 @@
...
+import Dispatch
+import Foundation
+import TSCBasic
+
+/// The queue used to synchronize printing uninterrupted diagnostic messages.
+private let printQueue = DispatchQueue(label: "com.apple.swift-format.printDiagnosticToStderr")
+
+/// Prints a diagnostic to standard error.
+func printDiagnosticToStderr(_ diagnostic: TSCBasic.Diagnostic) {
+  printQueue.sync {
```
dylansturg, line 22: I wonder if there would be any appreciable performance difference using `async` here, especially in parallel mode. In the end, we would need to flush the print queue though (so all diagnostics get printed before exiting), which I guess would have to be handled in the Linter and Format frontends. Might not be worth the added complexity.

dylansturg (approved):

Sources/swift-format/Utilities/DiagnosticPrinting.swift
allevato: It's certainly worth trying out as follow-up work to see if there's a meaningful difference.

PR #273, opened 2021-11-12
Fix up file URL handling.

allevato:
This change unifies the use of `URL`s for file paths throughout the
frontend instead of the current mix of `URL`s and `String`s. In
addition, it aims to preserve relative URLs where possible,
improving the presentation of diagnostics; the filenames will be
printed relative to the current working directory where swift-format
is run, rather than always printing their absolute paths.

The diagnostic/finding types, however, still represent the path of a
diagnostic or finding as the `String` to be displayed or serialized,
since they are obtained from the SwiftSyntax `SourceLocationConverter`
which uses `String` to present the file path. At the API layer (the
`SwiftFormatter` and `SwiftLinter` classes), the
`SourceLocationConverter` will always be created using the input
URL's `relativePath`, so if one wishes to force absolute paths to be
produced by findings, call `absoluteURL` before passing them to the
API.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,9 @@ import Foundation
 
 let package = Package(
   name: "swift-format",
+  platforms: [
+    .macOS(.v10_11)
+  ],
   products: [
     .executable(name: "swift-format", targets: ["swift-format"]),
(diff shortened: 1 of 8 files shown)


allevato:
cc @dylansturg

dylansturg (approved):

PR #274, opened 2021-11-12
Add support for color diagnostics.

allevato:
By default, the diagnostic printer will detect whether stderr is
connected to a TTY and use ANSI color sequences if so. We use the
same colors that `swiftc` uses for its diagnostics: white for the
main text, red for error labels, magenta for warning labels, and
gray for note labels. Additionally, we use yellow to emphasize the
finding category for diagnostics that come from linter findings.

This behavior can be controlled manually using the
`--color-diagnostics/--no-color-diagnostics` flag pair.


allevato:
cc @dylansturg

dylansturg (approved):

PR #275, opened 2021-11-12
Remove the legacy entry point.

allevato:
We've supported subcommands via `swift-argument-parser` for a few
versions now, so remove the old `-m/--mode` flag that was only
kept for compatibility with very early releases.


allevato:
cc @dylansturg

dylansturg (approved):

PR #277, opened 2021-12-09
Sort imports case-insensitive

ahoppen:
Sort imports case-insensitive instead of case-sensitive as that seems more natural to me. Unless using case-sensitive sorting was a deliberate decision @allevato.

rdar://86096207 [SR-15565]

### DIFF
--- a/Sources/SwiftFormatRules/OrderedImports.swift
+++ b/Sources/SwiftFormatRules/OrderedImports.swift
@@ -13,6 +13,12 @@
 import SwiftFormatCore
 import SwiftSyntax
 
+fileprivate extension String {
+  func lexicographicallyPrecedesCaseInsensitive(_ other: String) -> Bool {
+    return self.lowercased().lexicographicallyPrecedes(other.lowercased())
+  }
+}
+
 /// Imports must be lexicographically ordered and logically grouped at the top of each source file.
 /// The order of the import groups is 1) regular imports, 2) declaration imports, and 3) @testable
 /// imports. These groups are separated by a single blank line. Blank lines in between the import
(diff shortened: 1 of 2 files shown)


allevato:
Lexicographical ordering was an intentional choice, but it may be influenced by our own primary use cases (Swift module names auto-derived from Bazel build targets all start with lowercase letters due to our repo path structure, so this conveniently separates system frameworks from our own modules; and it matches the order produced by companion tools that sort those build targets also using lexicographical ordering).

Since `String.lowercased()` in Swift isn't locale-sensitive, I'd be fine with adding case-insensitive `import` ordering as a _configuration option,_ but I'd like it to be off by default so that we don't change the behavior of the default configuration from what it is today.

grigorye:
I wonder if the original problem is related to the fact that capital letters are put first in lexicographical order, not to case-sensitivity per se. Introducing case insensitivity as a solution would break the (in my opinion) "expected" sorting in cases of grouping by common "namespace"-like prefix, in the presense of another prefix that just happens to include the first prefix. Personally I would keep case-sensitive sorting there in favor of the proper grouping.

Below is just an illustration of the idea that the proper grouping depends on case-sensivity in some cases:
```
extension String {
    func lexicographicallyPrecedesCaseInsensitive(_ other: String) -> Bool {
        return self.lowercased().lexicographicallyPrecedes(other.lowercased())
    }
}

let names = [
    "App",
    "AppCore",
    "AppleServices",
    "AppViews",
]

print("case-insensitive:")
print(names.sorted(by: { $0.lexicographicallyPrecedesCaseInsensitive($1) }).joined(separator: "\n"))
print("")
print("case-sensitive:")
print(names.sorted(by: { $0.lexicographicallyPrecedes($1) }).joined(separator: "\n"))
```

Output:
```
case-insensitive:
App
AppCore
AppleServices
AppViews

case-sensitive:
App
AppCore
AppViews
AppleServices
```

ahoppen:
Closing this because I probably won’t have enough time to add a configuration option. If anyone wants to pick this up, feel free to do so.

PR #279, opened 2022-01-09
UseSingleLinePropertyGetter: respect `async` and `throws` on getters

cbjeukendrup:
Resolves: SR-15671

With this patch, the `UseSingleLinePropertyGetter` rule will take `async` and `throws` on getters into account. When one of those is present, the rule won't be applied; otherwise, the semantics would be changed, by removing `async` and/or `throws`.
The tests are extended accordingly.


allevato (approved):
This looks great, thanks for the fix!

PR #280, opened 2022-01-10
Fix the way to parse a document comment of ValidateDocumentationComments rule

griffin-stewie:
This PR fixes [this issue](https://bugs.swift.org/browse/SR-15711).

## Current

A current implementation the way to parse a document comment is

1. take a first sentence from document comment is first component separated by period character `.` from whole doc comment.
2. Iterates to all the comments after the one sentence summary to find the parameter(s) return tags and get their description.

## Problem

If doc comment has only one period in the middle of description of parameters, you will have to parse the tags from the missing strings halfway through the process.

Let's say here is a function that I want to lint bellow.

```swift
/// This is a great function
/// - Parameters:
///   - arg1: This is a arg1.
///   - arg2: This is a arg2
public func warning(arg1: String, arg2: String) {
    // do something
}
```

This doc comment occurs swift-format lint warning 

```
[ValidateDocumentationComments]: change the parameters of warning's documentation to match its parameters
```

### What happened?

Parsed doc comment is like this.

```
▿ ParseComment
  ▿ oneSentenceSummary : Optional<String>
    - some : "This is a great function\n - Parameters:\n   - arg1: This is a arg1"
  ▿ commentParagraphs : Optional<Array<String>>
    ▿ some : 1 element
      - 0 : " - arg2: This is a arg2"
  ▿ parameters : Optional<Array<Parameter>>
    - some : 0 elements
  - throwsDescription : nil
  - returnsDescription : nil
```

It means, this doc comment does not contain any parameter description even it exists. so linter say "the documentation to match its parameters"

## Another case

Code

```swift
/// This doc comment looks OK
/// - Parameters:
///   - arg1: This is an arg1
///   - arg2: This is an arg2.
///
public func ok(arg1: String, arg2: String) {
    // do something
}
```

ParseComment contains

```
▿ ParseComment
  ▿ oneSentenceSummary : Optional<String>
    - some : " This doc comment looks OK\n - Parameters:\n   - arg1: This is a arg1\n   - arg2: This is a arg2"
  ▿ commentParagraphs : Optional<Array<String>>
    - some : 0 elements
  ▿ parameters : Optional<Array<Parameter>>
    - some : 0 elements
  - throwsDescription : nil
  - returnsDescription : nil
```

This case no check at all because it only contains `oneSentenceSummary`. It looks all OK because no warnings from user view. If a user added one more parameter on this method and won't update doc comment, swift-format still no warnings.

## How I fixed

The key is parsing doc comment structure first. After that, takes the first sentence of the comment from commentParagraphs. ParseComment always contains structure of contents like parameters' description etc correctly.

How If a single sentence summary is the only documentation, parameter(s) and returns tags may be ommitted is changed. All the plain sentences in the doc comment are parsed into `commentParagraphs`, so never happened `commentInfo.oneSentenceSummary != nil && commentInfo.commentParagraphs!.isEmpty`.

## Any Changed Behavior

No.


griffin-stewie:
@swift-ci Please smoke test

allevato:
I believe the underlying issue here was fixed when we switched to the Markdown-based comment parser (https://github.com/apple/swift-format/pull/560). If you still see problems, feel free to re-open this with a fix based on the new implementation.

PR #281, opened 2022-03-10
Update Package.swift for Swift 5.6 release.

thii:


circAssimilate:
Is there anything I can do to help with this? Or are there any recommended workarounds while we wait for the 5.6 release?

michaellenaghan:
Without this patch, or something like it, building the 5.6 branch using Xcode 13.3 produces this error:

```
[0/1] Planning builderror: the executable 'swift-format' requires macos 10.11, but 
depends on the product 'ArgumentParser' which requires macos 10.15; consider 
changing the executable 'swift-format' to require macos 10.15 or later, or the product 
'ArgumentParser' to require macos 10.11 or earlier.
```

allevato:
As of right now, we're going to need https://bugs.swift.org/browse/SR-15989 to be resolved before we can release a 5.6 compatible version, to fix the SPM dependency issue. https://github.com/apple/swift-syntax/pull/372 should fix that; they'll need to cut a new release after that.

kylebrowning:
Looks like this should be ready @allevato

allevato:
This has been superseded by https://github.com/apple/swift-format/pull/286 (to use the newly tagged `0.50600.1` swift-syntax release and a different swift-argument-parser release.

I'll be tagging the new swift-format release momentarily.

PR #282, opened 2022-03-16
Remove override for non existent type

sergiocampama:


compnerd:
I didn't see this before, but #289 I believe is what you were suggesting @allevato.

allevato:
Merged https://github.com/apple/swift-format/pull/289 instead, since it switches this over to `ConstrainedSugarTypeSyntax`.

PR #287, opened 2022-03-19
Utilities: adjust `FileHandle` extension for Windows

compnerd:
TSCBasic provides an extension for `FileHandle` on Windows which
conforms to `WritableByteStream` which results in a duplicate protocol
conformance in the extension in swift-format.  Extract the conformance
under a guard to permit building swift-format on Windows once again.


compnerd:
Superseded by #290

PR #290, opened 2022-03-21
Use a wrapper type instead of retro-conforming `FileHandle` to `TextOutputStream`.

allevato:
TSCBasic has a Windows-only retroactive `FileHandle: WritableByteStream`
conformance which confers `TextOutputStream` conformance, conflicting with the
one in our package. Instead of trying to reconcile this across all platforms,
just replace our retroactive conformance with a new wrapper type which can
never conflict with anything from an outside package.

### DIFF
--- a/Sources/swift-format/Frontend/FormatFrontend.swift
+++ b/Sources/swift-format/Frontend/FormatFrontend.swift
@@ -39,7 +39,7 @@ class FormatFrontend: Frontend {
       return
     }
 
-    var stdoutStream = FileHandle.standardOutput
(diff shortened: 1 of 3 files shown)


allevato:
cc @compnerd (I can't add you directly as a reviewer)

I chose this approach instead of https://github.com/apple/swift-format/pull/288 to avoid a fragile retroactive conformance entirely. PTAL and then I'll merge this and cherry-pick it into 5.6, and tag a new release.

compnerd (approved):
Thanks, this definitely is much better!  I assume that this will go to both 5.6 and main.

PR #293, opened 2022-03-22
Delete XCTest manifests.

allevato:
Test discovery has been supported on non-Darwin platforms long
enough that we can drop these.


allevato:
@swift-ci test

PR #295, opened 2022-04-16
Don't emit AlwaysUseLowerCamelCase for decls that `override`

dylansturg:
Specifically this applies to function and variable decls. When one of those is overriding a decl in a super class, then this specific instance of the decl cannot be renamed to fix the issue. The original decl in the super class must be renamed. The finding is still raised if the super class is linted.

Raising the finding in every subclass is noisy and frustrating, since it can't really be fixed in the subclass.


allevato (approved):
If only we could do this for protocol requirements, too (without semantic info).

PR #296, opened 2022-04-21
Don't overwrite files if they are unchanged

myyra:
Always writing to the file, even if unchanged, updates the modified time and prevents incremental builds in Xcode. This PR adds a simple equal check before actually writing anything, preserving timestamps if nothing in the output was changed.

I don't remember this being an issue for me before, but I also couldn't find any change that would've started to cause this. In any case, this completely resolves the problem.

In my testing with a small folder of files, this change also makes the formatting slightly faster.


allevato (approved):
This is a nice improvement, thank you!

myyra:
Would it be possible to cherry-pick this into the 5.6 branch/release? We'd love this fix into production soon 🙏

PR #374, opened 2022-05-23
Cherry-pick "Don't overwrite files if they are unchanged" to release/5.6

myyra:
Cherry-picked the commit from #296 so that it can be used with the latest Xcode.


allevato (approved):
Thanks!

PR #380, opened 2022-06-28
Check for else statement during where clause substitution

chuckluck44:
### Summary
The `UseWhereClausesInForLoops` rule fails to check for an `else` keyword when replacing a single `if` statement. The formatted statement drops the behavior defined in the `else` body completely. 

### Example
```swift
for i in [0, 1, 2, 3] {
   if i > 30 {
     print(i)
   } else {
     print(i)
   }
 }

// Is formatted as:

for i in [0, 1, 2, 3] where i > 30 {
   print(i)
 }
```

### Solution
This fix adds an additional check to ensure that there is no `else` case attached to the `if` statement. Unit tests were updated to cover this behavior.


allevato (approved):
This looks straightforward and obviously correct. Thanks for the fix!

PR #383, opened 2022-07-20
Add fix and tests for postfix pound ifs

### LINKED ISSUE #303: [SR-15305] Postfix #if expressions want necessary newline removed
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-15305      |
   |Radar             | None         |
   |Original Reporter | @klanchman      |
   |Type              | Bug    |

<details>
  <summary>Environment</summary>

-   macOS 11.6 (20G165)

-   swift-format 0.50500.0, release configuration

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | swift-format    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 06f5c44e7c1b5b10eb313c28ce66331b

  </details>

**Issue Description:**

When using swift-format in `release` configuration with default settings, using [postfix `#if` expressions](https://github.com/apple/swift-evolution/blob/main/proposals/0308-postfix-if-config-expressions.md) causes swift-format to remove a necessary newline.

Using the code from the evolution proposal, `swift-format lint` outputs a warning:

``` java
VStack {
  Text("something") <<< (swift-format) warning: [RemoveLine]: remove line break
  #if os(iOS)
  .iOSSpecificModifier()
  #endif
  .commonModifier()
}
```

Formatting results in code that causes a compiler error:

``` java
VStack {
  Text("something")#if os(iOS) <<< (Swift) error: Consecutive statements on a line must be separated by a newline
  .iOSSpecificModifier()
  #endif
  .commonModifier()
}
```

DavidBrunow:
Fixes #303 

This solution feels kind of messy and I'd be happy to clean it up some if y'all have any guidance. I branched off of `release/5.6` since this issue exists there, please let me know if using a different base is better.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1308,7 +1308,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
       before(tokenToOpenWith.nextToken, tokens: .break(breakKindClose, newlines: .soft), .close)
     }
 
-    if Syntax(node).isPostfixIfConfig {
+    if isNestedInPostfixIfConfig(node: Syntax(node)) {
       before(
         node.firstToken,
         tokens: [
@@ -3468,7 +3468,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
       if let calledMemberAccessExpr = calledExpression.as(MemberAccessExprSyntax.self) {
         if calledMemberAccessExpr.base != nil {
-          if Syntax(calledMemberAccessExpr).isPostfixIfConfig {
+          if isNestedInPostfixIfConfig(node: Syntax(calledMemberAccessExpr)) {
             before(calledMemberAccessExpr.dot, tokens: [.break(.same, size: 0)])
           } else {
             before(calledMemberAccessExpr.dot, tokens: [.break(.contextual, size: 0)])
@@ -3497,9 +3497,8 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 }
 
-extension Syntax {
-  var isPostfixIfConfig: Bool {
-    var this: Syntax? = self
+private func isNestedInPostfixIfConfig(node: Syntax) -> Bool {
+    var this: Syntax? = node
 
     while this?.parent != nil {
       if this?.parent?.is(PostfixIfConfigExprSyntax.self) == true {
@@ -3510,7 +3509,6 @@ extension Syntax {
     }
 
     return false
-  }
 }
 
 extension Syntax {
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1308,7 +1308,15 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
       before(tokenToOpenWith.nextToken, tokens: .break(breakKindClose, newlines: .soft), .close)
     }
 
-    if let condition = node.condition {
+    if Syntax(node).isPostfixIfConfig {
+      before(
+        node.firstToken,
+        tokens: [
+          .printerControl(kind: .enableBreaking),
+          .break(.reset),
+        ]
+      )
+    } else if let condition = node.condition {
       before(condition.firstToken, tokens: .printerControl(kind: .disableBreaking))
       after(
         condition.lastToken,
@@ -3460,7 +3468,11 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
       if let calledMemberAccessExpr = calledExpression.as(MemberAccessExprSyntax.self) {
         if calledMemberAccessExpr.base != nil {
-          before(calledMemberAccessExpr.dot, tokens: [.break(.contextual, size: 0)])
+          if Syntax(calledMemberAccessExpr).isPostfixIfConfig {
+            before(calledMemberAccessExpr.dot, tokens: [.break(.same, size: 0)])
+          } else {
+            before(calledMemberAccessExpr.dot, tokens: [.break(.contextual, size: 0)])
+          }
         }
         before(calledMemberAccessExpr.dot, tokens: beforeTokens)
         after(expr.lastToken, tokens: afterTokens)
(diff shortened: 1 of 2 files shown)


allevato (commented):
Thanks for looking into this! This looks like a great start; I've left a couple comments about extra things we should make sure to consider.

Typically we like to have changes made only on the `main` branch, and then we can cherry-pick them into the release branches. But the main branch has fallen behind the latest development snapshots until https://github.com/apple/swift-format/pull/384 is merged, so feel free to keep it on the 5.6 branch for now until we have that sorted out and then you can rebase it and use a snapshot toolchain to test it.

allevato:
> But the main branch has fallen behind the latest development snapshots until #384 is merged

That PR has now been merged, so if you make the requested changes and rebase on top of `main`, I can help with any testing to make sure it works on that branch as well.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1289,7 +1289,21 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
       before(tokenToOpenWith.nextToken, tokens: .break(breakKindClose, newlines: .soft), .close)
     }
 
-    if let condition = node.condition {
+    if let previousToken = node.previousToken,
+       node.condition?.nextToken?.text == ".",
+       (previousToken.text == ")" || previousToken.text == "}") {
```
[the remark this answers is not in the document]
DavidBrunow: I’ll check those out, thanks for the pointers!

DavidBrunow:
> > But the main branch has fallen behind the latest development snapshots until #384 is merged
> 
> That PR has now been merged, so if you make the requested changes and rebase on top of `main`, I can help with any testing to make sure it works on that branch as well.

I finally found some time to work on this. I think my solution is greatly improved but I'm not sure if the output is what is expected in all cases -- I'd appreciate feedback on that.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -3460,7 +3468,11 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
       if let calledMemberAccessExpr = calledExpression.as(MemberAccessExprSyntax.self) {
         if calledMemberAccessExpr.base != nil {
-          before(calledMemberAccessExpr.dot, tokens: [.break(.contextual, size: 0)])
+          if Syntax(calledMemberAccessExpr).isPostfix {
```
DavidBrunow, line 3471: Thinking about it more, this logic is probably incorrect and won't indent postfix member access even when there isn't a `#if`. I'll add a scenario to test that if one doesn't already exist.

Without this conditional I saw the following output:

```swift
...
  Text("")
  #if os(iOS)
    .iOSSpecificModifier()  
      .anotherModifier() // this line has extra indentation
  #endif
  .commonModifier()
...

File: Tests/SwiftFormatPrettyPrintTests/IfConfigTests.swift [hunk 3]
```diff
@@ -230,4 +230,68 @@ final class IfConfigTests: PrettyPrintTestCase {
...
+        .accessibilityValue(
+          binding.wrappedValue == true
+            ? "On" : "Off"
+        )
+      }
+
+      """
+
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 45)
+  }
```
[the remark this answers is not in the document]
DavidBrunow: Added here: https://github.com/apple/swift-format/pull/383/files#diff-e366fdb32497e39bd5e8e8f666541fb18fbded18ad436459b26665555fd1e190R294

Tests/SwiftFormatPrettyPrintTests/IfConfigTests.swift
```diff
@@ -230,4 +230,68 @@ final class IfConfigTests: PrettyPrintTestCase {
...
+        .commonModifier()
+      }
+      """
+
+    let expected =
+      """
+      VStack {
+        Text("something")
+        #if os(iOS)
+          .iOSSpecificModifier()
```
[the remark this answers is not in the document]
DavidBrunow: Added here: https://github.com/apple/swift-format/pull/383/files#diff-e366fdb32497e39bd5e8e8f666541fb18fbded18ad436459b26665555fd1e190R261

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -3485,6 +3497,22 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 }
 
+extension Syntax {
+  var isPostfixIfConfig: Bool {
```
allevato, line 3501: I'm a little wary of making this an extension on `Syntax`, since it makes it look like a natural API provided by swift-syntax (I know that's kind of the point of extensions, but in this case, the API is specialized enough that it doesn't feel like it should be given that kind of prominence).

Could you make this just a file-level private function that just takes the node as an argument, instead? Named something like `isNestedInPostfixIfConfig` (the current name reads like it's checking that the node *itself* is the postfix-if-config, rather than checking that it's a descendant of one).

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -3485,6 +3497,22 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+    var this: Syntax? = self
+
+    while this?.parent != nil {
```
allevato, line 3504: Do we need to check all the ancestors, not just the immediate parent? Is this required to make nested clauses have the right indentation? (If it is, that's fine, I just wanted to verify the reason.)

allevato (commented):
This looks great (and the output is what I would expect). Just one minor request about API naming and I think we're good to go.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
DavidBrunow: I pushed this change but haven't tested it yet -- I can't get `main` to build so I'll either need to figure that out or cherry-pick to the 5.6 branch. I hope to get to that later today.

allevato: I just pulled it down and tested against `main`, and everything looks good, so I'll go ahead and merge and save you the trouble. Thanks again!

PR #384, opened 2022-07-22
Fix swift-format In Response to SwiftSyntax Changes

CodaFi:
There's been a bit of churn in the mainline of swift-syntax lately which has broken swift-format. Repair the build.


allevato (approved):
Sorry for the delay, I've been mostly off the grid for several days.

Thanks for doing this update! This all looks good; I don't currently have access to my workstation to do any testing of my own, so as long as all the tests pass against a recent development snapshot, feel free to merge it.

allevato:
I had a chance to test this and everything passes except for two tests that are fixed by #385 and #386 (which apparently I uploaded back in June but never created the PRs for). So I'll go ahead and merge this along with those two. Thanks again!

PR #388, opened 2022-08-08
Bump platform requirement

### LINKED ISSUE #387: Incompatible requirements when building on release/5.6
Building the 5.6 branch using Xcode 13.3 produces this error:

> the executable 'swift-format' requires macos 10.11, but depends on the product 'TSCBasic' which requires macos 10.13; consider changing the executable 'swift-format' to require macos 10.13 or later, or the product 'TSCBasic' to require macos 10.11 or earlier.

stephencelis:
Fixes #387.

The swift-tools-support-core dependency bumped its platform requirements, so swift-format should follow suit: https://github.com/apple/swift-tools-support-core/pull/329

Alternatively we could lock the dependency to an older version with more relaxed platform requirements.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.4
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift.org open source project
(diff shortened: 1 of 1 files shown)


allevato:
Thanks for the PR! But I think given https://github.com/apple/swift-format/issues/387#issuecomment-1221669953, we can close this out without making any changes here.

PR #390, opened 2022-08-21
Fix build errors, test failures, and move to non-deprecated APIs in latest swift-syntax.

allevato:


allevato:
Tested with the 2022-08-18 snapshot from swift.org:

```
$ TOOLCHAINS=org.swift.57202208181a SWIFTCI_USE_LOCAL_DEPS=1 swift test
...
Test Suite 'All tests' passed at 2022-08-21 16:18:35.650.
	 Executed 584 tests, with 0 failures (0 unexpected) in 5.911 (5.925) seconds
```

PR #395, opened 2022-09-10
Adopt the Swift Parser

CodaFi:
Adopt the new Swift parser in swift-format

This would otherwise be a fairly straightfoward patchset, but there's some existing tests that are failing on mainline that I've gone ahead and inserted explicit skips for. Tests that are expecting "unknown nodes" have also been skipped because the new parser never produces them. Finally, this patchset actually resolves a regression in mainline by properly attaching unexpected trivia text to the correct side of the type AST for `(\u{feff})`, so we recover a missing diagnostic in `ReturnVoidInsteadOfEmptyTupleTests`.

### DIFF
--- a/Sources/SwiftFormat/SwiftFormatter.swift
+++ b/Sources/SwiftFormat/SwiftFormatter.swift
@@ -16,7 +16,7 @@ import SwiftFormatCore
 import SwiftFormatPrettyPrint
 import SwiftFormatRules
 import SwiftSyntax
-import SwiftSyntaxParser
+import SwiftParser
 
 /// Formats Swift source code or syntax trees according to the Swift style guidelines.
 public final class SwiftFormatter {
@@ -64,8 +64,8 @@ public final class SwiftFormatter {
     if FileManager.default.fileExists(atPath: url.path, isDirectory: &isDir), isDir.boolValue {
       throw SwiftFormatError.isDirectory
     }
-    let sourceFile = try SyntaxParser.parse(url, diagnosticHandler: parsingDiagnosticHandler)
     let source = try String(contentsOf: url, encoding: .utf8)
+    let sourceFile = try Parser.parse(source: source)
     try format(syntax: sourceFile, assumingFileURL: url, source: source, to: &outputStream)
   }
 
@@ -87,8 +87,7 @@ public final class SwiftFormatter {
     to outputStream: inout Output,
     parsingDiagnosticHandler: ((Diagnostic) -> Void)? = nil
   ) throws {
-    let sourceFile =
-      try SyntaxParser.parse(source: source, diagnosticHandler: parsingDiagnosticHandler)
+    let sourceFile = try Parser.parse(source: source)
     try format(syntax: sourceFile, assumingFileURL: url, source: source, to: &outputStream)
   }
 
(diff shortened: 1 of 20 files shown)


Sources/SwiftFormat/SwiftFormatter.swift
```diff
@@ -54,8 +54,7 @@ public final class SwiftFormatter {
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func format<Output: TextOutputStream>(
     contentsOf url: URL,
-    to outputStream: inout Output,
-    parsingDiagnosticHandler: ((Diagnostic) -> Void)? = nil
```
allevato, line 58: I have mixed feelings about dropping this entirely. For the overload that takes a `SourceFileSyntax`, it makes sense to assume that the user has run the diagnostic-generating visitor over the tree first if they're interested in those. But the overloads that take source text or a URL are a nice convenience and I don't think we want them to silently drop diagnostics.

One idea I had was to change the callback to take the syntax tree instead and the caller could get the diagnostics that way, but by that point it's more natural for the user to just parse the file themselves and run the visitor. What I was doing in my branch was to update the callback to take the `Diagnostic` and `SourceLocation` and then run the visitor here to send that information back.

WDYT?

Tests/SwiftFormatPrettyPrintTests/AsExprTests.swift
```diff
@@ -1,5 +1,9 @@
+import XCTest
+
 final class AsExprTests: PrettyPrintTestCase {
-  func testWithoutPunctuation() {
+  func testWithoutPunctuation() throws {
+    throw XCTSkip("As expression grouping does not account for new sequence expression structure.")
```
allevato, line 5: Let's wait for https://github.com/apple/swift-syntax/pull/619 to be merged so that we can handle all of these properly without major formatting regressions.

allevato (commented):
Ah, I wish we had synced up! I have a branch locally that was doing the same update but I was waiting for https://github.com/apple/swift-syntax/pull/619 to be merged so I could finish it.

I like your changes to `TokenStreamCreator` better than mine though, so let's keep this one. Just a couple questions about handling some cases that `SwiftParser` does differently.

Tests/SwiftFormatPrettyPrintTests/AsExprTests.swift
CodaFi: To be clear, this has already regressed regardless of parsing backend as of https://github.com/apple/swift/commit/ed34dfc637732c600f9143fde6cee56a79848e9e

Sources/SwiftFormat/SwiftFormatter.swift
CodaFi: I agree, this is definitely a convenient spot to run the diagnostic post-pass. I'll try to restore this.

Tests/SwiftFormatPrettyPrintTests/AsExprTests.swift
allevato: Yes, you're right. We _could_ fix that by lifting swift-format's operating folding implementation into SwiftFormatCore instead of SwiftFormatPrettyPrint but I don't think it's worth the churn since https://github.com/apple/swift-syntax/pull/619 is on the horizon.

Ok, once you've made the change you mentioned in the other comment thread, we can land this and then do follow-ups.

CodaFi: Very much agreed. Excited to be able to have a common operator folding library available for y'all. Thanks Tony.

allevato:
Thank you! The new parser makes swift-format development immeasurably more approachable. 🎉

PR #398, opened 2022-09-16
Don't transform `Optional<T>` -> `T?` if it would cause a performance issue

### LINKED ISSUE #304: [SR-15204] UseShorthandTypeNames should not replace Optional<X> with X? on stored property declarations
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-15204      |
   |Radar             | None         |
   |Original Reporter | @karwa      |
   |Type              | Bug    |

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | swift-format    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 640847747f502d1efb6fc7028d0cd127

  </details>

**Issue Description:**

It isn't always just a shorthand; it can have meaningfully different behaviour. See: SR-11777 7 and discussion at <https://forums.swift.org/t/what-is-a-variable-initialization-expression/52132>

 

Currently, the only alternative is to disable this rule entirely, for all types, everywhere. That's less than ideal: I'd like to enable this rule and to use shorthand names when they *really are* just shorthands.

allevato:
Don't use shorthand syntax for `Optional<T>` if all of the following are true:

*   It is the type of a variable (file-, function-, or type-scoped)
*   It is mutable (`var`, not `let`)
*   It is not a computed property
*   It does not have an initializer expression

If all of these are true, shortening it to `T?` would introduce a useless nil-initialization that can be problematic in situations that want to optimize performance (see https://forums.swift.org/t/what-is-a-variable-initialization-expression/52132).

Fixes #304.

### DIFF
--- a/Sources/SwiftFormatRules/UseShorthandTypeNames.swift
+++ b/Sources/SwiftFormatRules/UseShorthandTypeNames.swift
@@ -68,6 +68,10 @@ public final class UseShorthandTypeNames: SyntaxFormatRule {
         trailingTrivia: trailingTrivia)
 
     case "Optional":
+      guard !isTypeOfUninitializedStoredVar(node) else {
+        newNode = nil
+        break
+      }
       guard let typeArgument = genericArgumentList.firstAndOnly else {
         newNode = nil
         break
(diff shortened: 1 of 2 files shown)


allevato:
@karwa , does this address all the situations you're aware of?

allevato:
Going to merge this as-is for now; let me know if there's anything that doesn't work for you!

karwa:
Thanks @allevato! I think this is fine; I'll let you know if there's anything else.

PR #401, opened 2022-09-17
Use the `SwiftOperators` module to fold sequence expressions.

allevato:
This change deletes the custom folding implementation that was used previously in swift-format. Folding was previously only performed before the pretty-printing stage (after the "rules" stage), but now folding occurs before the folding stage, allowing us to write rules that are aware of operator precedence.

The exception to the above is that if you use the `format` or `lint` APIs that take a `SourceFileSyntax`, you must manually fold the syntax tree first (any `SequenceExpr`s left in the tree will trap).


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -3241,8 +3223,10 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
     }
     if let binaryOperator = operatorExpr.as(BinaryOperatorExprSyntax.self) {
       let operatorText = binaryOperator.operatorToken.text
-      if let precedence = operatorContext.infixOperator(named: operatorText)?.precedenceGroup,
-        precedence === operatorContext.precedenceGroup(named: .assignment)
+      // TODO: `OperatorTable` doesn't give us a way to look up the precedence group of an
+      // operator, so we have to hard-code some operators instead of relying on the precedence
+      // groups as we did before. Look at adding this API to `OperatorTable`.
+      if operatorText.hasSuffix("=") && !nonAssignmentOperatorsEndingInEqual.contains(operatorText)
```
allevato, line 3229: @DougGregor Any opposition to adding a precedence group lookup function to `OperatorTable` so that we can move back to the more general way of writing this (similarly on lines 3297 and 3348)?

DougGregor: I think adding a precedence group lookup function is a great idea. Do you want to come up with a PR against swift-syntax to add the API you like?

DougGregor:
I browsed through the various changes here, and they look great. I'd love to hear how the SwiftOperators library is working out for you, and please feel free to file issues or ping me if/when you run into problems.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
allevato: Already on it 🙂

allevato: Thanks for the quick merge! I've switched back to the precedence-group-based behavior in this PR. Aside from that, `SwiftOperators` is doing everything I need without any other changes. 👍🏻

Issue #392, opened 2022-08-25
Formatting issue with postfix if configs on `main`


DavidBrunow:
I recently worked on #383 but noticed some behavior that I think is incorrect with the fix. 

Given this code:

```swift
  EmptyView()
    .padding([.vertical])
  #if os(iOS)
    .iOSSpecificModifier()
  #endif  
```

The formatter incorrectly does not indent the `.padding([.vertical])` on line 2:

Output:

```swift
  EmptyView()
  .padding([.vertical])
  #if os(iOS)
    .iOSSpecificModifier()
  #endif  
```

I have a potential fix and will put up a pull request soon.

ahoppen:
Tracked in Apple’s issue tracker as rdar://126948251

PR #402, opened 2022-09-17
Fix more postfix pound if scenarios

DavidBrunow:
Demonstrates the issue in #392 and provides a potential fix. The fix feels fragile but from what I understand from the syntax tree I'm not sure how to make it less fragile.

My [previous pull request](https://github.com/apple/swift-format/pull/383) only looked up the tree to see if an ancestor was a postfix pound if but this is not sufficient to format properly because a child node having a postfix pound if ancestor does not always mean that that child is nested inside of the postfix pound if. This change traverses back down the tree to ensure that the child node is nested properly.

While typing that out I might have had an epiphany about a better way to solve this, if so I'll update. Either way I'd like to get the test scenarios out there to document those failures.


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -3510,18 +3510,40 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
-      this = this?.parent
     }
 
     return false
+  }
+
+  var this: Syntax? = Syntax(node)
+
+  while this?.parent != nil {
+    if this?.is(TupleExprElementSyntax.self) == true {
```
DavidBrunow, line 3531: This handles the scenario [in this test](https://github.com/apple/swift-format/pull/402/files#diff-e366fdb32497e39bd5e8e8f666541fb18fbded18ad436459b26665555fd1e190R427)

DavidBrunow:
> While typing that out I might have had an epiphany about a better way to solve this, if so I'll update. Either way I'd like to get the test scenarios out there to document those failures.

This did not pan out. My thought was "why don't we just look for the right ancestor instead of the postfix if config ancestor?" But from what I can tell, the tree for a postfix if config looks something like this: 

`PostfixIfConfigExprSyntax` -> `IfConfigClauseSyntax` -> `IfConfigClauseListSyntax`

Given this example:

```swift
  EmptyView()
    .padding([.vertical])
    #if os(iOS)
      .iOSSpecificModifier()
    #endif
```

The top level is the `PostfixIfConfigExprSyntax`, so an ancestor of `.padding([.vertical])` is a `PostfixIfConfigExprSyntax` so I don't see a way to differentiate which things are inside the if config clause without going up to an ancestor and back down to a child again.

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
DavidBrunow: This is the most fragile part which I dislike the most. It handles the scenario in that test fine, but what other scenarios are there that this does not handle, or that this breaks?

Tests/SwiftFormatPrettyPrintTests/IfConfigTests.swift
```diff
@@ -390,4 +390,68 @@ final class IfConfigTests: PrettyPrintTestCase {
...
+      """
+
+    let expected =
+      """
+      EmptyView()
+        .padding([.vertical])
+      #if os(iOS)
+        .iOSSpecificModifier()
+      #endif
+        .commonModifier()
```
DavidBrunow, line 412: This is not handled properly by these changes because I didn't think to try adding this until I was creating this pull request. Will look into how to solve this but wanted to have it in this pull request so I didn't forget to mention it.

DavidBrunow:
> `PostfixIfConfigExprSyntax` -> `IfConfigClauseSyntax` -> `IfConfigClauseListSyntax`

I had this bit a little wrong, I believe this is correct:

`PostfixIfConfigExprSyntax` -> `IfConfigDeclSyntax` -> `IfConfigClauseSyntax` -> `IfConfigClauseListSyntax`

DavidBrunow:
This should probably be two different pull requests, I'm happy to break them up if others agree.

Tests/SwiftFormatPrettyPrintTests/IfConfigTests.swift
allevato: Have you had any luck with this one? For consistency we'd definitely like if the `#if` aligns with the preceding `.` and then the contents are indented one more level.

jesus-mg-ios (approved):

Tests/SwiftFormatPrettyPrintTests/IfConfigTests.swift
DavidBrunow: I have not yet, things at work got really busy soon after I created this. I'll take another look at it soon.

DavidBrunow: Took me a bit longer than I expected to get back to this. I think [this latest commit](https://github.com/apple/swift-format/pull/402/commits/d7b390fafce18fa8e734c196a6a90fbde430f8a0) works as intended. Let me know if I misunderstood.

I don't love the implementation I landed on so I'l be looking to improve it if I can.

DavidBrunow: I'm happier where this is now. Would appreciate any and all feedback :)

DavidBrunow: @allevato Friendly ping for awareness, not urgency, here since it took me so long to get back to this. I think it is behaving how you described now.

allevato: Sorry for the delay; I lost track of this one. These results look great, thanks for getting all these cases working so well!

allevato (approved):

PR #416, opened 2022-10-10
Migrate `generate-pipeline` to a SwiftPM build tool plug-in.

allevato:
The biggest change that isn't just the plug-in use itself is how we detect opt-in rules now. The previous version "cheated" by linking the rules themselves into `generate-pipeline` and then used some dynamic class lookup to get the actual value at runtime. This was replaced with a syntax tree scan that looks for a simple `true` or `false` return value in a static/class property named `isOptIn`. (A future version of swift-format should probably use compile-time constants for this property, when that's officially supported.)


allevato:
cc @dylansturg

PR #418, opened 2022-10-11
Correctly identify XCTest-like methods in `#if` blocks.

allevato:
The original implementation was only looking for direct members of a `ClassDecl` that were of type `FunctionDecl`, which excluded indirect children nested inside `IfConfigDecl`s.


allevato:
cc @dylansturg

dylansturg (approved):

PR #420, opened 2022-10-11
Add a quick workaround for keypaths starting with `\.?.`.

allevato:
This can and should be removed once https://github.com/apple/swift-syntax/issues/928 is fixed.

This change also adds tests for keypaths (we never had them!) and once the above issue is resolved, we can properly support wrapping for them in general. For now we just print them without any breaks, which is the behavior today (since we don't do any special processing for them).


allevato:
cc @dylansturg

dylansturg (approved):

PR #423, opened 2022-10-12
Revert the use of `generate-pipeline-plugin` on Windows platforms.

### LINKED ISSUE #422: build on windows, error: plugin process ended by an uncaught signal: 1033
`swift.exe build -c release`
```
error: plugin process ended by an uncaught signal: 1033 <command: d:\tmp\swift-format\.build\plugins\cache\generate_pipeline_plugin.exe>, <output:
>
```
image

allevato:
This temporarily removes the use of the plugin only on Windows, where it is not working properly due to a likely bug in SPM. We can keep the files checked in for now and updated using the
`Scripts/generate-all-pipelines.sh` script, and those files will be ignored on other platforms (the plugin will still be used there).

Fixes #422.


allevato:
@compnerd , can you please try this out?

compnerd:
FWIW, ignoring #424, this almost works.  swift-argument-parser doesn't build, but that seems unrelated to this?

PR #425, opened 2022-10-12
Fix ignore-unparsable-files flag so that it works.

dylansturg:
The flag was broken after SwiftSyntax switched how it represents unsupported syntax nodes, from Unknown* type nodes to Missing* nodes. Now the tool exits early whenever there are parsing diagnostics, and `--ignore-unparsable-files` disables emitting those diagnostics. Regardless of the flag's value, the formatter won't try to operate on a file with any parsing diagnostics.


Sources/SwiftFormat/Parsing.swift
```diff
@@ -40,15 +40,19 @@ func parseAndEmitDiagnostics(
...
       SourceLocationConverter(file: url?.path ?? "<unknown>", tree: sourceFile)
-    let diagnostics = ParseDiagnosticsGenerator.diagnostics(for: sourceFile)
     for diagnostic in diagnostics {
       let location = diagnostic.location(converter: expectedConverter)
       parsingDiagnosticHandler(diagnostic, location)
     }
   }
 
+  guard diagnostics.isEmpty else {
+    throw SwiftFormatError.fileContainsInvalidSyntax
```
allevato, line 54: I wonder if we should eventually ditch the `parsingDiagnosticsConsumer` model and just throw the array of diagnostics as part of this error, since we don't plan on doing anything else after that.

That's a bigger API change though so we can save it for later, to make sure there aren't any other weird side effects.

allevato (approved):

PR #426, opened 2022-10-12
Remove `try` before `Parser.parse` calls.

allevato:
As of https://github.com/apple/swift-syntax/pull/912, this call no longer throws.


allevato:
cc @dylansturg

dylansturg (approved):

PR #433, opened 2022-10-21
Update for the more restrictive return types in SyntaxRewriter

ahoppen:
Companion of https://github.com/apple/swift-syntax/pull/1003


allevato:
These are definitely an improvement, IMO!

Now that I've merged #434 to work around the trivia changes, the tests are all passing again so you should be able to rebase and make sure that this doesn't cause any other odd runtime effects.

PR #434, opened 2022-10-21
Workaround new trivia behavior in SwiftSyntax.

allevato:
As of https://github.com/apple/swift-syntax/pull/985, the parser treats all trivia following a token up to the next line break as trailing trivia, including comments. This is a change from the former logic, which treated such comments as leading trivia of the following token.

A great deal of the logic in swift-format was written in terms of that behavior, and attempts to rewrite it to use the new behavior instead have been difficult to get right in all cases. In the interest of expediency, I'm merging this workaround that shifts the trivia around to restore the old tree that we expected.

This shouldn't be considered a long-term fix; we should eventually fix the misbehavior with the new layout and drop the extra mutation.

---

This PR also includes some other fixes for API changes in SwiftSyntax:

* `visit(TokenSyntax)` now returns `TokenSyntax`
* Object literals are now `MacroExpansionExprSyntax` instead of `MacroExpansionDeclSyntax`

### DIFF
--- a/Sources/SwiftFormat/Parsing.swift
+++ b/Sources/SwiftFormat/Parsing.swift
@@ -12,6 +12,7 @@
 
 import Foundation
 import SwiftDiagnostics
+import SwiftFormatCore
 import SwiftOperators
 import SwiftParser
 import SwiftSyntax
@@ -54,5 +55,5 @@ func parseAndEmitDiagnostics(
     throw SwiftFormatError.fileContainsInvalidSyntax
   }
 
-  return sourceFile
+  return restoringLegacyTriviaBehavior(sourceFile)
 }
--- a/Sources/SwiftFormatCore/LegacyTriviaBehavior.swift
+++ b/Sources/SwiftFormatCore/LegacyTriviaBehavior.swift
@@ -1,0 +1,44 @@
+import SwiftSyntax
+
+/// Rewrites the trivia on tokens in the given source file to restore the legacy trivia behavior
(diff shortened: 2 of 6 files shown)


allevato:
cc @ahoppen — these changes were taking longer than I'd hoped, so I went with a cheap workaround for the time being with plans to revisit this later. In particular `OrderedImportsRule` and the pretty printer try to be smart when it comes to moving comments around, and that was proving to be challenging in a world where we might have comments split across different trivia.

ahoppen:
Hmm, OK. Do you think this will significantly complicate swift-format or is it just a matter of rewriting a few rules? If you’ve got good arguments in favor of the previous trivia rule, I’m happy to revisit it.

allevato:
I think the new rule is sensible; it's definitely more natural to treat end-of-line comments as belonging to the token on that line instead of the token on the next line, and the special logic we have to extract those from the following token today could be avoided.

The cases that were tricky were the opposite situation, where we had multiple line comments following a token that we wanted to move, for example:

```
if foo &&  // comment
  // next comment
  bar
{}
```

It wasn't the trivia partitioning itself that was challenging, but rather how it interacted with line breaks that we were inserting into the pretty printer command stream. I don't think we should revisit the trivia rule for these cases; instead swift-format should probably take a more holistic approach and write helpers for the situations where we care about _the trivia between two tokens_ versus _the trivia before/after a token_, and adjust accordingly.

PR #441, opened 2022-11-01
Update rules to use the operator-folded syntax tree.

allevato:
These rules/tests were broken when we switched to the new parser because they operated on the unfolded tree (and we now fold before the rules pipeline, rather than just before the pretty printer), but the tests for those rules never caught the problem because they *also* weren't folding the tree.


allevato:
cc @dylansturg

dylansturg (approved):

PR #442, opened 2022-11-01
Add a rule that prevents assignment from appearing in expression contexts.

allevato:
The main goal of this rule is to prevent "clever" code like this:

```
func foo() {
  guard someCondition else { return someProperty = someValue }
  // ...
}
```

That assignment can be easily misread as the Boolean `==` instead, and most readers wouldn't expect side effects to appear in a `return` statement like that. Just write it as two separate statements.

### DIFF
--- a/Tests/SwiftFormatRulesTests/NoAssignmentInExpressionsTests.swift
+++ b/Tests/SwiftFormatRulesTests/NoAssignmentInExpressionsTests.swift
@@ -1,0 +1,164 @@
+import SwiftFormatRules
+
+final class NoAssignmentInExpressionsTests: LintOrFormatRuleTestCase {
+  func testAssignmentInExpressionContextIsDiagnosed() {
+    XCTAssertFormatting(
+      NoAssignmentInExpressions.self,
+      input: """
+        foo(bar, baz = quux, a + b)
+        """,
+      expected: """
+        foo(bar, baz = quux, a + b)
+        """
+    )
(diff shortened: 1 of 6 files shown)


allevato:
Creating this as a draft right now since it depends on the test fix in #441 to pass.

cc @dylansturg

Tests/SwiftFormatRulesTests/NoAssignmentInExpressionsTests.swift
```diff
@@ -0,0 +1,110 @@
...
+        func foo() {
+          a += b
+        return
+        }
+        """
+    )
+    XCTAssertDiagnosed(.moveAssignmentToOwnStatement, line: 2, column: 10)
+  }
+
+  func testReturnStatementWithAssignmentHasCommentMoved() {
```
dylansturg, line 91: What about if they're doing something _silly_ like opening a multiline comment on the line with the return? I think the trailing trivia will get moved so that the return is sitting inside of the comment?

```
func foo() {
  return a = b /* blargh
    continues here */
}
```

allevato: The trivia will never break within a block comment, so the entire comment (including its line break) will be a single trivia piece. We're also still using the ~~hack~~ workaround I wrote to preserve the legacy trivia behavior, so that comment would be in the leading trivia of the `}` token.

I did tweak this logic though to make sure it handles more cases around weird comments and added tests for those, PTAL.

dylansturg (approved):

PR #444, opened 2022-11-03
Update Package.swift

### LINKED ISSUE #440: Adding support for iOS
Please add support for iOS:
platforms: [
    .iOS("13.0"),
  ]

HassanTaleb90:
Adding support for iOS 13.0 +

Fixes #440.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,6 +17,7 @@ import PackageDescription
 let package = Package(
   name: "swift-format",
   platforms: [
+    .iOS("13.0"),
     .macOS("10.15")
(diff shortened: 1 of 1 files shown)


allevato:
Thanks! This seems like a reasonable thing to allow now that the underlying parser is pure Swift.

I just tested this in Xcode and all the libraries built fine when targeting iOS; I'm running it through the CI now to make sure this doesn't introduce any odd side effects to the toolchain build.

allevato:
CI looks good!

PR #424, opened 2022-10-12
Revert generate pipeline plugin

DougGregor:
Revert https://github.com/apple/swift-format/pull/416 and https://github.com/apple/swift-format/pull/419 because they appear to be breaking Swift pull request testing.

PR #449, opened 2022-11-09
Support CommandPlugin

zunda-pixel:


Package.swift
```diff
@@ -33,6 +33,14 @@ let package = Package(
       name: "SwiftFormatConfiguration",
       targets: ["SwiftFormatConfiguration"]
     ),
+    .plugin(
+      name: "Format Source Code",
```
allevato, line 37: I think we should remove the spaces from the _target_ names and subdirectory names. Names like `FormatPlugin` and `LintPlugin` would align well with the naming we've used elsewhere.

For the `name` argument of the `.plugin`, I'm less sure. I believe this surfaces in the Xcode UI? Is there anywhere else that this string is used (like a command line) where spaces would cause problems?

Package.swift
```diff
@@ -91,7 +99,30 @@ let package = Package(
...
+        ]
+      ),
+      dependencies: [
+        .target(name: "swift-format")
+      ]
+    ),
+    .plugin(
+      name: "Lint Source Code",
+      capability: .command(
+        intent: .sourceCodeFormatting(),
```
allevato, line 117: What's the behavior of `swift package format-source-code` when there are two plugins with this intent?

I wonder if we should use something like `.custom(verb: "lint-source-code")` here instead.

Plugins/Format Source Code/plugin.swift
```diff
@@ -0,0 +1,93 @@
...
+      guard let target = target as? SourceModuleTarget else {
+        Diagnostics.warning("Skipping target “\(target.name)” because it doesn't contain Swift sources")
+        continue
+      }
+      
+      let swiftFormatExec = URL(fileURLWithPath: swiftFormatTool.path.string)
+      var swiftFormatArgs = [
+        "format",
+        target.directory.string,
+        "-r",
```
allevato, line 30: Here and throughout, can you use the `--long-form` of the flags instead, so it's easier for a reader to understand when they quickly scan it?

Plugins/Format Source Code/plugin.swift
```diff
@@ -0,0 +1,93 @@
...
+        "-p",
+        "-i",
+      ]
+      
+      if let configuration {
```
allevato, line 35: Here and elsewhere, let's avoid using new Swift 5.7 syntax for now; Package.swift still specifies a 5.6 tools version, and I'm not sure how it plays with CI (some other projects avoid it).

Plugins/Format Source Code/plugin.swift
```diff
@@ -0,0 +1,93 @@
...
+    let swiftFormatTool = try context.tool(named: "swift-format")
+    
+    // Extract the arguments that specify what targets to format.
+    var argExtractor = ArgumentExtractor(arguments)
+    let targetNames = argExtractor.extractOption(named: "target")
+    let targetsToFormat = try context.package.targets(named: targetNames)
+    
+    let configuration = argExtractor.extractOption(named: "configuration").first
+    
+    for target in targetsToFormat {
```
allevato, line 19: Instead of spawning `swift-format` once for each target, it would be more efficient to collect the directory paths in an array and then pass them all at once in the args array to just do a single invocation.

Plugins/Format Source Code/plugin.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+}
+
+#if canImport(XcodeProjectPlugin)
+import XcodeProjectPlugin
+
+extension FormatSourceCode: XcodeCommandPlugin {
+  func performCommand(context: XcodeProjectPlugin.XcodePluginContext, arguments: [String]) throws {
+    let swiftFormatTool = try context.tool(named: "swift-format")
```
allevato, line 60: This code is mostly duplicated from the function above. Can you factor out the commonalities into a file-scoped function and just have the plugins call that?

I think a single function that takes the tool returned by `context.tool`, the arguments, and the list of directories to format would do the trick.

Plugins/Format Source Code/plugin.swift
```diff
@@ -0,0 +1,93 @@
...
+    
+    var argExtractor = ArgumentExtractor(arguments)
+    let configuration = argExtractor.extractOption(named: "configuration").first
+    
+    let swiftFormatExec = URL(fileURLWithPath: swiftFormatTool.path.string)
+    var swiftFormatArgs = [
+      "format",
+      ".",
```
allevato, line 68: From the `XcodePluginContext` you can reach through `context.xcodeProject.directory`, which might be safer than assuming the working directory of the plugin will always be the project directory.

Plugins/Lint Source Code/plugin.swift
```diff
@@ -0,0 +1,92 @@
+import PackagePlugin
```
allevato, line 1: Same comments from the Format Source Code file apply here (factoring out the common implementations, making a single invocation to swift-format, accessing the Xcode project directory).

allevato (commented):
Thanks for starting this! I haven't explored the plugin space much yet but this looks useful to have.

I'll caution you that the previous time I tried adding a plugin, we had to revert it because of CI and Windows-related failures: https://github.com/apple/swift-format/pull/424. But that was a _build tool_ plugin that was generated sources to use during swift-format's own build, so that might have been different. Either way, once this is in shape, we'll see if CI has any problems with it. 🤞🏻

Package.swift
zunda-pixel: I fixed with reference to [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/76259ebe8c477ecc4e938d0831aea4c25f9ee3ee/Package%40swift-5.6.swift)

[zunda-pixel/swift-format/pull/1](https://github.com/zunda-pixel/swift-format/pull/1)

Xcode  uses [Package.targets.plugin.name](https://github.com/zunda-pixel/swift-format/blob/41cf80954c9ab0bc8b3a562d89ad788534515649/Package.swift#L103) as UI

Package.swift
zunda-pixel: I fixed it. https://github.com/zunda-pixel/swift-format/commit/8bcb0bb94af287b080aa5d631360cc24737d9ba0

```shell
swift-format$ swift package plugin --list
‘format-source-code’ (plugin ‘Format Source Code’ in package ‘swift-format’)
‘generate-manual’ (plugin ‘Generate Manual’ in package ‘swift-argument-parser’)
‘lint-source-code’ (plugin ‘Lint Source Code’ in package ‘swift-format’)
```

Plugins/Format Source Code/plugin.swift
zunda-pixel: I fix it. https://github.com/zunda-pixel/swift-format/commit/2fee7b9c1ad81955a054c02e6ca9884246b5d2b5

Plugins/Format Source Code/plugin.swift
zunda-pixel: I fixed It.

Plugins/Format Source Code/plugin.swift
zunda-pixel: I fixed it. https://github.com/zunda-pixel/swift-format/commit/60c66fc7a8404c4bbfd30506d38450042b292e5f

Plugins/Format Source Code/plugin.swift
zunda-pixel: I don't know `--long-form`. what is this?

allevato: Every flag that has a short form (like `-r`) has a longer form (`--recursive`). Using the longer ones will make the usage clearer. They're documented here: https://github.com/apple/swift-format#command-line-usage

Plugins/FormatPlugin/plugin.swift
```diff
@@ -0,0 +1,94 @@
...
+    let targetNames = argExtractor.extractOption(named: "target")
+    let targetsToFormat = try context.package.targets(named: targetNames)
+    
+    let configurationFilePath = argExtractor.extractOption(named: "configuration").first
+    
+    let sourceCodeTargets = targetsToFormat.compactMap{ $0 as? SourceModuleTarget }
+    
+    try format(
+      tool: swiftFormatTool,
+      targetDirectories: sourceCodeTargets.map(\.directory).map(\.string),
```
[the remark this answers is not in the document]
zunda-pixel: I fixed it. https://github.com/apple/swift-format/pull/449/commits/28f9876c84f1252049a5bd87b83d41fdcb4438f1

allevato (approved):
Thank you! Waiting for CI results, which will be reported at https://github.com/apple/swift-syntax/pull/1086.

allevato:
CI looks good. Can you squash the commits and re-push them so we can keep the history (even after merge) a bit cleaner?

zunda-pixel:
I cleaned up git history.
I don't know git squash well, Is it ok?

allevato (approved):
This looks great; thank you again for the contribution!

PR #452, opened 2022-12-03
Update for SwiftSyntax if/switch expression work

hamishknight:
Update for https://github.com/apple/swift-syntax/pull/1228


allevato (approved):
An elegantly small change on this side!

PR #454, opened 2022-12-11
Remove References to Unknown Syntax

CodaFi:
The new Swift parser in SwiftSyntax does not ever construct Unknown*Syntax nodes. These are a relic of the C++, which would use these nodes as stand-ins for cases where the legacy parser did not known how to build a node, or built one incorrectly. Simultaneously, the formatter was relying on unknown nodes containing verbatim text it would leave untouched. The general replacement for that case is UnexpectedSyntax. Drop the Unknown*Syntax entrypoints in the visitors, and replace refernces to "unknown syntax" with "unexpected syntax".

See also https://github.com/apple/swift-syntax/pull/1131


allevato (approved):
Thanks! Feel free to merge it when you're ready.

PR #467, opened 2023-01-05
Pin dependencies other than swift-syntax to more specific versions.

allevato:
*   swift-argument-parser is pinned to the same version range used by
    swift-syntax, and these should be kept in sync.
*   swift-tools-support-core is pinned to a specific version (not a
    range).


allevato:
Kicking off CI over at https://github.com/apple/swift-syntax/pull/1086.

ahoppen (approved):

PR #477, opened 2023-01-20
Adjustment for removed `with<childName>` functions

ahoppen:
Companion of https://github.com/apple/swift-syntax/pull/1253


allevato (approved):
This is a really nice API improvement!

PR #495, opened 2023-03-28
Adjustments to split `FunctionParameterSyntax` into multiple nodes for function parameters, closure parameters and enum parameters

ahoppen:
Companion of https://github.com/apple/swift-syntax/pull/1455

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1122,7 +1122,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
       // When it's parenthesized, the input is a `ParameterClauseSyntax`. Otherwise, it's a
       // `ClosureParamListSyntax`. The parenthesized version is wrapped in open/close breaks so that
       // the parens create an extra level of indentation.
-      if let parameterClause = input.as(ParameterClauseSyntax.self) {
+      if let parameterClause = input.as(ClosureParameterClauseSyntax.self) {
         // Whether we should prioritize keeping ") throws -> <return_type>" together. We can only do
         // this if the closure has arguments.
         let keepOutputTogether =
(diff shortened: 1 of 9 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -2606,6 +2606,24 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
   }
 
+  /// Applies formatting to a collection of enum case parameters for a decl.
+  ///
+  /// - Parameters:
+  ///    - parameters: A node that contains the parameters that can be passed to a decl when its
+  ///      called.
+  ///    - forcesBreakBeforeRightParen: Whether a break should be required before the right paren
+  ///      when the right paren is on a different line than the corresponding left paren.
+  private func arrangeEnumCaseParameterClause(
```
ahoppen, line 2616: I just duplicated this method for `EnumCaseParameterClauseSyntax`. Alternatives would be
1. Make this function take a `leftParen: TokenSyntax, isEmpty: Bool, rightParen: TokenSyntax` so the logic can be shared with `arrangeParameterClause`
2. Create a protocol that offers the three parameters described in (1) and make both `ParameterClauseSyntax` and `EnumCaseParameterClauseSyntax` conform to that

(2) seemed like quite an overkill to me and (1) seemed ugly as well. Let me know if you’ve got preferences @allevato.

allevato: This is fine; the function is small enough that duplicating it is probably better than overengineering a generic thing that would only be used in one extra place.

allevato (approved):

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
ahoppen: That’s what I thought as well. But good to hear you’re on the same page.

ahoppen:
@allevato I have needed to duplicate some more code to make sure closure and enum case parameters are correctly formatted. Could you take another look whether you think it makes sense to unify some of these methods? I still think that it might be too much overhead but you decide.

allevato:
I still think the duplication is small enough that it's fine to have it for now, especially since all of these param lists are subtly different. If I want to revisit it in the future I can do that as part of a larger refactoring, but no reason to block you on it.

Issue #494, opened 2023-03-23
Guard let in blocks is merged into the same line


pronebird:
Swift-format does not add newline before guard statement in closures which is incredibly weird:

```swift
myObject.didFinish = { [weak self] in guard let self = self else { return }

    self.didFinish?(self)
}
```

stackotter:
Feel free to test out my fix and see if it works for you :) After my fix, your snippet gets formatted like so:

```swift
myObject.didFinish = { [weak self] in
    guard let self = self else { return }

    self.didFinish?(self)
}
```

PR #498, opened 2023-04-02
Fix pretty printing of multi-statement closures (issue #494)

stackotter:
Issue #494 is related to pretty printing of multi-statement closures, and can be reproduced on `main` by formatting the following input:

```swift
{ a in a + 1
  a + 2
}
```

The pretty printer just spits out the snippet as is. This is because the first break is made elective instead of soft. Note that the closure is formatted correctly if `a in` is removed.

The fix that I propose (and have implemented) is forcing the break before the first statement to be `soft` if the closure contains multiple statements. I believe this is reasonable because there is no situation that I know of where Swift Format should format a multi-statement closure as a single-line closure (`{ a + 1; a + 2 }` is pretty un-idiomatic).

I have added a test for this change that fails on main and succeeds after my change.

This is my first contribution to Swift Format, so please let me know if there's anything I should change about this PR or future PRs (e.g. test style).


allevato (approved):
Thank you!

Will kick off CI on the companion PR in swift-syntax after my other requests are done processing and merge when it's ready.

stackotter:
Awesome, thanks 👍

PR #501, opened 2023-04-03
Fix formatting of import with multiple attributes (fixes #445) and ensure that imports never wrap

### LINKED ISSUE #445: swift-format removes space between attributes in import statement when an attribute has an argument
Example:

```
@_spi(STP) @testable import StripeCore
```

is turned into

```
@_spi(STP)@testable import StripeCore
```

stackotter:
#445 occurs because Swift Format simply just adds a space after the attribute list of import statements. `arrangeAttributeList` can be reused to correctly format the attributes of an import. However, when modifying the test for imports to include an import decl with multiple attributes, I found that attributes with arguments (e.g. `@_spi(STP)`) get wrapped when the import decl exceeds the line length. The comment included with the existing tests says that import decls shouldn't wrap (and I take this to also mean that discretionary breaks should be removed if present). Thus, I have made additional changes to ensure that import statements never wrap in any situation, please correct me if this is not the desired behaviour.

To achieve this  behaviour (see snippet below), I had to modify the `disableBreaking` printer control to be able to configure whether discretionary breaks should be allowed when breaks are suppressed (currently they are, but for the case of import decls this needs to be overridden to ensure that imports never wrap).

```swift
// Input
@_spi(
  STP
)
@testable
import testModule

// Output
@_spi(STP) @testable import testModule
```

I can see people wanting to be able to write code that wraps between attributes (as is sometimes done with `@resultBuilder` on struct decls, or property wrappers). Should this be made configurable, or is it bad style?

I also fixed the typo of `isBreakingSupressed` (to `isBreakingSuppressed`) because it was only used in 3 places or so, which imo makes a merge conflict very unlikely. I'm happy to revert that if requested.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormatPrettyPrint/PrettyPrint.swift
@@ -129,10 +129,15 @@ public class PrettyPrinter {
   private var activeBreakSuppressionCount = 0
 
   /// Whether breaks are supressed from firing. When true, no breaks should fire and the only way to
-  /// move to a new line is an explicit new line token.
-  private var isBreakingSupressed: Bool {
+  /// move to a new line is an explicit new line token. Discretionary breaks aren't suppressed
+  /// if ``allowSuppressedDiscretionaryBreaks`` is true.
+  private var isBreakingSuppressed: Bool {
     return activeBreakSuppressionCount > 0
   }
+
+  /// Indicates whether discretionary breaks should still be included even if break suppression is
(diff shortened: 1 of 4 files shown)


allevato (approved):
This looks great, thanks! I agree that the desired behavior (for now, at least) is to forbid `import`s from wrapping under any circumstances, for consistency.

I'll kick off the CI over in https://github.com/apple/swift-syntax/pull/1086 (since we don't have the checks enabled on this repo yet).

stackotter:
Great, thank you! Ignore my question about companion PR in another thread, you already explained it here 😅

PR #509, opened 2023-04-10
Remove the `swift-tools-support-core` dependency from swift-format.

allevato:
This was only being used for the diagnostics engine, which is easy enough to roll by hand and it's one less dependency to keep in sync with the rest of the toolchain.

In the future, we may want an output mode that uses the new prettier diagnostic printer from swift-syntax.


allevato:
cc @dylansturg

dylansturg (approved):

PR #532, opened 2023-05-25
Fix indentation of multiline strings when part of a larger expression.

allevato:
The stacked indentation behavior wasn't robust enough to catch other kinds of subexpressions that a multiline string could be in and our tests weren't thorough enough to catch this. The most common occurrence was when a multiline string was followed by a dotted member access, which caused the string to become unindented one level from where it should have been.


allevato:
cc @dylansturg

dylansturg (approved):

PR #533, opened 2023-05-25
Fix `try`/`await` expressions in `NoAssignmentInExpressions`.

allevato:
This rule only checked whether the _immediate_ parent of an expression like `x = y` was a `CodeBlockItem`. In cases like `try x = y`, the `try` wraps the entire expression and the `CodeBlockItem` would be the parent of that. So we look through any `TryExpr`s or `AwaitExpr`s we see when searching for the `CodeBlockItem`.


allevato:
cc @dylansturg

dylansturg (approved):

PR #534, opened 2023-05-26
Further improve multiline string formatting.

allevato:
These changes remove unnecessary grouping around multiline string literals that were forcing subexpressions to wrap in less than ideal ways. Since multiline strings force hard line breaks after the open quotes, we can remove the grouping and produce better results when complex expressions are involved. For example,

```swift
let x = """
  abc
  def
  """ + """
    ghi
    jkl
    """
```

Before this change, we were forcing breaks after the `=` and before the `+`. Now, we only do so if the open quotes would overflow the line.


allevato:
cc @dylansturg

dylansturg (approved):

PR #535, opened 2023-05-26
Allow exceptions to `NoAssignmentInExpressions`.

allevato:
This is mostly intended to carve out a really useful exception for testing: `XCTAssertNoThrow(x = try ...)` is quite a bit clearer about intent than `x = try XCTUnwrap(try? ...)` if you need the value `x` later in the test.

In an effort to not be *too* specific in this rule, I've opted to make this a configurable list of exceptions that by default contains `XCTAssertNoThrow`. So it only handles function calls right now, but this could be made more flexible in the future.


allevato:
cc @dylansturg

dylansturg (approved):

PR #539, opened 2023-06-05
Fix `async throws` function types when they appear in an expression context.

### LINKED ISSUE #538: "async throws" gets smooshed together in some cases
Before formatting:

```swift
var tasks = [(() async throws -> Int)]()
```

After formatting:

```swift
var tasks = [(() asyncthrows -> Int)]()
```

Notably, this code isn't even valid anymore. I'm using swift-format 508.0.1.

allevato:
Also unify the way we handle `async throws` now that they are all contained inside effect specifier nodes that share a common `EffectSpecifiersSyntax` protocol conformance.

Fixes #538.


allevato:
cc @dylansturg

PR #540, opened 2023-06-05
Replace deprecated code from swift-syntax

TTOzzi:
Replaced the deprecated code and also made contextual modifications to the related code.
Due to the changes in https://github.com/apple/swift-syntax/pull/1722 and https://github.com/apple/swift-syntax/pull/1728.


allevato (approved):
Thanks!

Issue #542, opened 2023-06-20
No space between macro and trailing closure


kimberninger:
When formatting a freestanding macro expression with a trailing closure, any space in front of the closure's opening brace will be removed.

For example,
```swift
#Preview {
    // ...
}
```
would be re-formatted to
```swift
#Preview{
    // ...
}
```
This also applies to macro expressions using generic parameters e.g.,
```swift
let p = #Predicate<Int> { $0 == 0 }
```
is re-formatted to
```swift
let p = #Predicate<Int>{ $0 == 0 }
```

Tested on latest status of the `main` branch (revision 8793d965ecd830bae26d86c6cddca607a3a7fa93).

PR #544, opened 2023-06-20
Insert white space before trailing closure of a macro expression

kimberninger:
This small PR fixes a bug causing the space before the trailing closure of a freestanding macro expression to be removed when running `swift-format`.

For example, the following line of code
```swift
let p = #Predicate<Int> { $0 == 0 }
```
would originally be formatted to
```swift
let p = #Predicate<Int>{ $0 == 0 }
```

In addition to resolving this issue, this PR adds some basic tests to ensure the correct behavior.

For more information and examples see also issue #542.

### DIFF
--- a/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
@@ -1296,11 +1296,23 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 
   override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+    let arguments = node.argumentList
+
+    // If there is a trailing closure, force the right parenthesis down to the next line so it
+    // stays with the open curly brace.
+    let breakBeforeRightParen =
+      (node.trailingClosure != nil && !isCompactSingleFunctionCallArgument(arguments))
+      || mustBreakBeforeClosingDelimiter(of: node, argumentListPath: \.argumentList)
+
+    before(
+      node.trailingClosure?.leftBrace,
+      tokens: .break(.same, newlines: .elective(ignoresDiscretionary: true)))
+
     arrangeFunctionCallArgumentList(
-      node.argumentList,
+      arguments,
       leftDelimiter: node.leftParen,
       rightDelimiter: node.rightParen,
-      forcesBreakBeforeRightDelimiter: false)
+      forcesBreakBeforeRightDelimiter: breakBeforeRightParen)
     return .visitChildren
   }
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
```diff
@@ -1250,6 +1250,10 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 
   override func visit(_ node: MacroExpansionExprSyntax) -> SyntaxVisitorContinueKind {
+    before(
```
allevato, line 1253: Before this, we need to pick up the same logic we're doing here: https://github.com/apple/swift-format/blob/8793d965ecd830bae26d86c6cddca607a3a7fa93/Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift#L981-L985

And then pass that value into `forcesBreakBeforeRightDelimiter` below.

allevato (commented):
Thanks for the fix! Can you address the comment below and add test cases for it, similar to how they're done for function calls?

Sources/SwiftFormatPrettyPrint/TokenStreamCreator.swift
kimberninger: Thank you for the feedback! I have pushed an update addressing the suggestion and adding another test case similar to the [corresponding one](https://github.com/apple/swift-format/blob/8793d965ecd830bae26d86c6cddca607a3a7fa93/Tests/SwiftFormatPrettyPrintTests/FunctionCallTests.swift#L276) in `FunctionCallTests.swift`.

Tests/SwiftFormatPrettyPrintTests/MacroCallTests.swift
```diff
@@ -0,0 +1,117 @@
...
+      """
+
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 50)
+  }
+
+  func testKeepWhiteSpaceBeforeTrailingClosure() {
+    let input =
+      """
+      #Preview {}
+      #Preview("MyPreview) {
```
allevato, line 28: Here and throughout below, it looks like you're missing the trailing quote after `"MyPreview`, so we can't really trust the test results (if they do pass, the parser may be fixing the literal for us automatically)

kimberninger: Thanks for the thorough attention, I would have totally missed that! Sorry for the inconvenience 🙇
The strings should now properly be closed with the latest update.

allevato (approved):
We're good to go now, thank you!

PR #545, opened 2023-06-20
Wrap keypath literals appropriately.

allevato:
We were previously just letting these get printed verbatim; at some point, this behavior changed and we were getting rid of any breaks the user had put inside them (I observed this in internal code where a keypath was broken across lines; we don't have any test cases covering that).


allevato:
cc @dylansturg

PR #558, opened 2023-07-01
Fix deprecated SwiftSyntax api warnings 

kitasuke:
main branch of SwiftSyntax renamed some properties and functions, so this PR applied them in swift-format and fix the pipeline generator.

### DIFF
--- a/Sources/SwiftFormat/Pipelines+Generated.swift
+++ b/Sources/SwiftFormat/Pipelines+Generated.swift
@@ -321,28 +321,28 @@ class LintPipeline: SyntaxVisitor {
 
 extension FormatPipeline {
 
-  func visit(_ node: Syntax) -> Syntax {
+  func rewrite(_ node: Syntax) -> Syntax {
     var node = node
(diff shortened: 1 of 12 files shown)


Sources/SwiftFormat/Pipelines+Generated.swift
```diff
@@ -321,28 +321,28 @@ class LintPipeline: SyntaxVisitor {
 
 extension FormatPipeline {
 
-  func visit(_ node: Syntax) -> Syntax {
+  func rewrite(_ node: Syntax) -> Syntax {
```
allevato, line 324: This file is generated by https://github.com/apple/swift-format/blob/main/Sources/generate-pipeline/PipelineGenerator.swift. Can you update that code and then rerun it to verify that the generated output here is correct?

kitasuke: I totally overlooked that... Addressed and combine previous commits into one.

allevato (approved):
Thanks!

PR #566, opened 2023-07-13
Remove --mode flag for configuration dump

Gray-Wind:
Looks like this flag is a leftover from older times, little above `dump-configuration` command is being used without `--mode` flag.


allevato (approved):
Thank you!

PR #567, opened 2023-07-13
Use newer equivalents of deprecated nodes names

Matejkob:
I have eliminated the usage of deprecated node names. All instances where deprecated names were used have been updated to their newer, current equivalents. 

I also have changed the names of some local properties to keep consistency with node names.


allevato (approved):
Thank you!

PR #568, opened 2023-07-16
Add XCTest exclusion comment for lint rules

kitasuke:
Added a rule exclusion of XCTest to docs just like https://github.com/apple/swift-format/blob/main/Sources/SwiftFormatRules/NeverUseForceTry.swift#L18:L19


allevato (approved):
Thanks!

PR #570, opened 2023-07-22
Fix deprecated warnings regarding swift-syntax changes

kitasuke:
Applied below changes to fix deprecated warnings
- [x] Pound keyword suffix
- [x] CollectionType method


allevato (approved):
Thank you!

PR #579, opened 2023-08-02
Fix deprecation warnings from renamed nodes / types in SwiftSyntax

ahoppen:
Fix all deprecation warnings in swift-format that arose because of node renames in swift-syntax.


allevato (approved):
Whew! Thanks for all that cleanup. 🙏🏻

PR #584, opened 2023-08-09
Migrate away from the latest deprecated APIs.

allevato:


Sources/SwiftFormatCore/Context.swift
```diff
@@ -76,7 +76,8 @@ public final class Context {
     self.fileURL = fileURL
     self.importsXCTest = .notDetermined
     let tree = source.map { Parser.parse(source: $0) } ?? sourceFileSyntax
-    self.sourceLocationConverter = SourceLocationConverter(file: fileURL.relativePath, tree: tree)
+    self.sourceLocationConverter =
+      SourceLocationConverter(fileName: fileURL.relativePath, tree: tree)
```
Matejkob, line 80: Nitpick: should it be formatted in the following way:
```suggestion
    self.sourceLocationConverter = SourceLocationConverter(
        fileName: fileURL.relativePath,
        tree: tree
    )
```

allevato: While swift-format supports that, I personally prefer breaking before the function call if it means the entire function call can remain on its own line.

PR #590, opened 2023-08-12
Make `shebang` a child of `SourceFileSyntax`

StevenWong12:
To sync changes in https://github.com/apple/swift-syntax/pull/1979


allevato (approved):

Wilfred:
I think this may have broken shebangs? See #1194.

PR #605, opened 2023-09-02
Adjust for "remark" diagnostic severity

DougGregor:


Sources/swift-format/Utilities/DiagnosticsEngine.swift
```diff
@@ -101,6 +101,7 @@ final class DiagnosticsEngine {
     case .error: severity = .error
     case .warning: severity = .warning
     case .note: severity = .note
+    case .remark: severity = .note // should we model this?
```
allevato, line 104: This is good enough as a minimal change to unblock the other PR. I can go back later and look at modeling it as a unique kind in swift-format later on. Thanks!

allevato (approved):

Sources/swift-format/Utilities/DiagnosticsEngine.swift
DougGregor: Thank you. I wasn't sure if it made sense for swift-format to model remarks or not. They're only used in narrow places in the compiler itself, and might not be interesting here.

allevato: Are there plans to have any remarks come out of the parser or just the later phases of the compiler? Because parser diagnostics are the only ones that swift-format cares about.

DougGregor: No plans to have remarks come out of the parser. At the moment, it's usually later phases providing extra information about the build itself---when a Swift module was rebuilt from its interface, which plugin is providing a macro.

PR #608, opened 2023-09-03
Default to all targets when plugin `--target` parameter missing. Fix #483

### LINKED ISSUE #483: Cannot invoke plugins from command line
Hi

I'm trying to add swift-format to our project; we aim at being able to run this manually, and from our CI pipeline.
swift-format is added as a dependency in our Package.swift. (in my tests, I'm using `0.50800.0-SNAPSHOT-2022-12-29-a` tag).

If I run "Lint Source Code" or "Format Source Code" from XCode, everything works ok.
However, If I try to run it from command line, I get the following errors:

```
$ swift package plugin lint-source-code
Building for debugging...
Build complete! (0.58s)
Error: '--recursive' is only valid when formatting or linting files
Usage: swift-format lint [--configuration <configuration>] [--assume-filename <assume-filename>] [--recursive] [--ignore-unparsable-files] [--parallel] [--color-diagnostics] [--no-color-diagnostics] [<paths> ...] [--strict]
  See 'swift-format lint --help' for more information.
error: swift-format invocation failed: NSTaskTerminationReason(rawValue: 1):64
```

Can you please help me make this work?

Thank you very much

BrianHenryIE:
When either of the swift-format plugins is run without the `--target` parameter, an error is returned. This PR implements the fix by @soumyamahunt in the [#483 discussion](https://github.com/apple/swift-format/issues/483#issuecomment-1534889674), default to all targets when none is set:

```
let targetsToFormat = targetNames.isEmpty ? context.package.targets : try context.package.targets(named: targetNames)
```

Before:

```
$ swift package lint-source-code                                          
Building for debugging...
Build complete! (2.15s)
Error: '--recursive' is only valid when formatting or linting files
Usage: swift-format lint [--configuration <configuration>] [--assume-filename <assume-filename>] [--recursive] [--ignore-unparsable-files] [--parallel] [--color-diagnostics] [--no-color-diagnostics] [<paths> ...] [--strict]
  See 'swift-format lint --help' for more information.
error: swift-format invocation failed: NSTaskTerminationReason(rawValue: 1):64
```

After:

```
$ swift package lint-source-code  
Building for debugging...
Build complete! (2.28s)
Linted the source code.
```

The test works, but maybe there's a BDD library I should be using instead. It sets up a project in a tmp dir with `swift-format` as a dependency, then confirms the error is absent from the stderr output.

In the sample SPM project I included one each of executable, library, test, plugin targets. AIUI, `context.package.targets` does not return the plugin target. I.e. as this fix stands, omitting `--target` plugin parameter and expecting "all" targets to be linted, does not lint any _plugin_ target.

I tried various ways to unit-test the plugin first. Discussion on the Swift forums says [it cannot be done](https://forums.swift.org/t/how-can-we-test-swiftpm-plugins/61279)! I symlinked the plugin file into the test target and added SPM as a dependency but got blocked by access control levels as I tried to mock a PluginContext.

I also tried to implement the tests the same as [SPM's own fixtures tests](https://github.com/apple/swift-package-manager/blob/a2bdb7207098189deb9141b09fbf7765bae37c75/Tests/FunctionalTests/PluginTests.swift) but didn't manage to get that working either.

With the current setup (write Package.swift to a temp dir, run `swift package ...` in a NSTask Process) I was only able to verify the output of stdout and stderr. I tried a few ways to determine what parameters the swift-format plugin was calling the executable – subscribing to `NSWorkspace.willLaunchApplicationNotification`, looping `NSWorkspace.shared.runningApplications`, and an attempt to implement `ps -ax -o command | grep swift-format`. I couldn't find any way to determine sub-processes of the Process the test was creating.

So, this does fix #483, but I'm open to changes to the tests if there's a more conventional way to write them.


allevato:
I'm a bit nervous about the complexity of that test. Setting up a new SPM workspace, running SPM, and then checking output feels like it's a recipe for flakiness or potential problems if SPM upstream ever changes their behavior.

I'm good with the change to the plugin itself, but can you drop the test? I'd like to think more holistically about how to get better coverage there and in other parts of the codebase that have historically been lacking it.

BrianHenryIE:
I've deleted the test. I think half the value in tests is that they make you proof-read your code again and again, so we did get at least that benefit from it.

SPM's [PluginTests::testCommandPluginInvocation()](https://github.com/apple/swift-package-manager/blob/a2bdb7207098189deb9141b09fbf7765bae37c75/Tests/FunctionalTests/PluginTests.swift#L171) does exactly that – creates a package in a temp dir from a string and executes the plugin. It also has a [Fixtures/Plugins](https://github.com/apple/swift-package-manager/tree/a2bdb7207098189deb9141b09fbf7765bae37c75/Fixtures/Miscellaneous/Plugins) directory which are the projects used in other tests in that class.

It's definitely nicer than what I wrote!

```swift
// Invoke the command plugin that prints out various things it was given, and check them.
testCommand(package: package, plugin: "PluginPrintingInfo", targets: ["MyLibrary"], arguments: ["veni", "vidi", "vici"]) { output in
  output.check(diagnostic: .equal("Root package is MyPackage."), severity: .info)
  output.check(diagnostic: .and(.prefix("Found the swiftc tool"), .suffix(".")), severity: .info)
}
```

[Sources/SPMTestSupport/misc.swift](https://github.com/apple/swift-package-manager/blob/c86a7dab09f6e4fb6ebdd74e099e130982dd2dba/Sources/SPMTestSupport/misc.swift) is the set of functions for working with the fixtures. The SPMTestSupport library `.target` is not published as a product in [SPM's Package.swift](https://github.com/apple/swift-package-manager/blob/main/Package.swift) and AIUI, that's preventing me from setting it as a dependency for the `.testTarget` I had added here.

Maybe all it would take to use it here would be to add this to SPM's Package.swift:
```swift
.library(
  name: "SPMTestSupport",
  targets: ["SPMTestSupport"]
),
```

I'll take another look at using it – I'll try symlinking it or checking out and editing SPM – and if it works well for this test case I'll open an issue in the SPM repo to start a discussion.

allevato:
> [Sources/SPMTestSupport/misc.swift](https://github.com/apple/swift-package-manager/blob/c86a7dab09f6e4fb6ebdd74e099e130982dd2dba/Sources/SPMTestSupport/misc.swift) is the set of functions for working with the fixtures. The SPMTestSupport library .target is not published as a product in [SPM's Package.swift](https://github.com/apple/swift-package-manager/blob/main/Package.swift) and AIUI, that's preventing me from setting it as a dependency for the .testTarget I had added here.
>
> Maybe all it would take to use it here would be to add this to SPM's Package.swift:

The problem with this approach is that depending on that would make all of the `swift-package-manager` repository a dependency of `swift-format`, which we wouldn't want.

I suppose that could be partially solved by making a completely separate test package in a subdirectory, but then you have to remember to run those tests separately since `swift test` in the repo root wouldn't do it.

This feels like an area that SwiftPM needs to come up with a good solution for. While it would be nice to have total coverage here, I also don't feel that we should go to heroics if it's not easy to do or would be potentially fragile. (It's easier for SPM to do those full integration tests by constructing packages in a temp dir because if they change something under the hood, that project's own tests would catch it and be modified to support it. It's less easy for a downstream project.)

BrianHenryIE:
> This feels like an area that SwiftPM needs to come up with a good solution for. 

Agreed. I hope the SPM maintainers already have it on their mind to spin `SPMTestSupport` into its own package. Rewriting the test at this stage is almost more for the benefit of the `swift-package-manager` repository than `swift-format`!

allevato (approved):
Thanks for the fix! Merging now.

Issue #611, opened 2023-09-04
Unit tests for swift-format CLI tool and for Plugins


natikgadzhi:
Currently, `swift-format` test suite covers Rules and PrettyPrint. Several components are not covered:
- `generate-pipeline` that builds `RuleRegistry` and friends
- `swift-format` the CLI executable target, including `Configuration`, `Frontend`, and the whole argument-parser based thing.
- `Plugins`. 

**Disclaimer:** my understanding and knowledge are _very_ high-level and basic still, I'm only diving in.

I _think_ with some guidance, we could split these components up into smaller chunks, and each of those chunk-unit-test-issues could be a great good-first-issue for beginners.

BrianHenryIE:
I introduced a test for the plugin in #608: [PluginRunTests.swift](https://github.com/apple/swift-format/pull/608/files#diff-6a73569e872fb33eff5ee337553dbf0c97b484dd640fb809967178166d8839fb). I expect there is some Gherkin style BDD library that could be used to write the test in a nicer way, but I've never used one in Swift and I don't know the policy on introducing third party libraries to the project (I notice all dependencies currently are Apple projects).

ahoppen:
Tracked in Apple’s issue tracker as rdar://126948360

Issue #610, opened 2023-09-04
Should we lint and format code of swift-format with swift-format?


natikgadzhi:
I was working on a prototype for #607 and noticed a few lines with trailing spaces here and there, and thought I'd check for formatting rules. Apologies if I missed this in contributing docs, but it seems that: 
- The repository is not using any formatting tool explicitly
- No `.swift-format` configuration in particular
- No formatting preference documented in [CONTRIBUTING.md](CONTRIBUTING.md) or [Documentation/Development.md](Documentation/Development.md)
- There is `Scripts/format-diff.sh`[Scripts/format-diff.sh], though, but I don't know if it's hooked up in a CI (and no GitHub CI checks are set up, seemingly?)

## How are we doing now?

I tried running the default configuration, and `swift-format` is a bit angry with itself: 

<summary>
There are 266 offenses: 

<details>

```bash
$ swift-format lint --recursive Sources
Sources/SwiftFormatConfiguration/Configuration.swift:222:1: warning: [Indentation] indent by 2 spaces
Sources/SwiftFormatConfiguration/Configuration.swift:237:1: warning: [Indentation] indent by 2 spaces
Sources/SwiftFormatConfiguration/Configuration.swift:261:1: warning: [LineLength] line is too long
Sources/SwiftFormatConfiguration/Configuration.swift:263:1: warning: [LineLength] line is too long
Sources/SwiftFormatConfiguration/Configuration.swift:264:1: warning: [LineLength] line is too long
Sources/SwiftFormatConfiguration/Configuration.swift:265:1: warning: [LineLength] line is too long
Sources/swift-format/Frontend/LintFrontend.swift:36:29: warning: [AddLines] add 1 line break
Sources/swift-format/Frontend/LintFrontend.swift:37:1: warning: [Indentation] unindent by 2 spaces
Sources/swift-format/Frontend/LintFrontend.swift:38:1: warning: [Indentation] unindent by 2 spaces
Sources/swift-format/Frontend/LintFrontend.swift:39:1: warning: [Indentation] unindent by 2 spaces
Sources/swift-format/Frontend/LintFrontend.swift:40:1: warning: [Indentation] unindent by 2 spaces
Sources/swift-format/Frontend/LintFrontend.swift:41:1: warning: [Indentation] unindent by 2 spaces
Sources/swift-format/Frontend/Frontend.swift:17:1: warning: [OrderedImports] sort import statements lexicographically
[...]

PR #612, opened 2023-09-04
Output a warning for unknown configuration rules in `.swift-format`

### LINKED ISSUE #607: Implement a warning or error if an unrecognized Rule is specified in .swift-format
Right now, if one made a typo in a rule name in `.swift-format` and then ran the formatter, it will ignore the typo and, obviously, not apply the rule.

I was working on https://github.com/apple/swift-syntax/pull/2145, and of course, that's precisely what I did. _Oh, nice, our imports are already ordered, right?_

## Proposed solution
Since we know the list of all available rules, we could verify that each string in the Rules block matches an existing rule.

```swift
// Config.swift
//
// If the `rules` key is not present at all, default it to the built-in set
// so that the behavior is the same as if the configuration had been
// default-initialized. To get an empty rules dictionary, one can explicitly
// set the `rules` key to `{}`.
self.rules =
  try container.decodeIfPresent([String: Bool].self, forKey: .rules)
  ?? defaults.rules
```

It should be relatively easy, we could just generate an `enum` with all rule names, and use that as the key when decoding the rules.

Should I draft this in a PR?

natikgadzhi:
**Summary**:
- Helps developers surface any errors and typos in their `.swift-format` config files by showing a warning when running `swift-format` if the configuration includes invalid rules.
- The actual check is in `Frontend.swift`.
- The dictionary of rules for verification is in `RuleRegistry`, which is made `public` to be able to import it in `swift-format` target.

Example output: 

```bash
 ❯ ../swift-format/.build/debug/swift-format lint Sources/SwiftSyntax/Convenience.swift
/Users/nategadzhi/src/apple/swift-syntax/.swift-format:10:10: warning: Configuration contains an unrecognized rule IAmABrokenRule
/Users/nategadzhi/src/apple/swift-syntax/.swift-format:14:10: warning: Configuration contains an unrecognized rule LolNope
```

**Motivation**:
- Closes #607

**Reviewing**
- I'm a beginner — pretty please "well actually" me if you see me making mistakes, I'm here to learn ❤️
- ~~I've made `RuleRegistry.rules` public. We could in theory use `_@spi` instead. `public` seems simple and okay — do you think that's fine?~~ `RuleRegistry` is now `@_spi(Internal)`.
- **There are no tests whatsoever**, as in the whole `swift-format` executable target is not tested. See #611. I think I can start us off with a very simple test that tests specifically `Frontend.configuration()`.

**Potential improvements**
- ✅  ~~This PR does not report the source location, hence the `<unknown>` in the warning line. Hypothetically, we could read the configuration into an array of strings, and find the index of the string containing the unsupported rule, and construct the source location from that. Since the performance hit is negligible, that could be a nice touch.~~ Source location reporting is in, but it's sliiiightly ugly.

@allevato, how does that look to you? What else can I improve? If the general direction seems good, I would be happy to invest a bit more time and perhaps provide source locations.


Sources/SwiftFormatConfiguration/RuleRegistry+Generated.swift
```diff
@@ -12,8 +12,8 @@
 
 // This file is automatically generated with generate-pipeline. Do Not Edit!
 
-enum RuleRegistry {
-  static let rules: [String: Bool] = [
+public enum RuleRegistry {
```
allevato, line 15: We should use an `@_spi` here; we don't want this table to be public API to anyone who uses `SwiftFormat`.

`@_spi(Internal)` would be fine, I think.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -208,4 +211,16 @@ class Frontend {
     // default configuration.
     return Configuration()
   }
+
+  /// Checks if all the rules in the given configuration are supported by the registry.
+  /// If there are any rules that are not supported, they are emitted as a warning.
+  private func verifyConfigurationRules(for configuration: Configuration) -> Void {
```
allevato, line 217:
```suggestion
  private func verifyConfigurationRules(for configuration: Configuration) {
```

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -208,4 +211,16 @@ class Frontend {
...
+    // If any rules in the decoded configuration are not supported by the registry,
+    // emit them into the diagnosticsEngine as warnings.
+    // That way they will be printed out, but we'll continue execution on the valid rules.
+    let invalidRules = configuration.rules.filter { !RuleRegistry.rules.keys.contains($0.key) }
+    if !invalidRules.isEmpty {
+      diagnosticsEngine.emitWarning("Configuration contains an unsupported rule: \(invalidRules.keys.joined(separator: ", "))", location: nil)
```
allevato, line 223:
```suggestion
      diagnosticsEngine.emitWarning("Configuration contains an unrecognized rule: \(invalidRules.keys.joined(separator: ", "))", location: nil)
```

allevato: "Unrecognized" feels clearer than "Unsupported". In my mind the latter could be interpreted as "this rule exists, but it's not supported for some reason".

allevato (commented):
Can you update this now that #614 has been merged, which moved everything from the `SwiftFormatConfiguration` module into `SwiftFormat`?

Sources/SwiftFormatConfiguration/RuleRegistry+Generated.swift
natikgadzhi: Thank you for the review!

Sure, this seems reasonable — let me clean this up a bit later today.

I want to also add the pluralization of the warning message, and see if I can get source location to work _easily_. I think that `<unknown>` in CLI output (when it's the only line) does not make it clear that it's supposed to be the source location, and it's just empty.

natikgadzhi:
@allevato, I've addressed your first round of feedback:
- `@_spi` for `RuleRegistry`
- Wording

Next up: potentially do source location. If you're happy with what is currently here, it would be okay by me to merge and follow up with an improvement. But, up to you!

Sources/SwiftFormat/Core/RuleRegistry+Generated.swift
```diff
@@ -12,8 +12,8 @@
 
 // This file is automatically generated with generate-pipeline. Do Not Edit!
 
-enum RuleRegistry {
-  static let rules: [String: Bool] = [
+@_spi(Internal) public enum RuleRegistry {
```
natikgadzhi, line 15: @allevato, `@_spi` is done here.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -11,10 +11,11 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
-import SwiftFormat
 import SwiftSyntax
 import SwiftParser
 
+@_spi(Internal) import SwiftFormat
```
natikgadzhi, line 17: Reordered imports, but we should probably just setup formatter in #610. I'm happy to do it, just say the word ;)

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -176,35 +177,80 @@ class Frontend {
     at configurationFileURL: URL?,
     orInferredFromSwiftFileAt swiftFileURL: URL?
   ) -> Configuration? {
-    // If an explicit configuration file path was given, try to load it and fail if it cannot be
-    // loaded. (Do not try to fall back to a path inferred from the source file path.)
+    // The actual config file that this frontend is going to use.
+    let configURL: URL?
```
natikgadzhi, line 181: I need the config file URL later to figure out the source location, so this function got a little refactor in.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -176,35 +177,80 @@ class Frontend {
...
+    }
+  }
+
+  /// Checks if all the rules in the given configuration are supported by the registry.
+  /// If there are any rules that are not recognized by this version of swift-format, they are emitted as a warning.
+  ///
+  /// - parameters:
+  ///   - configuration: `Configuration` to verify
+  ///   - configURL: Configuration file `URL` to use in `Diagnostic.Location`
+  private func verifyConfigurationRules(for configuration: Configuration, at configURL: URL) throws {
```
natikgadzhi, line 221: I'm _meh_ on naming here. It does get the job done, but I don't think the name is 💯. 

`verify` implies that if the config is invalid, it will throw and the program will exit with an error.

Perhaps an explicit `checkForUnrecognizedRules(in: config, at: path)`?

Also: I don't think it `throws` anymore.

Sources/swift-format/Utilities/Diagnostic.swift
```diff
@@ -47,6 +47,12 @@ struct Diagnostic {
       self.line = findingLocation.line
       self.column = findingLocation.column
     }
+
+    init(file: String, line: Int, column: Int) {
```
natikgadzhi, line 51: It seems like before we only made `Diagnostic.Location` out of `SourceLocation`, but I did not want to import `SwiftSyntax` into `swift-format` target. Cool to use it directly?

Sources/swift-format/Utilities/DiagnosticsEngine.swift
```diff
@@ -65,6 +65,20 @@ final class DiagnosticsEngine {
         message: message))
   }
 
+  /// Emits a generic warning message.
+  ///
+  /// - Parameters:
+  ///   - message: The message associated with the error.
+  ///   - location: The `Diagnostic.Location` in the source code associated with the error, or nil if there is no
+  ///     location associated with the error.
+  func emitWarning(_ message: String, location: Diagnostic.Location? = nil) {
```
natikgadzhi, line 74: I'm not _entirely_ happy with this — this function is very similar to `emitError`, but it takes `Diagnostic.Location?` instead of `SourceLocation`. Nothing wrong with that, but the API now is slightly inconsistent. 

I don't know if it's worth providing the other two overloads for consistency, or if that's fine.

natikgadzhi (commented):
@allevato ready for another review! I made the source locations work in a not entirely ugly way, I think!

Sources/swift-format/Frontend/Frontend.swift
allevato: `checkForUnrecognizedRules(in: config, at: path)` sounds good to me. And yeah, remove the `throws` if it's no longer necessary.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -176,35 +177,80 @@ class Frontend {
...
+      diagnosticsEngine.emitWarning(
+        "Configuration contains an unrecognized rule \(rule.key)",
+        location: location
+      )
+    }
+  }
+
+  /// Returns a `Diagnostic.Location` of the substring in an array of Strings.
+  /// It's intented to find the location of a configuration Rule in an array of Strings representing the encoded JSON swift-format configuration file.
+  private func diagnosticLocationFor(substring: String, in configuration: String, filePath: String) -> Diagnostic.Location? {
```
allevato, line 243: For what it's worth, I don't think this is really necessary. If someone mistypes a rule name, it won't take them long to figure out where in the configuration it is. And if we could get the information during decoding, it _might_ be useful (still questionable though), but since this is just doing raw string searches, it's too prone to false positives if the invalid rule name matches something somewhere else in the configuration string.

Can you remove this and just emit a warning with no location?

natikgadzhi:
> Can you remove this and just emit a warning with no location?

Sounds great to me. I felt that it's too much ceremony, but liked the neat detail of having a line number.

I'll do that, and some light cleaning up — that should be ready in a couple of hours.

natikgadzhi:
@allevato, alright, cleaned it up. Pained me to reduce clarity by removing source location, but you're right — re-reading config and doing raw string match is clumsy and buggy.

allevato (approved):
Looks good now, thank you!

Issue #606, opened 2023-09-02
Improve `dump-configuration` with inline comments to explain each rule


natikgadzhi:
Right now, `swift-format dump-configuration` is a great (and automatically up to date!) source of documentation on what rules are available:

```bash
swift-syntax internal/swift-format/omit-returns ≡2s
 ❯ swift-format dump-configuration
{
  "fileScopedDeclarationPrivacy" : {
    "accessLevel" : "private"
  },
  "indentation" : {
    "spaces" : 2
  },
  "indentConditionalCompilationBlocks" : true,
  "indentSwitchCaseLabels" : false,
  "lineBreakAroundMultilineExpressionChainComponents" : false,
  "lineBreakBeforeControlFlowKeywords" : false,
  "lineBreakBeforeEachArgument" : false,
  "lineBreakBeforeEachGenericRequirement" : false,
  "lineLength" : 100,
  "maximumBlankLines" : 1,
  "noAssignmentInExpressions" : {
    "allowedFunctions" : [
      "XCTAssertNoThrow"
    ]
  },
  "prioritizeKeepingFunctionOutputTogether" : false,
  "respectsExistingLineBreaks" : true,
  "rules" : {
    "AllPublicDeclarationsHaveDocumentation" : false,
    "AlwaysUseLowerCamelCase" : true,
    "AmbiguousTrailingClosureOverload" : true,
    "BeginDocumentationCommentWithOneLineSummary" : false,
    "DoNotUseSemicolons" : true,
    "DontRepeatTypeInStaticProperties" : true,
    "FileScopedDeclarationPrivacy" : true,
    "FullyIndirectEnum" : true,
    "GroupNumericLiterals" : true,
    "IdentifiersMustBeASCII" : true,
    "NeverForceUnwrap" : false,
    "NeverUseForceTry" : false,
    "NeverUseImplicitlyUnwrappedOptionals" : false,
    "NoAccessLevelOnExtensionDeclaration" : true,
    "NoAssignmentInExpressions" : true,
    "NoBlockComments" : true,
    "NoCasesWithOnlyFallthrough" : true,
    "NoEmptyTrailingClosureParentheses" : true,
    "NoLabelsInCasePatterns" : true,
    "NoLeadingUnderscores" : false,
    "NoParensAroundConditions" : true,
    "NoVoidReturnOnFunctionSignature" : true,
    "OmitExplicitReturns" : false,
    "OneCasePerLine" : true,
[...]

PR #615, opened 2023-09-05
Generate rule docs automatically

natikgadzhi:
## Summary 
- First step of #606, adds documentation for the list of available rules in `Documentation` directory.

## Changes
- Adds a `generate-pipeline` generator `RuleDocumentationGenerator` that extracts each rule DocC comment, and writes it as a table into `Documentation/RuleDocumentation.md`.
- In `Documentation/Configuration.md`, I've added a paragraph on rules configuration that links to the new auto-generated docs file, and also tells users that they can run `swift-format dump-configuration` to check out the list of available rules.

## How to review this
- Check out the format of the markdown file in `RuleDocumentation.md`.
- Navigate to Documentation files in the PR repo explorer. Does the markup and content look good?
- Does the layout make sense? I started with a table layout, but it wasn't as readable as I would hope (no way to tell lines to vertically align to top, no way to specify column width on GitHub markdown viewer — let's leave this until we can move to DocC?)

## TODOs

- [x] Rule description markup breaks where the original DocC content had a line break just because of long lines.


Documentation/Configuration.md
```diff
@@ -103,6 +103,18 @@ An example `.swift-format` configuration file is shown below.
 }
 ```
 
+## Linter and Formatter Rules Configuration
+
+In the `rules` block of `.swift-format`, you can specify which rules to apply
+when linting and formatting your project. Read the
+[rules documentation](Documentation/RuleDocumentation.md) to see the list of all
```
natikgadzhi, line 110: Ooops, the links are broken.
```suggestion
[rules documentation](RuleDocumentation.md) to see the list of all
```

Documentation/RuleDocumentation.md
```diff
@@ -0,0 +1,478 @@
+<!-- This file is autogenerated with `generate-pipeline` utility of
+`swift-format`. DO NOT EDIT MANUALLY! -->
+
+# `swift-format` Lint and Format Rules
+
+Use the rules below in the `rules` block of your `.swift-format`
+configuration file, as described in
+[Configuration](Documentation/Configuration.md). All of these rules can be
```
natikgadzhi, line 8:
```suggestion
[Configuration](Configuration.md). All of these rules can be
```

Sources/generate-pipeline/RuleCollector.swift
```diff
@@ -20,6 +20,11 @@ import SwiftParser
 final class RuleCollector {
   /// Information about a detected rule.
   struct DetectedRule: Hashable {
+
+    /// The DocC comments of the rule,
+    /// presumably in the leading trivia of the rule declaration.
```
natikgadzhi, line 25:
```suggestion
    /// extracted from the .leadingTribia of the rule class or struct.
```

Sources/generate-pipeline/RuleDocumentationGenerator.swift
```diff
@@ -0,0 +1,83 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
```
natikgadzhi, line 5: Someone told me to put just 2023 if the file is new.

Sources/generate-pipeline/RuleDocumentationGenerator.swift
```diff
@@ -0,0 +1,83 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
natikgadzhi, line 13: Honest question: why do we need `Foundation` in our generators?

natikgadzhi:
Rebased over #614 merge. 

@allevato, how does that look to you?

Documentation/RuleDocumentation.md
```diff
@@ -0,0 +1,478 @@
...
+applied in the linter, but only some of them can format your source code
+automatically.
+
```
allevato, line 11: Could you generate an unordered list here that would act as a table of contents, linking to each of the headers below?

Sources/generate-pipeline/RuleDocumentationGenerator.swift
allevato: `write(into:)` takes a `FileHandle` argument.

Sources/generate-pipeline/RuleDocumentationGenerator.swift
```diff
@@ -0,0 +1,83 @@
...
+
+  private func ruleFormatSupportDescription(for rule: RuleCollector.DetectedRule) -> String {
+    return rule.canFormat ?
+      "`\(rule.typeName)` rule can format your code automatically." :
+      "`\(rule.typeName)` is a linter-only rule."
+  }
+
+  /// Takes the DocC comment of the rule and strip `///` from the beginning of each line.
+  /// Also removes empty lines with under 4 characters.
+  private func ruleDescription(for rule: RuleCollector.DetectedRule) -> String {
```
allevato, line 65: We already have more general code for parsing doc comments in `DocumentationCommentText`. You could use that here instead (it's under `@_spi`, but that's fine).

`generate-pipeline` already imports `SwiftFormat` because it actually pokes at some of the runtime properties of the rules. It's kind of weird, but it means you don't have to worry about introducing that dependency.

allevato (commented):
Thanks, this looks really good! A few suggestions but after that I think this is good to go and the iterate on afterwards.

Sources/generate-pipeline/RuleCollector.swift
```diff
@@ -23,6 +23,10 @@ final class RuleCollector {
     /// The type name of the rule.
     let typeName: String
 
+    /// The description of the rule, extracted from the rule class or struct DocC comment
+    /// with `DocumentationCommentText(extractedFrom:)`
+    let description: String?
```
natikgadzhi, line 28: Optional because `DocumentCommentText()` may return nil, but that's fine for us.

Sources/generate-pipeline/RuleDocumentationGenerator.swift
natikgadzhi: So technically we could just write into a file differently, but we just don't aim to support environments where Foundation is not available yet?

natikgadzhi (commented):
Ready for another pass, @allevato!

Sources/generate-pipeline/RuleDocumentationGenerator.swift
allevato: I can't think of any platforms that don't have Foundation that I'd want to run generate-pipeline on.

The rest of swift-format also uses Foundation, so I don't know why I'd want to single out generate-pipeline specifically.

allevato (approved):
Thank you for this! I'm really happy with how it turned out and it'll be a big help keeping the rule documentation updated going forward.

If you're looking for something else to do 😁 , it would be nice to get coverage for the rest of the fields in `Configuration` and the types associated with them. That's a bit more work though, because we're not already collecting the decls today. But basically we'd just want to scan Configuration.swift for all the fields that are `public var`s, and then do the same for the other types defined in that file when we have config fields that are themselves complex values.

natikgadzhi:
@allevato, thank you for reviewing this and guiding me through it! ❤️ 

I'm happy to help more. I was curious about trying to convert the Documentation directory to a DocC site first, and maybe even recording that working session / streaming it to get more people comfortable with DocC for their projects. That's likely some time next week. 

Do you think it would be okay to convert to DocC first, and then add more documentation coverage in Configuration, as well as in code otherwise?

allevato:
> I'm happy to help more. I was curious about trying to convert the Documentation directory to a DocC site first, and maybe even recording that working session / streaming it to get more people comfortable with DocC for their projects. That's likely some time next week.
> 
> Do you think it would be okay to convert to DocC first, and then add more documentation coverage in Configuration, as well as in code otherwise?

I think that all sounds fantastic! The public API for swift-format (everything in the API subfolder) is pretty small so getting DocC support up and running should hopefully be fairly straightforward. And I love the idea of recording your work session for the community.

If we can only get `Configuration` documented in API docs, then that's still a great start, and maybe we could link to that in the user docs. But it would be nice to figure out a way to surface it in the user docs directly too, because I'm thinking it would look better to have the "this is how you configure swift-format" docs look more like the JSON representation than the underlying API representation.

I wonder if, instead of scanning the source, we could consume the symbol graphs that DocC uses to generate that alternate form instead...

I haven't used DocC much myself, so I'm interested to see what you come up with!

PR #619, opened 2023-09-07
Add option to disable trailing commas on multi-line collections

mlavergn:
Extends the configuration options to allow disabling the addtion of trailing commas on multi-line collections.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -42,7 +42,7 @@ public struct Configuration: Codable, Equatable {
     case rules
     case spacesAroundRangeFormationOperators
     case noAssignmentInExpressions
-    case multilineCollectionTrailingCommas
+    case multiElementCollectionTrailingCommas
   }
 
   /// A dictionary containing the default enabled/disabled states of rules, keyed by the rules'
@@ -163,8 +163,8 @@ public struct Configuration: Codable, Equatable {
   /// Contains exceptions for the `NoAssignmentInExpressions` rule.
   public var noAssignmentInExpressions: NoAssignmentInExpressionsConfiguration
 
-  /// Determines whether multi-line list initializers should have trailing commas.
-  public var multilineCollectionTrailingCommas: Bool
+  /// Determines whether multi-element collection literals should have trailing commas.
+  public var multiElementCollectionTrailingCommas: Bool
 
   /// Constructs a Configuration by loading it from a configuration file.
   public init(contentsOf url: URL) throws {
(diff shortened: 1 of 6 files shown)


Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -162,6 +163,9 @@ public struct Configuration: Codable, Equatable {
   /// Contains exceptions for the `NoAssignmentInExpressions` rule.
   public var noAssignmentInExpressions: NoAssignmentInExpressionsConfiguration
 
+  /// Determines whether multi-line list initializers should have trailing commas
```
allevato, line 166: Can you end this sentence with a period?

Tests/SwiftFormatTests/PrettyPrint/CommaTests.swift
```diff
@@ -0,0 +1,145 @@
+import SwiftFormat
+
+final class CommaTests: PrettyPrintTestCase {
```
allevato, line 3: Can you add some tests that cover dictionary literals as well?

mlavergn: Note that tests for single-line dictionary setters wrap differently than single-line array setters, that's reflected in the expected test output.

mlavergn:
I gleaned the name `multilineCollectionTrailingCommas` from an older PR, but since this does affect single line dictionaries, I wonder if `collectionTrailingCommas` would be more appropriate.

allevato:
> I gleaned the name `multilineCollectionTrailingCommas` from an older PR, but since this does affect single line dictionaries, I wonder if `collectionTrailingCommas` would be more appropriate.

Yeah, "line" is the wrong word here, because the logic isn't based on whether the collection wraps across multiple lines, it's about whether it has multiple elements—single element arrays/dictionaries never have a trailing comma inserted, even if they wrap (because of Swift parser subtleties around how collection types are parsed in expression contexts). I think there are other places in swift-format where we use "line" when we really mean "statement" or "element", but fixing those up require ensuring compatibility with older configs.

How do you feel about `multiElementCollectionTrailingCommas`?

mlavergn:
> > I gleaned the name `multilineCollectionTrailingCommas` from an older PR, but since this does affect single line dictionaries, I wonder if `collectionTrailingCommas` would be more appropriate.
> 
> Yeah, "line" is the wrong word here, because the logic isn't based on whether the collection wraps across multiple lines, it's about whether it has multiple elements—single element arrays/dictionaries never have a trailing comma inserted, even if they wrap (because of Swift parser subtleties around how collection types are parsed in expression contexts). I think there are other places in swift-format where we use "line" when we really mean "statement" or "element", but fixing those up require ensuring compatibility with older configs.
> 
> How do you feel about `multiElementCollectionTrailingCommas`?

It's a mouthful, but it certainly self-documents what the setting is for. I'll refactor it.

mlavergn:
Refactored the setting name to `multiElementCollectionTrailingCommas` and use `literal` rather than `initializer` in the docs.

mlavergn:
Added a DocC header to auto-gen proper documentation based on the styling of `indentSwitchCaseLabels`

allevato (approved):
This looks great. Thanks for your contribution!

PR #625, opened 2023-09-14
Add option to print number of instructions executed by `swift-format`

ahoppen:
I’m planning to use this option to measure the performance of `swift-format` periodically on a fixed set of project to detect any performance regressions.

I made the `--measure-instructions` option hidden so that it only shows up when running `swift-format lint --help-hidden`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -48,6 +48,10 @@ let package = Package(
   ],
   targets: [
     .target(
+      name: "_InstructionCounter"
+    ),
+
+    .target(
       name: "SwiftFormat",
       dependencies: [
         .product(name: "Markdown", package: "swift-markdown"),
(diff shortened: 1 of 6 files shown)


allevato (approved):
Nice! Is there a way we can have these numbers reported in CI, say, once we set up swift-format to verify its own formatting?

I guess what I'd want is to have swift-ci comment back on the PR any performance changes, similar to the toolchain's benchmark runs, but I don't know how difficult that would be to set up (especially since we still don't have triggers on this repo itself).

ahoppen:
Yes, that would be nice but for that we would also need to have a stable project to format as a baseline. My plan for now is to run these tests locally. I’ll let you know if I discover any performance regressions. Once I have finished the performance testing script, I can also share it with you so you can run it locally before merging PRs that might have performance impact.

PR #627, opened 2023-09-15
Add the `UseExplicitNilCheckInConditions` rule.

allevato:
This rule diagnoses and replaces occurrences of patterns like `if let _ = x` with `if x != nil` (likewise for conditions in `guard` and `while` statements). The rationale for this is that explicitly testing for `nil` is clearer about intent than binding and immediately discarding the wrapped value.

### DIFF
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -12,6 +12,7 @@ automatically.
 Here's the list of available rules:
 
 - [AllPublicDeclarationsHaveDocumentation](#AllPublicDeclarationsHaveDocumentation)
+- [AlwaysUseLiteralForEmptyCollectionInit](#AlwaysUseLiteralForEmptyCollectionInit)
 - [AlwaysUseLowerCamelCase](#AlwaysUseLowerCamelCase)
 - [AmbiguousTrailingClosureOverload](#AmbiguousTrailingClosureOverload)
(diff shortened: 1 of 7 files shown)


allevato:
cc @dylansturg

Documentation/RuleDocumentation.md
```diff
@@ -59,6 +61,17 @@ Lint: If a public declaration is missing a documentation comment, a lint error i
...
 `AllPublicDeclarationsHaveDocumentation` is a linter-only rule.
 
+### AlwaysUseLiteralForEmptyCollectionInit
+
+Never use `[<Type>]()` syntax. In call sites that should be replaced with `[]`,
+for initializations use explicit type combined with empty array literal `let _: [<Type>] = []`
+Static properties of a type that return that type should not include a reference to their type.
+
+Lint:  Non-literal empty array initialization will yield a lint error.
+Format: All invalid use sites would be related with empty literal (with or without explicit type annotation).
```
dylansturg, line 71: "would be _replaced_ with empty literal..."?

dylansturg (approved):

Documentation/RuleDocumentation.md
allevato: Ah, that's a typo from earlier, there was a race between the generate-the-docs PR and the one adding this rule so it didn't get generated the first time. I'll update that in a separate follow-up because I need to do a pass over all the rule docs anyway to make sure they're formatted nicely in the new Markdown file.

PR #629, opened 2023-09-15
Revert "Cherrypick everything since release/5.9 was cut."

glessard:
Reverts apple/swift-format#621

in order to unblock PR testing


allevato:
@glessard Apologies for the breakage! I overlooked that CI would still be using the local checkouts even when I reference the `509.0.0` SPM dependency, and we don't have full CI support on this repo yet. 😞

PR #632, opened 2023-09-18
Rename `_InstructionCounter`

bnbarham:
In the unified build there ends up being two of these, one in swift-syntax and one in swift-format. Rename so that the target is unique.


glessard (approved):
I was going to suggest exactly this name.

bnbarham:
I'm not really sure why cross PR testing didn't catch this 🤔

allevato:
Doesn't the `import` statement here need to be updated as well with the new name? https://github.com/apple/swift-format/blob/8488a21aa3051b4d37d319fe21a7cd2c733f6bc3/Sources/swift-format/Subcommands/PerformanceMeasurement.swift#L14

bnbarham:
It does :(

PR #633, opened 2023-09-19
Fix up import of renamed `_InstructionCount`

bnbarham:


glessard (approved):
Thanks @allevato!

allevato (approved):
I haven't tested it but since main is already broken, YOLO

bnbarham:
I just did a quick build 😅 Should be good this time.

PR #639, opened 2023-09-23
Add installation instructions for Homebrew

beatrupp:
Add a section to the `README.md` to install swift-format via Homebrew


allevato (approved):
This seems fine; the community appears to do a pretty good job of keeping the Homebrew spec updated (since we don't maintain it ourselves). Thanks!

PR #640, opened 2023-09-27
change to non optional utf8 function

zunda-pixel:
String.data(using)? -> Data(String.utf8)
String(data:encoding)? -> String(decoding:as)

### DIFF
--- a/Sources/generate-swift-format/FileGenerator.swift
+++ b/Sources/generate-swift-format/FileGenerator.swift
@@ -38,7 +38,7 @@ extension FileGenerator {
 extension FileHandle: TextOutputStream {
   /// Writes the provided string as data to a file output stream.
   public func write(_ string: String) {
-    guard let data = string.data(using: .utf8) else { return }
+    let data = Data(string.utf8)
     write(data)
   }
 }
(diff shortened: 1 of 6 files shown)


allevato:
Could you measure the performance of these changes? I have a suspicion that the `Data` initializer taking a `Sequence<UInt8>` might be slower than the `String` method from Foundation (though I'd be happy to be wrong).

You should be able to use the recently added `--measure-instructions` flag before and after your change to get rough numbers to start.

zunda-pixel:
I don't known how to use `--measure-instructions`.
is this `swift test`?

zunda-pixel:
I found source code. but I don't known this code used in current Swift.
https://github.com/apple/swift-corelibs-foundation/blob/8a9b69b5041b5069360239cd18304fd8e2eb91d9/Sources/Foundation/NSStringAPI.swift#L901-L921

allevato:
> I don't known how to use `--measure-instructions`. is this `swift test`?

It's a swift-format flag. Try running the following commands with and without your change and report the number of instructions in each case:

```
swift-format lint --measure-instructions Package.swift
swift-format format --measure-instructions Package.swift
```

zunda-pixel:
number of instructions is unstable...

| Name | Lint | Format |
| :- | :- | :-|
| Data(String.utf8) | 83694608 | 103411109 |
| String.data(using: .utf8) | 89545458 | 130931785 |

PR #641, opened 2023-09-28
Fix a bug where an unfolded `SequenceExpr` would make it to the pretty-printer.

allevato:
Since the `UseExplicitNilCheckInConditions` rule was using string interpolation based parsing, it returned a `SequenceExpr` instead of an `InfixOperatorExpr`. This unfolded expression then made it to the pretty printer, which it didn't expect (because the input is folded before being processed by the rules, but not after that).

In this case, creating the new expression directly as nodes is barely more work than using the string-based parsing and then folding it, so I just did that.

Tests didn't catch this because the sequence expr is valid Swift code when stringified (for comparison in unit tests). To address this, we now do a pretty-printer pass on all the format rule outputs. We don't assert based on the pretty-printed output (that's not really a "unit" test anymore), but it at least causes a test crash if the output wasn't suitable for pretty printing.

Doing so caught a similar latent bug in `UseShorthandTypeNames`.


allevato:
cc @thomasvl

PR #642, opened 2023-09-28
Ignore too long end of line comments when they're wrapped in `printCo…

dylansturg:
…ntrol` tokens that disable breaking.

The rationale here is that if the comment exists in a section of code where breaking is purposefully disabled, then there's probably a reason that the comment cannot be moved to another line. The `printControl` tokens that disable breaking are the signal to the pretty print algorithm for which sections are those special "no breaks allowed here", and it can just respect those when diagnosing long lines.


allevato (approved):
Thanks!

PR #644, opened 2023-09-29
Ignore symlinks and hidden (dot) files during `--recursive`.

### LINKED ISSUE #643: Is there a configuration to exclude folders from formatting/linting?
This might be staring me in the face, but when I run `swift-format . -i --recursive` that includes all the `.build` folders that are created in my process (including checkouts and whatever else). 

So, what should be an instantaneous format turns into a minute or so.

I was hoping there was a config key or command line argument that would allow exclusions to certain folders.

allevato:
Note that hidden files will *always* be honored if they are specified explicitly among the paths on the command line. Symlinks encountered during recursive traversal *or* passed on the command line will only be followed if the flag `--follow-symlinks` is passed. This is meant to avoid symlinks being hidden by shell expansions; for example, running `swift-format *` where the current directory contains a symlink and following that symlink might be surprising to users.

This PR also adds a new test target for code in the `swift-format` binary in order to test `FileIterator` without factoring the binary's code out to a separate library target. We should use this in the future to expand coverage of other functionality like the frontends and commands.

Fixes #643.

### DIFF
--- a/Sources/swift-format/Utilities/FileIterator.swift
+++ b/Sources/swift-format/Utilities/FileIterator.swift
@@ -14,54 +14,82 @@ import Foundation
 
 /// Iterator for looping over lists of files and directories. Directories are automatically
 /// traversed recursively, and we check for files with a ".swift" extension.
-struct FileIterator: Sequence, IteratorProtocol {
+@_spi(Testing)
+public struct FileIterator: Sequence, IteratorProtocol {
 
   /// List of file and directory URLs to iterate over.
-  let urls: [URL]
+  private let urls: [URL]
+
+  /// If true, symlinks will be followed when iterating over directories and files. If not, they
+  /// will be ignored.
+  private let followSymlinks: Bool
 
   /// Iterator for the list of URLs.
-  var urlIterator: Array<URL>.Iterator
+  private var urlIterator: Array<URL>.Iterator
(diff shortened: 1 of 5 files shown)


Sources/swift-format/Utilities/FileIterator.swift
```diff
@@ -14,54 +14,84 @@ import Foundation
...
+          else {
+            break
+          }
+          next = URL(fileURLWithPath: destination)
+          fallthrough
+
+        case .directory:
+          dirIterator = FileManager.default.enumerator(
+            at: next,
+            includingPropertiesForKeys: nil,
```
dylansturg, line 90: You could prefetch the `fileResourceTypeKey` here. I'm not sure how much it'll speed up the iteration, but I think it'll combine the later look-up into whenever the iterator does file IO.

dylansturg (approved):

Sources/swift-format/Utilities/FileIterator.swift
allevato: Unfortunately it looks like the whole `URL.resourceValues` API is busted on Linux. Switched over to `FileManager.attributesOfItem`, which doesn't have the issue (but also doesn't benefit from pre-fetching).

PR #659, opened 2023-10-25
Respect SyntaxVisitorContinueKind of rules when run in Pipeline

thunderseethe:
Pipelines+Generated.swift previously visited all enabled rules at all nodes regardless of what the rule returned for it's `SyntaxVisitorContinueKind`. However, some rules rely on returning `.skipChildren` to lint/format correctly.

To remedy this, modify the pipeline to track the the skip state of each rule. When a rule returns skip, our pipeline stops calling its visit methods until we leave the skipped node.

Alongside this change, modify tests to check both the rule in isolation and the pipeline. We were only testing the rule in isolation so our tests were respecting the `SyntaxVisitorContinueKind`, even though our pipeline was not.


allevato (approved):
Thanks! This will help us avoid a bunch of silly errors in the future.

PR #666, opened 2023-11-20
Various fixes for some 509.0.0 bugs.

### LINKED ISSUE #657: UseShorthandTypeNames incorrect with Sendable functions
If I start with the code

`var a: Optional< @Sendable (Int) -> (Float)> = nil`

and apply `swift-format format --parallel --recursive --in-place Sources` version 509.0.0 the code becomes

`var a: @Sendable (Int) -> (Float)? = nil`

The optionalness has moved from the whole phrase to just the return type.  This has a different meaning.  The new code doesn't compile because of this.

### LINKED ISSUE #663: Key path subscript formatting loses parameter name whitespace
While subscripts with parameter names preserve whitespace, _e.g._ `foo[id: 0]`, the key path version trims it, _e.g._ `\Foo.[id:0]`:

```sh
$ swift-format -v
509.0.0
$ echo "foo[id: 0]" | swift-format
foo[id: 0]
$ echo "\Foo.[id: 0]" | swift-format
\Foo.[id:0]
```

### LINKED ISSUE #665: 509 Regression: swift-format is deleting comments
I don't think swift-format should be proactively deleting code comments, but I'm seeing the following behavior with the latest release:

```sh
$ swift-format
withMacroTesting(
  // isRecording: true
) {
  super.invokeTest()
}
^D
```
```
withMacroTesting {
  super.invokeTest()
}
```

allevato:
- Fix shorthand for optional attributed types in `UseShorthandTypeNames` (fixes #657).
- Fix formatting of keypath subscript components with labels in the pretty printer (fixes #663).
- Do not remove parens if they contain comments in `NoEmptyTrailingClosureParentheses` (fixes #665).

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -1894,9 +1894,29 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   }
 
   override func visit(_ node: KeyPathSubscriptComponentSyntax) -> SyntaxVisitorContinueKind {
-    after(node.leftSquare, tokens: .break(.open, size: 0), .open)
-    before(node.rightSquare, tokens: .break(.close, size: 0), .close)
+    var breakBeforeRightParen = !isCompactSingleFunctionCallArgument(node.arguments)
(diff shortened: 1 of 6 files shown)


allevato:
CI came back green. Will need to wait until the holiday freeze is over to merge.

PR #677, opened 2024-01-09
build: introduce a CMake based build for swift-format

compnerd:
This is in preparation to use SwiftFormat from SourceKit-LSP which is distributed as part of the toolchain.

On Windows, we are now able to build swift-format against the shared Swift Syntax package, yielding an overall size reduction:

SPM
  swift-format.exe: 75,683,840 b
CMake
  swift-format.exe:    830,464 b
  SwiftFormat.dll:   7,818,240 b

Net Savings: 67,035,136 b

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,92 @@
+#[[
+This source file is part of the swift-format open source project
+
+Copyright (c) 2024 Apple Inc. and the swift-format project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+cmake_minimum_required(VERSION 3.19.0)
+
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+if(POLICY CMP0091)
+  cmake_policy(SET CMP0091 NEW)
+endif()
+
+project(SwiftFormat
+  LANGUAGES C Swift)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+set(CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
+
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+
+include(FetchContent)
+include(GNUInstallDirs)
+include(SwiftSupport)
+
+find_package(Foundation CONFIG)
+
+set(_SF_VENDOR_DEPENDENCIES)
+
+set(BUILD_EXAMPLES NO)
+set(BUILD_TESTING NO)
+
+find_package(ArgumentParser CONFIG)
+if(NOT ArgumentParser_FOUND)
+  FetchContent_Declare(ArgumentParser
+    GIT_REPOSITORY https://github.com/apple/swift-argument-parser
+    GIT_TAG 1.2.3)
+  list(APPEND _SF_VENDOR_DEPENDENCIES ArgumentParser)
+endif()
+
+find_package(cmark-gfm CONFIG)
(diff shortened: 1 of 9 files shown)


compnerd:
This does require apple/swift-cmark#69 as well as apple/swift-markdown#141

compnerd:
CC: @ahoppen

Sources/SwiftFormat/CMakeLists.txt
```diff
@@ -0,0 +1,105 @@
+#[[
+This source file is part of the Swift System open source project
+
+Copyright (c) 2020 Apple Inc. and the Swift System project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_library(SwiftFormat
```
allevato, line 10: What safeguards do we have against failing to update these source listings? Will the toolchain builds start using these instead of the SPM build?

Since the CI story on this repository still isn't great (cc @shahmishal), I want to make sure that this change doesn't add undue burden on contributors. Right now, the only coverage we have is through paired PR testing in swift-syntax, which is quite slow.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,105 @@
...
+
+  install(TARGETS ${module}
+    ARCHIVE DESTINATION lib/${swift}/${swift_os}
+    LIBRARY DESTINATION lib/${swift}/${swift_os}
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+  if(type STREQUAL EXECUTABLE)
+    return()
+  endif()
+
+  get_swift_host_arch(swift_arch)
```
allevato, line 86: Just to make sure I understand how this is being installed, is the `swift-format` frontend executable being dynamically linked with the `SwiftFormat` dylib, and then both are installed into the toolchain from the calls to this macro above?

In other words, `swift-format` is not a standalone executable when built with CMake? (This is fine if it's meant to be installed in toolchains, I suppose, but I just want to understand the behavior.)

CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
+#[[
+This source file is part of the Swift System open source project
```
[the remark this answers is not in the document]
compnerd: Copy pasta is difficult 🥲

Sources/SwiftFormat/CMakeLists.txt
compnerd: Correct, I will be putting up changes to start consuming this build for the Windows builds/

cmake/modules/SwiftSupport.cmake
compnerd: That is how this is setup. I however left it to the packager's discretion as to the linking. No explicit library type is specified for `SwiftFormat`, instead that is controlled by the standard `BUILD_SHARED_LIBS` variable. Windows will be distributing `SwiftFormat` as a shared library as that is to permit it to be consumed by both `swift-format` and `SourceKit-LSP` with  https://github.com/apple/sourcekit-lsp/pull/769.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,105 @@
+#[[
```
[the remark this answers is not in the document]
compnerd: Yeah, its replicated everywhere because we don't have a good home for the module. I would like to obsolete it by adding support for installing Swift modules to CMake, but it is the current state. I agree with you though that it is a lot to keep in sync - I've felt the pain as I did the ARM64 fixes for Windows.

CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+  list(APPEND _SF_VENDOR_DEPENDENCIES Markdown)
+endif()
+
+find_package(SwiftSyntax CONFIG)
+if(NOT SwiftSyntax_FOUND)
+  FetchContent_Declare(Syntax
+    GIT_REPOSITORY https://github.com/apple/swift-syntax
+    GIT_TAG main)
+  list(APPEND _SF_VENDOR_DEPENDENCIES Syntax)
+endif()
```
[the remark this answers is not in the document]
compnerd: I did have it working when I tested it, but I might've missed some of the changes (alias targets).

bnbarham (commented):
LGTM now other than the copyright headers still 😅. Same in the other linked PRs.

And would be good to run full toolchain builds in https://github.com/apple/swift/pull/70814 first.

compnerd:
Now that the base change is in, yes, it is possible to do the toolchain test - which is not a "good" but really a "must" as we risk breaking the Windows installer build at the very least and we need to ensure that the packaged content is correct.

bnbarham (approved):

allevato (approved):

PR #679, opened 2024-01-21
Allow cross-compiling swift-format

ahoppen:
As usual, I did some cleanup of `build-script-helper.py` first. Changes are split into multiple commits.

---

Companion of https://github.com/apple/swift/pull/71042

---

This should build swift-format as a fat binary containing both an x86_64 and an arm64 slice in the open source toolchains.

The implementation is mostly stolen from `build-script-helper.py` in the `sourcekit-lsp` repo.

rdar://121326143

### DIFF
--- a/build-script-helper.py
+++ b/build-script-helper.py
@@ -20,6 +20,7 @@ import os, platform
 import subprocess
 from pathlib import Path
 from typing import List, Union, Optional
+import json
 
 # -----------------------------------------------------------------------------
 # General utilities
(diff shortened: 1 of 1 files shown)


allevato (approved):
Thanks!

build-script-helper.py
```diff
@@ -87,8 +106,17 @@ def get_swiftpm_options(
...
+        swift_exec, cross_compile_config=cross_compile_config
+    )
+    build_os = build_target.split("-")[2]
+    if build_os.startswith("macosx"):
+        args += [
+            "-Xlinker",
+            "-rpath",
+            "-Xlinker",
+            "/usr/lib/swift",
+        ]
```
bnbarham, line 119: Do we actually need either of the paths after this?

bnbarham (approved):

build-script-helper.py
ahoppen: I’m not sure. I wanted to start with the same paths that sourcekit-lsp uses and then try to simplify the paths again.

PR #680, opened 2024-01-23
Don’t include the toolchain rpath when installing swift-format

ahoppen:
When installing swift-format on Linux, we don’t want it to have an absolute rpath to where the stdlib was installed on the builder machine.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,6 +14,24 @@
 import Foundation
 import PackageDescription
 
+// MARK: - Parse build arguments
+
+func hasEnvironmentVariable(_ name: String) -> Bool {
+  return ProcessInfo.processInfo.environment[name] != nil
+}
+
+// When building the toolchain on the CI, don't add the CI's runpath for the
+// final build before installing.
+let installAction = hasEnvironmentVariable("SOURCEKIT_LSP_CI_INSTALL")
+
+
+// MARK: - Compute custom build settings
+
+var swiftformatLinkSettings: [LinkerSetting] = []
+if installAction {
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -106,7 +124,8 @@ let package = Package(
         .product(name: "ArgumentParser", package: "swift-argument-parser"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftParser", package: "swift-syntax"),
-      ]
+      ],
+      linkerSettings: swiftformatLinkSettings
```
allevato, line 128: Can this be made into a function that returns the desired value, so that its definition can be moved down to the bottom of the file? That way there's no noise before the package definition itself that average users don't care about.

Package.swift
```diff
@@ -14,6 +14,24 @@
 import Foundation
 import PackageDescription
 
+// MARK: - Parse build arguments
+
+func hasEnvironmentVariable(_ name: String) -> Bool {
```
allevato, line 19: You can also update the `SWIFTCI_USE_LOCAL_DEPS` env var lookup at the bottom of the file to use this.

Package.swift
ahoppen: My goal was to keep the structure similar to that of [sourcekit-lsp’s Package.swift](https://github.com/apple/sourcekit-lsp/blob/main/Package.swift) and [swift-syntax’s Package.swift](https://github.com/apple/swift-syntax/blob/main/Package.swift) for consistency across the toolchain packages. 

I think moving those variables to be functions at the bottom of the file might be a good idea but I’d like to try that across all three packages first and see if it works for all of them. I’ll do it in a follow-up PR.

PR #682, opened 2024-01-25
Order package manifest to start with the package declaration

ahoppen:
Instead of starting with environment variable parsing, start with the actual package manifest, which is most likely the most interesting to most users.

As suggested in https://github.com/apple/swift-format/pull/680#discussion_r1463458831.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,24 +14,6 @@
 import Foundation
 import PackageDescription
 
-// MARK: - Parse build arguments
-
-func hasEnvironmentVariable(_ name: String) -> Bool {
-  return ProcessInfo.processInfo.environment[name] != nil
-}
-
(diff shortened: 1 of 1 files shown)


allevato (approved):
Looks great, thanks!

MahdiBM (commented):
I don't think this is possible to do with simple variables. See: https://github.com/apple/swift/issues/66045
Unless that bug is something even more complicated.

You can, however, use `extension`s to another type at the end of the file. Then use those declarations that are in the `extension`s higher up in the `package` constant.

ahoppen:
That’s why all the variables are computed instead of stored, to work around the limitations of https://forums.swift.org/t/on-the-behavior-of-variables-in-top-level-code/52230.

PR #683, opened 2024-01-25
Clean up rpaths of installed swift-format

ahoppen:
- Don’t explicitly add rpaths to `/usr/lib/swift`, `@executable_path/../lib/swift/macosx` and `@executable_path/../lib/swift-5.5/macosx` on Darwin. I don’t know what they were needed for but they don’t seem to be necessary. Also standard Swift command line tools created from Xcode don’t contain these rpaths.
- Increase the deployment target to macOS 12.0. This removes an rpath in the built binary that is an absolute path to the toolchain on the host, which was used to build swift-format.
- Set `--disable-local-rpath` on Linux. Otherwise the swift-format executable has an `$ORIGIN` rpath on Linux, which we don’t want.
- Rename `SOURCEKIT_LSP_CI_INSTALL` -> `SWIFTFORMAT_CI_INSTALL`. Just cleaning up after copy-pasting.

rdar://121400644

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,8 +17,8 @@ import PackageDescription
 let package = Package(
   name: "swift-format",
   platforms: [
-    .iOS("13.0"),
-    .macOS("10.15")
+    .macOS("12.0"),
+    .iOS("13.0")
   ],
   products: [
     .executable(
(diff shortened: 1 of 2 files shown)


allevato (approved):
I'm not sure if any users will bristle at the higher macOS deployment version requirement. I'm personally ok with it (folks should at least be on Monterey by now, and it allows us to use `Regex`), but I noticed that sourcekit-lsp is still on macOS 12. Do you plan on updating that one as well? I wonder if they should all be consistently using the same value.

ahoppen:
Usually the policy has been that sourcekit-lsp and swift-format should have the same deployment target as the Xcode that is released alongside it (which usually is the current and last OS). IMO we should always be able to bump the deployment target to that version but if there’s no need to increase the deployment target, I think it’s nice to keep it as low as possible.

KidiIT (commented):
All reviewed and approved🫶❤️

KidiIT:
All checked out and approved. Thank you and you're welcome🌚❤️

Issue #684, opened 2024-01-25
Find config from CWD when formatting from stdin


judemille:
It would appear that multiple tools prefer to feed swift-format via stdin.

Currently, swift-format seems to use the default configuration unless a file is specified on the CLI, when formatting a file from stdin. It would be preferable if the utility would find the configuration with the normal logic, but based on the CWD, when reading a file from stdin.

From a cursory look, this shouldn't be too hard to implement. I'll see if I can do it.

allevato:
I think that's a sensible feature to have. Happy to review if you send a PR!

judemille:
Finished via #685.

PR #685, opened 2024-01-25
Find configuration based on CWD when formatting from stdin

judemille:
See #684.

This change is useful for formatting clients that prefer to format via
stdio.

The signature of `ConfigurationLoader.configuration(forSwiftFileAt:URL)`
was changed to `configuration(forPath:URL)` since the function is useful
not just for Swift files, but any arbitrary file path.

### DIFF
--- a/Sources/swift-format/Frontend/Frontend.swift
+++ b/Sources/swift-format/Frontend/Frontend.swift
@@ -213,7 +213,7 @@ class Frontend {
     // then try to load the configuration by inferring it based on the source file path.
     if let swiftFileURL = swiftFileURL {
       do {
-        if let configuration = try configurationLoader.configuration(forSwiftFileAt: swiftFileURL) {
+        if let configuration = try configurationLoader.configuration(forPath: swiftFileURL) {
           self.checkForUnrecognizedRules(in: configuration)
           return configuration
         }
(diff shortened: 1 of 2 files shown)


DaveEwing:
This looks ready to go to me. (I was gonna grab [the issue](https://github.com/apple/swift-format/issues/684) myself as I need this ability too, and then saw this PR.) Requesting a review from @allevato.

DaveEwing (approved):

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -223,11 +223,26 @@ class Frontend {
           "Unable to read configuration for \(swiftFileURL.path): \(error.localizedDescription)")
         return nil
       }
+    } else {
+      // If reading from stdin and no explicit configuration file was given,
+      // walk up the file tree from the cwd to find a config.
+
+      let cwd = URL(FileManager.default.currentDirectoryPath)
```
allevato, line 230: I think this needs to be `URL(fileURLWithPath: ...)`? I don't see an unlabeled `URL.init` documented anywhere, and it doesn't build on my system (Swift 5.9.2, macOS).

allevato (commented):
One comment but otherwise LGTM!

Sources/swift-format/Frontend/Frontend.swift
judemille: Oops, let me fix that. No idea why I didn't catch that.

allevato (approved):
Thank you for the contribution!

PR #687, opened 2024-02-02
Ignore sentence terminators inside quotes when applying the 'BeginDocumentationCommentWithOneLineSummary' option.

### LINKED ISSUE #669: BeginDocumentationCommentWithOneLineSummary complains about punctuation enclosed in backticks (`!`)
If `BeginDocumentationCommentWithOneLineSummary` is enabled it will complain about this declaration:

```
  /// Creates an instance with the same raw value as `x` failing iff `x.kind != Subject.kind`.
  public init?<Other: NodeIDProtocol>(_ x: Other) {
    if x.kind == Subject.kind {
      self.init(rawValue: x.rawValue)
    } else {
      return nil
    }
  }
```

The warning is:

```
warning: [BeginDocumentationCommentWithOneLineSummary] add a blank comment line after this sentence: "Creates an instance with the same raw value as `x` failing iff `x.kind !"
```

TTOzzi:
Resolve #669 

When `SentenceTerminator` is present between `OpenQuote` and `CloseQuote`, changed the filtering logic to ignore it and prevent it from being recognized as a new sentence.

### DIFF
--- a/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
+++ b/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
@@ -130,8 +130,14 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
         in: text.startIndex..<text.endIndex,
         scheme: NSLinguisticTagScheme.lexicalClass.rawValue,
         tokenRanges: &tokenRanges)
+      var isInsideQuotes = false
       let sentenceTerminatorIndices = tags.enumerated().filter {
-        $0.element == "SentenceTerminator"
+        if $0.element == "OpenQuote" {
+          isInsideQuotes = true
+        } else if $0.element == "CloseQuote" {
+          isInsideQuotes = false
+        }
+        return !isInsideQuotes && $0.element == "SentenceTerminator"
       }.map {
         tokenRanges[$0.offset].lowerBound
       }
--- a/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
+++ b/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
@@ -11,6 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+#if os(macOS)
+import NaturalLanguage
+#endif
 import SwiftSyntax
 
 /// All documentation comments must begin with a one-line summary of the declaration.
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
```diff
@@ -130,8 +130,14 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
         in: text.startIndex..<text.endIndex,
         scheme: NSLinguisticTagScheme.lexicalClass.rawValue,
         tokenRanges: &tokenRanges)
+      var isInsideQuotes = false
       let sentenceTerminatorIndices = tags.enumerated().filter {
-        $0.element == "SentenceTerminator"
+        if $0.element == "OpenQuote" {
+          isInsideQuotes = true
+        } else if $0.element == "CloseQuote" {
+          isInsideQuotes = false
```
allevato, line 138: I was going to point out that this won't handle nested quotes correctly, but it looks like the Foundation API itself doesn't categorize them correctly either:

```swift
func tags(_ text: String) {
  var tokenRanges = [Range<String.Index>]()
  let tags = text.linguisticTags(in: text.startIndex..<text.endIndex, scheme: NSLinguisticTagScheme.lexicalClass.rawValue, tokenRanges: &tokenRanges)
  print(Array(tags))
}
tags("\"Hello 'world'\"")
// ["OpenQuote", "Interjection", "Whitespace", "OpenQuote", "Noun", "CloseQuote", "OpenQuote"]
```

So an attempt to increment/decrement a nested quote count won't work as intended. In other words, I wanted to point it out but you don't need to change anything.

Tests/SwiftFormatTests/Rules/BeginDocumentationCommentWithOneLineSummaryTests.swift
```diff
@@ -139,4 +139,24 @@ final class BeginDocumentationCommentWithOneLineSummaryTests: LintOrFormatRuleTe
...
+    assertLint(
+      BeginDocumentationCommentWithOneLineSummary.self,
+      """
+      /// Creates an instance with the same raw value as `x` failing iff `x.kind != Subject.kind`.
+      struct TestBackTick {}
+      
+      /// A set of `Diagnostic` that can answer the question 'was there an error?' in O(1).
+      struct TestSingleQuotes {}
+      
+      /// A set of `Diagnostic` that can answer the question “was there an error?” in O(1).
```
allevato, line 153: Both this example and the one below use smart quotes. Can you add one with regular ASCII (straight) double quotes?

TTOzzi: Sure, I'll add it!

Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
TTOzzi: Yes, this is entirely dependent on the behavior of the Foundation API.
And since `linguisticTags` is a deprecated API, it's unlikely that we'll see any improvements to it 🤔 

The replacement, `NaturalLanguage`, works the same as `linguisticTags`, but I've migrated to it in case it's improved in the future.

Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
```diff
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import NaturalLanguage
```
allevato, line 14: Please add `#if os(macOS)` around this import to match where it's used, since it won't exist on Linux.

TTOzzi: Oh, I missed it.

allevato (approved):
Thanks! There are probably still some tweaks we could do here, but this is still an improvement over the original cade.

TTOzzi:
Thank you 😊

PR #690, opened 2024-02-11
Add BuildToolPlugin and XcodeBuildToolPlugin

BrentMifsud:
Saw a few issues referencing this so I thought I'd implement it.

Unfortunately, it looks like this will need a hosted binary of swift-format in order to run:

Screenshot 2024-02-10 at 11 42 31 PM

Would appreciate it if someone could assist with that.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -34,9 +34,17 @@ let package = Package(
       targets: ["Format Source Code"]
     ),
     .plugin(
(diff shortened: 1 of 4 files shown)


BrentMifsud:
Closing this as there does not seem to be any way to produce a binary, while also relying on the same binary.

Probably need to make the build plugin in a separate repo.

PR #696, opened 2024-03-04
Renamed plugin configuration parameter

### LINKED ISSUE #668: Can't run plugin with --configuration argument
I'm trying to run the following command locally by using the swift-format linting plugin.
`swift package lint-source-code --configuration .swift-format.json`
where `.swift-format.json` just contains the default config output by `swift-config dump-configuration` (created by the brew installed version of swift-format, not the plugin)
This results in the following error which seems to be a swift build related error instead of related to finding the configuration file. 
`error: The value '.swift-format.json' is invalid for '--configuration <configuration>'. Please provide one of 'debug' and 'release'.`

Interestingly enough this give me a different error when I pass the argument `--configuration debug` then it prints: 
`<unknown>: error: Unable to read configuration: The file “debug” couldn’t be opened because there is no such file.`

Am I using the plugin in the wrong way or is this a bug?

dmonagle:
As trivial as this is, this change is to prevent a conflict with running the plugin with the `--configuration` parameter which is recognised and parsed by the Swift Package Manager and expects to be either `debug` or `release`.

The replacement of `--swift-format-configuration` is verbose, but other choices looked too ambiguous or confusing whereas this is at least clear as to what the parameter is doing.

resolves  #668

### DIFF
--- a/Plugins/FormatPlugin/plugin.swift
+++ b/Plugins/FormatPlugin/plugin.swift
@@ -40,7 +40,7 @@ extension FormatPlugin: CommandPlugin {
     let targetNames = argExtractor.extractOption(named: "target")
     let targetsToFormat = targetNames.isEmpty ? context.package.targets : try context.package.targets(named: targetNames)
     
-    let configurationFilePath = argExtractor.extractOption(named: "configuration").first
+    let configurationFilePath = argExtractor.extractOption(named: "swift-format-configuration").first
     
     let sourceCodeTargets = targetsToFormat.compactMap{ $0 as? SourceModuleTarget }
     
@@ -60,7 +60,7 @@ extension FormatPlugin: XcodeCommandPlugin {
     let swiftFormatTool = try context.tool(named: "swift-format")
     
     var argExtractor = ArgumentExtractor(arguments)
-    let configurationFilePath = argExtractor.extractOption(named: "configuration").first
+    let configurationFilePath = argExtractor.extractOption(named: "swift-format-configuration").first
     
     try format(
       tool: swiftFormatTool,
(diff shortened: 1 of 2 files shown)


ahoppen (commented):
I don’t think we can change this as tooling might depend on the parameter being named `--configuration`. sourcekit-lsp, for example, does: https://github.com/apple/sourcekit-lsp/blob/cbb569320831d2aa48d251cadf63199c8f7ce840/Sources/SourceKitLSP/Swift/DocumentFormatting.swift#L185

We could add `swift-format-configuration` as another spelling of the same option, but I think we should keep `--configuration` as well.

allevato:
sourcekit-lsp is invoking the `swift-format` tool directly, right, not using the plugin? In that case, I don't think it wouldn't be affected; from what I can tell, this PR is only changing the name of the flag that the SPM plugins look for, not the flag that `swift-format` itself uses.

Separately, the macOS CI is failing but I'm not sure why—I don't see anything relevant there, the failure appears to be a TSAN issue.

dmonagle:
As @allevato said, existing tooling that is using `--configuration` would be doing it via the `swift-format` executable, not via the plugin. There doesn't seem to be any valid way to use custom configuration via the plugin due to the bug that this PR was created to resolve. I don't believe this will have an adverse affects, but it certainly makes the plugin usable for those of us that need to change the default behaviour.

ahoppen (approved):

ahoppen:
Ah, OK. Sorry, I missed the distinction between the binary and the plugin. LGTM then

dmonagle:
> Ah, OK. Sorry, I missed the distinction between the binary and the plugin. LGTM then

All good. I still think it’s a bit cringey to have to do this, but unless SPM changes its behaviour to ignore command line options after the module name, this is the only way around it. Might be worth raising to avoid things like this in the future.

PR #708, opened 2024-03-22
Support for formatting a selection

DaveEwing:
Work in progress at the moment. I have a handful of tweaks I want to make, but it appears to work pretty well overall.

### DIFF
--- a/Sources/SwiftFormat/API/Selection.swift
+++ b/Sources/SwiftFormat/API/Selection.swift
@@ -19,11 +19,11 @@ public enum Selection {
   case ranges([Range<AbsolutePosition>])
 
   /// Create a selection from an array of utf8 ranges. An empty array means an infinite selection.
-  public init(offsetPairs: [Range<Int>]) {
-    if offsetPairs.isEmpty {
+  public init(offsetRanges: [Range<Int>]) {
+    if offsetRanges.isEmpty {
       self = .infinite
     } else {
-      let ranges = offsetPairs.map {
+      let ranges = offsetRanges.map {
         AbsolutePosition(utf8Offset: $0.lowerBound) ..< AbsolutePosition(utf8Offset: $0.upperBound)
       }
       self = .ranges(ranges)
@@ -51,7 +51,7 @@ public enum Selection {
 
 
 public extension Syntax {
-  /// return true if the node is _completely_ inside any range in the selection
+  /// - Returns: `true` if the node is _completely_ inside any range in the selection
   func isInsideSelection(_ selection: Selection) -> Bool {
     switch selection {
     case .infinite:
--- a/Sources/SwiftFormat/API/Selection.swift
+++ b/Sources/SwiftFormat/API/Selection.swift
@@ -1,0 +1,63 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import SwiftSyntax
+
+/// The selection as given on the command line - an array of offets and lengths
+public enum Selection {
+  case infinite
+  case ranges([Range<AbsolutePosition>])
+
+  /// Create a selection from an array of utf8 ranges. An empty array means an infinite selection.
+  public init(offsetPairs: [Range<Int>]) {
+    if offsetPairs.isEmpty {
+      self = .infinite
+    } else {
+      let ranges = offsetPairs.map {
+        AbsolutePosition(utf8Offset: $0.lowerBound) ..< AbsolutePosition(utf8Offset: $0.upperBound)
+      }
+      self = .ranges(ranges)
+    }
+  }
+
+  public func contains(_ position: AbsolutePosition) -> Bool {
+    switch self {
+    case .infinite:
+      return true
+    case .ranges(let ranges):
+      return ranges.contains { $0.contains(position) }
+    }
+  }
+
+  public func overlapsOrTouches(_ range: Range<AbsolutePosition>) -> Bool {
+    switch self {
+    case .infinite:
+      return true
+    case .ranges(let ranges):
+      return ranges.contains { $0.overlapsOrTouches(range) }
+    }
+  }
+}
+
+
+public extension Syntax {
+  /// return true if the node is _completely_ inside any range in the selection
+  func isInsideSelection(_ selection: Selection) -> Bool {
+    switch selection {
+    case .infinite:
+      return true
+    case .ranges(let ranges):
+      return ranges.contains { return $0.lowerBound <= position && endPosition <= $0.upperBound }
+    }
+  }
+}
--- a/Sources/SwiftFormat/API/SwiftFormatter.swift
+++ b/Sources/SwiftFormat/API/SwiftFormatter.swift
@@ -21,9 +21,6 @@ public final class SwiftFormatter {
   /// The configuration settings that control the formatter's behavior.
   public let configuration: Configuration
 
-  /// the ranges of text to format
-  public var selection: Selection = .infinite
-
   /// An optional callback that will be notified with any findings encountered during formatting.
   public let findingConsumer: ((Finding) -> Void)?
 
--- a/Sources/SwiftFormat/API/SwiftFormatter.swift
+++ b/Sources/SwiftFormat/API/SwiftFormatter.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -20,6 +20,9 @@ public final class SwiftFormatter {
 
   /// The configuration settings that control the formatter's behavior.
   public let configuration: Configuration
+
+  /// the ranges of text to format
+  public var selection: Selection = .infinite
 
   /// An optional callback that will be notified with any findings encountered during formatting.
   public let findingConsumer: ((Finding) -> Void)?
@@ -70,6 +73,7 @@ public final class SwiftFormatter {
     try format(
       source: String(contentsOf: url, encoding: .utf8),
       assumingFileURL: url,
+      selection: .infinite,
       to: &outputStream,
       parsingDiagnosticHandler: parsingDiagnosticHandler)
   }
@@ -86,6 +90,7 @@ public final class SwiftFormatter {
   ///   - url: A file URL denoting the filename/path that should be assumed for this syntax tree,
   ///     which is associated with any diagnostics emitted during formatting. If this is nil, a
   ///     dummy value will be used.
+  ///   - selection: The ranges to format
   ///   - outputStream: A value conforming to `TextOutputStream` to which the formatted output will
   ///     be written.
   ///   - parsingDiagnosticHandler: An optional callback that will be notified if there are any
@@ -94,6 +99,7 @@ public final class SwiftFormatter {
   public func format<Output: TextOutputStream>(
     source: String,
     assumingFileURL url: URL?,
+    selection: Selection,
     to outputStream: inout Output,
     parsingDiagnosticHandler: ((Diagnostic, SourceLocation) -> Void)? = nil
   ) throws {
@@ -108,8 +114,8 @@ public final class SwiftFormatter {
       assumingFileURL: url,
       parsingDiagnosticHandler: parsingDiagnosticHandler)
     try format(
-      syntax: sourceFile, operatorTable: .standardOperators, assumingFileURL: url, source: source,
-      to: &outputStream)
+      syntax: sourceFile, source: source, operatorTable: .standardOperators, assumingFileURL: url,
+      selection: selection, to: &outputStream)
   }
 
   /// Formats the given Swift syntax tree and writes the result to an output stream.
@@ -122,32 +128,26 @@ public final class SwiftFormatter {
   ///
   /// - Parameters:
   ///   - syntax: The Swift syntax tree to be converted to source code and formatted.
+  ///   - source: The original Swift source code used to build the syntax tree.
   ///   - operatorTable: The table that defines the operators and their precedence relationships.
   ///     This must be the same operator table that was used to fold the expressions in the `syntax`
   ///     argument.
   ///   - url: A file URL denoting the filename/path that should be assumed for this syntax tree,
   ///     which is associated with any diagnostics emitted during formatting. If this is nil, a
   ///     dummy value will be used.
+  ///   - selection: The ranges to format
   ///   - outputStream: A value conforming to `TextOutputStream` to which the formatted output will
   ///     be written.
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func format<Output: TextOutputStream>(
-    syntax: SourceFileSyntax, operatorTable: OperatorTable, assumingFileURL url: URL?,
-    to outputStream: inout Output
-  ) throws {
-    try format(
-      syntax: syntax, operatorTable: operatorTable, assumingFileURL: url, source: nil,
-      to: &outputStream)
-  }
-
-  private func format<Output: TextOutputStream>(
-    syntax: SourceFileSyntax, operatorTable: OperatorTable,
-    assumingFileURL url: URL?, source: String?, to outputStream: inout Output
+    syntax: SourceFileSyntax, source: String, operatorTable: OperatorTable,
+    assumingFileURL url: URL?, selection: Selection, to outputStream: inout Output
   ) throws {
     let assumedURL = url ?? URL(fileURLWithPath: "source")
     let context = Context(
       configuration: configuration, operatorTable: operatorTable, findingConsumer: findingConsumer,
-      fileURL: assumedURL, sourceFileSyntax: syntax, source: source, ruleNameCache: ruleNameCache)
+      fileURL: assumedURL, selection: selection, sourceFileSyntax: syntax, source: source,
+      ruleNameCache: ruleNameCache)
     let pipeline = FormatPipeline(context: context)
     let transformedSyntax = pipeline.rewrite(Syntax(syntax))
 
@@ -158,6 +158,7 @@ public final class SwiftFormatter {
 
     let printer = PrettyPrinter(
       context: context,
+      source: source,
       node: transformedSyntax,
       printTokenStream: debugOptions.contains(.dumpTokenStream),
       whitespaceOnly: false)
--- a/Sources/SwiftFormat/API/SwiftLinter.swift
+++ b/Sources/SwiftFormat/API/SwiftLinter.swift
@@ -119,17 +119,18 @@ public final class SwiftLinter {
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func lint(
     syntax: SourceFileSyntax,
+    source: String,
     operatorTable: OperatorTable,
     assumingFileURL url: URL
   ) throws {
-    try lint(syntax: syntax, operatorTable: operatorTable, assumingFileURL: url, source: nil)
+    try lint(syntax: syntax, operatorTable: operatorTable, assumingFileURL: url, source: source)
   }
 
   private func lint(
     syntax: SourceFileSyntax,
     operatorTable: OperatorTable,
     assumingFileURL url: URL,
-    source: String?
+    source: String
   ) throws {
     let context = Context(
       configuration: configuration, operatorTable: operatorTable, findingConsumer: findingConsumer,
@@ -145,6 +146,7 @@ public final class SwiftLinter {
     // pretty-printer.
     let printer = PrettyPrinter(
       context: context,
+      source: source,
       node: Syntax(syntax),
       printTokenStream: debugOptions.contains(.dumpTokenStream),
       whitespaceOnly: true)
--- a/Sources/SwiftFormat/Core/Context.swift
+++ b/Sources/SwiftFormat/Core/Context.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -39,6 +39,9 @@ public final class Context {
   /// The configuration for this run of the pipeline, provided by a configuration JSON file.
   let configuration: Configuration
 
+  /// The optional ranges to process
+  let selection: Selection
+
   /// Defines the operators and their precedence relationships that were used during parsing.
   let operatorTable: OperatorTable
 
@@ -66,6 +69,7 @@ public final class Context {
     operatorTable: OperatorTable,
     findingConsumer: ((Finding) -> Void)?,
     fileURL: URL,
+    selection: Selection = .infinite,
     sourceFileSyntax: SourceFileSyntax,
     source: String? = nil,
     ruleNameCache: [ObjectIdentifier: String]
@@ -74,6 +78,7 @@ public final class Context {
     self.operatorTable = operatorTable
     self.findingEmitter = FindingEmitter(consumer: findingConsumer)
     self.fileURL = fileURL
+    self.selection = selection
     self.importsXCTest = .notDetermined
     let tree = source.map { Parser.parse(source: $0) } ?? sourceFileSyntax
     self.sourceLocationConverter =
@@ -86,8 +91,10 @@ public final class Context {
   }
 
   /// Given a rule's name and the node it is examining, determine if the rule is disabled at this
-  /// location or not.
+  /// location or not. Also makes sure the entire node is contained inside any selection.
   func isRuleEnabled<R: Rule>(_ rule: R.Type, node: Syntax) -> Bool {
+    guard node.isInsideSelection(selection) else { return false }
+
     let loc = node.startLocation(converter: self.sourceLocationConverter)
 
     assert(
--- a/Sources/SwiftFormat/Core/Context.swift
+++ b/Sources/SwiftFormat/Core/Context.swift
@@ -39,7 +39,7 @@ public final class Context {
   /// The configuration for this run of the pipeline, provided by a configuration JSON file.
   let configuration: Configuration
 
-  /// The optional ranges to process
+  /// The selection to process
   let selection: Selection
 
   /// Defines the operators and their precedence relationships that were used during parsing.
@@ -92,7 +92,7 @@ public final class Context {
 
   /// Given a rule's name and the node it is examining, determine if the rule is disabled at this
   /// location or not. Also makes sure the entire node is contained inside any selection.
-  func isRuleEnabled<R: Rule>(_ rule: R.Type, node: Syntax) -> Bool {
+  func shouldFormat<R: Rule>(_ rule: R.Type, node: Syntax) -> Bool {
     guard node.isInsideSelection(selection) else { return false }
 
     let loc = node.startLocation(converter: self.sourceLocationConverter)
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftSyntax
+import Foundation
 
 /// PrettyPrinter takes a Syntax node and outputs a well-formatted, re-indented reproduction of the
 /// code as a String.
@@ -66,6 +67,19 @@ public class PrettyPrinter {
   private var configuration: Configuration { return context.configuration }
   private let maxLineLength: Int
   private var tokens: [Token]
+  private var source: String
+
+  /// keep track of where formatting was disabled in the original source
+  ///
+  /// To format a selection, we insert `enableFormatting`/`disableFormatting` tokens into the
+  /// stream when entering/exiting a selection range. Those tokens include utf8 offsets into the
+  /// original source. When enabling formatting, we copy the text between `disabledPosition` and the
+  /// current position to `outputBuffer`. From then on, we continue to format until the next
+  /// `disableFormatting` token.
+  private var disabledPosition: AbsolutePosition? = nil
+  /// true if we're currently formatting
+  private var writingIsEnabled: Bool { disabledPosition == nil }
+
   private var outputBuffer: String = ""
 
   /// The number of spaces remaining on the current line.
@@ -172,11 +186,14 @@ public class PrettyPrinter {
   ///   - printTokenStream: Indicates whether debug information about the token stream should be
   ///     printed to standard output.
   ///   - whitespaceOnly: Whether only whitespace changes should be made.
-  public init(context: Context, node: Syntax, printTokenStream: Bool, whitespaceOnly: Bool) {
+  public init(context: Context, source: String, node: Syntax, printTokenStream: Bool, whitespaceOnly: Bool) {
     self.context = context
+    self.source = source
     let configuration = context.configuration
     self.tokens = node.makeTokenStream(
-      configuration: configuration, operatorTable: context.operatorTable)
+      configuration: configuration,
+      selection: context.selection,
+      operatorTable: context.operatorTable)
     self.maxLineLength = configuration.lineLength
     self.spaceRemaining = self.maxLineLength
     self.printTokenStream = printTokenStream
@@ -216,7 +233,9 @@ public class PrettyPrinter {
     }
 
     guard numberToPrint > 0 else { return }
-    writeRaw(String(repeating: "\n", count: numberToPrint))
+    if writingIsEnabled {
+      writeRaw(String(repeating: "\n", count: numberToPrint))
+    }
     lineNumber += numberToPrint
     isAtStartOfLine = true
     consecutiveNewlineCount += numberToPrint
@@ -238,13 +257,17 @@ public class PrettyPrinter {
   /// leading spaces that are required before the text itself.
   private func write(_ text: String) {
     if isAtStartOfLine {
-      writeRaw(currentIndentation.indentation())
+      if writingIsEnabled {
+        writeRaw(currentIndentation.indentation())
+      }
       spaceRemaining = maxLineLength - currentIndentation.length(in: configuration)
       isAtStartOfLine = false
-    } else if pendingSpaces > 0 {
+    } else if pendingSpaces > 0 && writingIsEnabled {
       writeRaw(String(repeating: " ", count: pendingSpaces))
     }
-    writeRaw(text)
+    if writingIsEnabled {
+      writeRaw(text)
+    }
     consecutiveNewlineCount = 0
     pendingSpaces = 0
   }
@@ -523,7 +546,9 @@ public class PrettyPrinter {
       }
 
     case .verbatim(let verbatim):
-      writeRaw(verbatim.print(indent: currentIndentation))
+      if writingIsEnabled {
+        writeRaw(verbatim.print(indent: currentIndentation))
+      }
       consecutiveNewlineCount = 0
       pendingSpaces = 0
       lastBreak = false
@@ -568,6 +593,40 @@ public class PrettyPrinter {
       if shouldWriteComma {
         write(",")
         spaceRemaining -= 1
+      }
+
+    case .enableFormatting(let enabledPosition):
+      // if we're not disabled, we ignore the token
+      if let disabledPosition {
+        let start = source.utf8.index(source.utf8.startIndex, offsetBy: disabledPosition.utf8Offset)
+        let end: String.Index
+        if let enabledPosition {
+          end = source.utf8.index(source.utf8.startIndex, offsetBy: enabledPosition.utf8Offset)
+        } else {
+          end = source.endIndex
+        }
+        var text = String(source[start..<end])
+        // strip trailing whitespace so that the next formatting can add the right amount
+        if let nonWhitespace = text.rangeOfCharacter(
+          from: CharacterSet.whitespaces.inverted, options: .backwards) {
+          text = String(text[..<nonWhitespace.upperBound])
+        }
+
+        writeRaw(text)
+        if text.hasSuffix("\n") {
+          isAtStartOfLine = true
+          consecutiveNewlineCount = 1
+        } else {
+          isAtStartOfLine = false
+          consecutiveNewlineCount = 0
+        }
+        self.disabledPosition = nil
+      }
+
+    case .disableFormatting(let newPosition):
+      // a second disable is ignored
+      if writingIsEnabled {
+        disabledPosition = newPosition
       }
     }
   }
@@ -673,6 +732,10 @@ public class PrettyPrinter {
         let length = isSingleElement ? 0 : 1
         total += length
         lengths.append(length)
+
+      case .enableFormatting, .disableFormatting:
+        // no effect on length calculations
+        lengths.append(0)
       }
     }
 
@@ -775,6 +838,14 @@ public class PrettyPrinter {
     case .contextualBreakingEnd:
       printDebugIndent()
       print("[END BREAKING CONTEXT Idx: \(idx)]")
+
+    case .enableFormatting(let pos):
+      printDebugIndent()
+      print("[ENABLE FORMATTING utf8 offset: \(String(describing: pos))]")
+
+    case .disableFormatting(let pos):
+      printDebugIndent()
+      print("[DISABLE FORMATTING utf8 offset: \(pos)]")
     }
   }
 
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -69,7 +69,7 @@ public class PrettyPrinter {
   private var tokens: [Token]
   private var source: String
 
-  /// keep track of where formatting was disabled in the original source
+  /// Keep track of where formatting was disabled in the original source
   ///
   /// To format a selection, we insert `enableFormatting`/`disableFormatting` tokens into the
   /// stream when entering/exiting a selection range. Those tokens include utf8 offsets into the
@@ -77,8 +77,6 @@ public class PrettyPrinter {
   /// current position to `outputBuffer`. From then on, we continue to format until the next
   /// `disableFormatting` token.
   private var disabledPosition: AbsolutePosition? = nil
-  /// true if we're currently formatting
-  private var writingIsEnabled: Bool { disabledPosition == nil }
 
   private var outputBuffer: String = ""
 
@@ -204,7 +202,9 @@ public class PrettyPrinter {
   ///
   /// No further processing is performed on the string.
   private func writeRaw<S: StringProtocol>(_ str: S) {
-    outputBuffer.append(String(str))
+    if disabledPosition == nil {
+      outputBuffer.append(String(str))
+    }
   }
 
   /// Writes newlines into the output stream, taking into account any preexisting consecutive
@@ -233,9 +233,7 @@ public class PrettyPrinter {
     }
 
     guard numberToPrint > 0 else { return }
-    if writingIsEnabled {
-      writeRaw(String(repeating: "\n", count: numberToPrint))
-    }
+    writeRaw(String(repeating: "\n", count: numberToPrint))
     lineNumber += numberToPrint
     isAtStartOfLine = true
     consecutiveNewlineCount += numberToPrint
@@ -257,17 +255,13 @@ public class PrettyPrinter {
   /// leading spaces that are required before the text itself.
   private func write(_ text: String) {
     if isAtStartOfLine {
-      if writingIsEnabled {
-        writeRaw(currentIndentation.indentation())
-      }
+      writeRaw(currentIndentation.indentation())
       spaceRemaining = maxLineLength - currentIndentation.length(in: configuration)
       isAtStartOfLine = false
-    } else if pendingSpaces > 0 && writingIsEnabled {
+    } else if pendingSpaces > 0  {
       writeRaw(String(repeating: " ", count: pendingSpaces))
     }
-    if writingIsEnabled {
-      writeRaw(text)
-    }
+    writeRaw(text)
     consecutiveNewlineCount = 0
     pendingSpaces = 0
   }
@@ -546,9 +540,7 @@ public class PrettyPrinter {
       }
 
     case .verbatim(let verbatim):
-      if writingIsEnabled {
-        writeRaw(verbatim.print(indent: currentIndentation))
-      }
+      writeRaw(verbatim.print(indent: currentIndentation))
       consecutiveNewlineCount = 0
       pendingSpaces = 0
       lastBreak = false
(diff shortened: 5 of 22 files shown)


Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,76 @@
...
+      return maxStart < minEnd
+    }
+  }
+}
+
+
+public extension Syntax {
+  /// return true if the node is completely inside any range in the selection (or if there's
+  /// no selection)
+  func isInsideSelection(_ selection: Selection?) -> Bool {
```
allevato, line 68: What if we had a default-initialized `Selection` represent an "infinite" selection that contains any offset? Then we could get rid of the optionality everywhere, simplifying the API.

Then, with a `Selection` always being present, we could invert this API and make it a method on `Selection` (`Selection.contains(_ node: Syntax)`) instead of a `Syntax` extension.

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -26,6 +26,18 @@ struct LintFormatOptions: ParsableArguments {
...
+  /// A JSON string containing the selected ranges to process.
+  ///
+  /// If not specified, the whole file will be formatted.
+  @Option(
+    name: .long,
+    help: """
+      A JSON string containing an array of {"offset":<Int>, "length":<Int>} pairs specifying \
+      the source code ranges to format.
+      """)
+  var selection: String?
```
allevato, line 38: This might just be for testing, but I think we can improve the user interface a little bit here; how about something like `--lines <start line>:<end line>` for line-based ranges and `--offsets <start offset>:<end offset>` for offset-based ranges? (I don't think "length" is a particularly good way to express this on the command line.) Those flags are a bit closer to what clang-format provides for its command line, and we could accept multiple occurrences of it.

To deal with parsing the `s:e` format, you can use swift-argument-parser's `ExpressibleByArgument` protocol instead of JSON decoding. Maybe the option type is just a pair of ints that implements `ExpressibleByArgument` and then the `validate` method constructs the `Selection` from any `--lines` and `--offsets` that it's given.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -775,6 +816,14 @@ public class PrettyPrinter {
     case .contextualBreakingEnd:
       printDebugIndent()
       print("[END BREAKING CONTEXT Idx: \(idx)]")
+
+    case .enableFormatting(let pos):
+      printDebugIndent()
+      print("[ENABLE FORMATTTING utf8 offset: \(pos)]")
```
allevato, line 822:
```suggestion
      print("[ENABLE FORMATTING utf8 offset: \(pos)]")
```

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -775,6 +816,14 @@ public class PrettyPrinter {
...
+
+    case .disableFormatting(let pos):
+      printDebugIndent()
+      print("[DISABLE FORMATTTING utf8 offset: \(pos)]")
```
allevato, line 826:
```suggestion
      print("[DISABLE FORMATTING utf8 offset: \(pos)]")
```

allevato (commented):
This is looking really great! A few high-level comments to start that are mainly API/UX things.

Can you plumb through a test or two that shows the whitespace issues you're seeing? That would make it easier to help debug what's going on.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
DaveEwing: 🙃

Sources/swift-format/Subcommands/LintFormatOptions.swift
DaveEwing: I like the idea of supporting `--lines <start line>:<end line>` as clang-format does. This seems _much_ easier for specifying ranges at the command line. The one issue there is having to convert to `AbsolutePosition`s, since that's what we have with a `Syntax` struct. But we could do that conversion at the start. (I'm already worried about performance of all the range comparisons that happen. We don't want to make that harder.) Anyhow, maybe this can be a second PR?

I was modeling the offset/length bit (kinda) off what clang-format does. It can take multiple `--offset`/`--length` pairs. (Would swift-argument-parser support that? I assumed not, but I admit that I never looked.) I ended up using JSON decoding because it was easy. 🤷‍♂️ clang-format doesn't support `--offsets <start offset>:<end offset>`, so there's no precedent there. My expectation is that a human wouldn't ever specify ranges with offsets (or offsets/lengths). It's just not convenient. But it is convenient for an editor shelling out to swift-format.

If I do manage to change `Selection.Range` to be a `typealias` to `Range<AbsolutePosition>`, then it might be a bit more convenient to pass two offsets, rather than offset/length. Do you have a strong opinion as to whether I should bother with this refactoring? I made a start down that road, and it was a kinda annoying set of changes. 🤣

DaveEwing:
> This is looking really great! A few high-level comments to start that are mainly API/UX things.
> 
> Can you plumb through a test or two that shows the whitespace issues you're seeing? That would make it easier to help debug what's going on.

I've got testing support plumbed through, but I fixed the whitespace issues I was seeing.

Sources/SwiftFormat/API/Selection.swift
DaveEwing: Actually, one thing I want to do is disable the handling of `swift-format-ignore` comments when there's a selection. (If a user's making a selection, then they might actually want to fix part of the code where formatting is normally disabled.) This is one case where we actually need to know if there's a selection. There might end up being others. Keeping `selection` optional allows for this.

(I'm not at all averse to inverting that API though. I don't like that it currently works in both directions. Should be cleaner.)

DaveEwing:
A note on performance.

To be honest, I was rather worried about the performance of this approach. So, I did a couple of tests on a release build of `swift-format` using `TokenStreamCreator.swift` from the project (which is 4268 lines long). I used the `--measure-instructions` option to show how many instructions are executed when formatting the file. I compared this branch to `main`.

The tests are:
- formatting the whole file (no selection)
- formatting a selection of the whole file
- formatting 200 lines of the file (starting at line 2000)

Here are the results:

|                     | No selection    |    Whole file selected   |  Selection lines 2000-2200 |
| ------------------- | --------------- | ------------------------ | ---------------- |
| Before              | 4073650908      |                          |                  |
| After               | 4166960035      |    4230133360            |  1552009297      |
                        
In all cases, formatting finished in well under a second in real time. (I didn't precisely meassure this.)

Formatting the whole file without a selection is about 2% slower. Formatting the whole file with a selection is less than 4% slower. Formatting a small range is _significantly_ faster. That shouldn't really be a surprise, but it is nice to confirm!

Sources/swift-format/Subcommands/LintFormatOptions.swift
allevato: My own personal opinion is that ranges are a bit more natural than an offset/length pair. I'm not sure what motivated clang-format's `--offset` and `--length` flags unless it was some very specific workflow that already had the information in that form, but I don't think ordered pairs of `--offset X --length Y ...` would work well in swift-argument-parser or for human users.

Sources/SwiftFormat/API/Selection.swift
allevato: I think I disagree with that; a `swift-format-ignore` comment is an explicit request that the formatter not touch that section of code under any circumstances.

Imagine this workflow: a user has some nested array/dictionary structure that they've meticulously aligned into columns, so they use `swift-format-ignore` to ensure that their horizontal spaces don't get collapsed. Then, they make some changes to that table (add a new row, or move things around), and finally, they run a hypothetical `git-swift-format` (à la `git-clang-format`) that passes a selection representing the lines that they changed. With the proposed behavior, it would reformat parts of that table that they asked not to be changed.

If you want to provide a way to "ignore the `ignore`s", I think that should be a separate feature (I'm not sure how useful it would be) and is orthogonal to whether a selection is passed.

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
```diff
@@ -0,0 +1,231 @@
+import SwiftFormat
+import XCTest
+
+final class Selections: PrettyPrintTestCase {
+  func testSelectAll() {
+    let input =
+      """
+      ➡️func foo() {
```
allevato, line 8: I wish these arrows showed up better in GitHub 🙁   I see that in Xcode they're rendered as emoji, but GitHub is rendering them using the non-emoji variant, and they're not even the same size in the font that GitHub uses on my machine.

image

How would you feel about using guillemets (`«...»`) instead? Those are unlikely to appear in Swift code that we'd use in tests, they're a natural bracket form, and they're easy to type on a Mac keyboard in US layout (Option+Backslash and Shift+Option+Backslash).

Sources/swift-format/Subcommands/LintFormatOptions.swift
ahoppen: FWIW, I would also prefer `--lines <start line>:<end line>`. And if we do the conversion up front with a `SourceLocationConverter`, I don’t think it should be too expensive. `SourceLocationConverter` needs to build up a line table but I would be surprised if that’s the same order of magnitude as formatting (would be nice to confirm though).

allevato: The `SourceLocationConverter` is already being created in other code paths anyway, so using it to map the line numbers to offsets or vice versa shouldn't create additional overhead; we just need to make sure we're using the same one and not creating a separate one just for that purpose.

DaveEwing: Yeah, I like `--lines <start line>:<end line>` as well. For an actual user, this is way more useful than offsets. But offsets works well when integrating with other tools. (It lets you, for instance, reformat spacing within a single line without affecting anything else. That's something you might want to do from within an editor.) Chatting with Tony on Friday we also agreed that using `--offsets <start offset>:<end offset>` is cleaner. I'll make that change in this PR. Supporting `--lines <start line>:<end line>` can come in a followup PR.

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
DaveEwing: I like guillemets, and it's what I actually thought of first. But I wish they showed up better like the boxed number emojis we're using.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -68,17 +69,32 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+      isInsideSelection = false
+    }
+
     // Because `walk` takes an `inout` argument, and we're a class, we have to do the following
     // dance to pass ourselves in.
     self.walk(node)
+
+    // Make sure we output any trailing text after the last selection range
+    if selection != nil {
+      appendToken(.enableFormatting(nil))
```
ahoppen, line 96: Do we also need to reset `isInsideSelection` to `true` here?

Also: Do you know if `makeStream` ever gets called for more than one node so that this `.enableFormatting` actually matters?

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2716,18 +2732,40 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+    generateDisable(token.endPositionBeforeTrailingTrivia)
+
     appendTrailingTrivia(token)
     appendAfterTokensAndTrailingComments(token)
 
     // It doesn't matter what we return here, tokens do not have children.
     return .skipChildren
   }
 
+  func generateEnable(_ range: Selection.Range) {
```
ahoppen, line 2753: I would name this something like `generateEnableFormattingIfNecessary`. `generateEnable` to me sounds like it unconditionally enables formatting. Same for `generateDisable`.

---

Can `generateEnable` and `generateDisable` be `private`?

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3991,14 +4054,14 @@ class CommentMovingRewriter: SyntaxRewriter {
   }
 
   override func visit(_ node: CodeBlockItemSyntax) -> CodeBlockItemSyntax {
-    if shouldFormatterIgnore(node: Syntax(node)) {
+    if shouldFormatterIgnore(node: Syntax(node)) || !Syntax(node).isInsideSelection(selection) {
```
ahoppen, line 4057: Don’t we need to perform the `isInsideSelection` check for all syntax nodes instead of just `CodeBlockItemSyntax` and `MemberBlockItemSyntax`?

Sources/SwiftFormat/PrettyPrint/Token.swift
```diff
@@ -196,6 +198,13 @@ enum Token {
   /// Ends a scope where `contextual` breaks have consistent behavior.
   case contextualBreakingEnd
 
+  /// Turn formatting back on at the given position in the original file
+  /// nil is used to indicate the rest of the file should be output
+  case enableFormatting(AbsolutePosition?)
+
+  /// Turn formatting off at the given position in the original file.
+  case disableFormatting(AbsolutePosition)
+
```
ahoppen, line 207: Maybe I’m missing something here but why do `enableFormatting` and `disableFormatting` need to carry a position? Shouldn’t we enable/disable formatting exactly where the `enableFormatting` and `disableFormatting` token occur within the token stream?

---

Design question for @allevato: Should these be unified into the `printControl` case? Or would it make sense to have a single case `enableFormattingControl` with a sub-enum like `PrinterControlKind` as we do for `printerControl`?

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -569,6 +584,40 @@ public class PrettyPrinter {
...
+          end = source.utf8.index(source.utf8.startIndex, offsetBy: enabledPosition.utf8Offset)
+        } else {
+          end = source.endIndex
+        }
+        var text = String(source[start..<end])
+        // strip trailing whitespace so that the next formatting can add the right amount
+        if let nonWhitespace = text.rangeOfCharacter(
+          from: CharacterSet.whitespaces.inverted, options: .backwards) {
+          text = String(text[..<nonWhitespace.upperBound])
+        }
```
ahoppen, line 603: How does the printer pick up the whitespace characters that were dropped here? Ie. how does it know to format those but not anything in `text`? Seems like I’m missing something here.

Tests/SwiftFormatTests/PrettyPrint/AccessorTests.swift
```diff
@@ -81,6 +81,82 @@ final class AccessorTests: PrettyPrintTestCase {
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 45)
   }
 
+  func testBasicAccessorsWithSelections() {
```
ahoppen, line 84: For all these test cases: Unless you’re explicitly testing that we handle multiple selection ranges, I would split them up into separate test cases. In my experience that makes it a lot easier to debug if one of the tests fail + you know what exactly the test is testing.

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
ahoppen: FWIW GitHub on Safari shows these arrows as boxes for me and GitHub on Google Chrome on macOS shows the left arrow as a box for me.

---

I would very much like to use some emoji that is rendered using a box because that way they really stand out to humans while looking at the source `«…»` fades too much into the source code for my taste. What do you think about ⏩️ and ⏪️ like we use in swift-syntax here: https://github.com/apple/swift-syntax/blob/55c4e4669c031d697e1924022b8ba250cfde0f2f/Sources/_SwiftSyntaxTestSupport/IncrementalParseTestUtils.swift#L182-L184

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
```diff
@@ -0,0 +1,231 @@
...
+      ➡️// do stuff⬅️
+      }
+      }
+      """
+
+    let expected =
+      """
+      func foo() {
+      if let SomeReallyLongVar = Some.More.Stuff(), let a = myfunc() {
+          // do stuff
```
ahoppen, line 42: Hmm, this is quite unfortunate. I would have preferred if we inferred the indentation from the unformatted context instead of from how the indentation should be if formatting was enabled.

FWIW running `clang-format` for lines 3-4 on 
```cpp
class Foo {
void test() {
int i = 0;
int j = 1;
}
}
```

produces

```cpp
class Foo {
void test() {
  int i = 0;
  int j = 1;
}
}
```

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
```diff
@@ -0,0 +1,231 @@
...
+      if let SomeReallyLongVar = Some.More.Stuff(), let a = myfunc() {
+      ➡️⬅️// do stuff
+      // do more stuff
+      ➡️⬅️var i = 0
+      }
+      }
+      """
+
+    // FIXME: this shouldn't indent the second comment
+    //    Also not idempotent since the selection doesn't get updated.
```
ahoppen, line 215: Just want to leave a comment here that we should fix this and enable the test before merging the PR. It’s too easy to forget about otherwise.

Sources/_SwiftFormatTestSupport/MarkedText.swift
```diff
@@ -18,23 +21,37 @@ public struct MarkedText {
   /// The text with all markers removed.
   public let textWithoutMarkers: String
 
+  /// If the marked text contains "➡️" and "⬅️", they're used to create a selection
+  public var selection: Selection?
+
```
ahoppen, line 26: Since all the markers are represented as `Int` that signify UTF-8 offsets, I think it would make sense to also store the ranges as `Range<Int>`.

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,85 @@
...
+    return ranges.contains { return $0.offset <= position && position < $0.end }
+  }
+
+  public func overlaps(_ range: Range) -> Bool {
+    return ranges.contains {
+      let maxStart = max($0.offset, range.offset)
+      let minEnd = (min($0.end, range.end))
+      return maxStart <= minEnd
+    }
+  }
```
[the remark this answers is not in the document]
DaveEwing: Ah. Nice!

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
DaveEwing: Yeah, I find it unfortunate too. The engine here isn't really set up to start "in the middle" like this. But perhaps we can move it in that direction down the road.

Sources/SwiftFormat/API/SwiftFormatter.swift
```diff
@@ -122,32 +128,26 @@ public final class SwiftFormatter {
   ///
   /// - Parameters:
   ///   - syntax: The Swift syntax tree to be converted to source code and formatted.
+  ///   - source: The original Swift source code used to build the syntax tree.
```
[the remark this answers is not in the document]
DaveEwing: Does the syntax tree hold on to the source (I didn't see that when I did this two months ago 🙃, but I can look...)

Sources/SwiftFormat/Core/Context.swift
```diff
@@ -39,6 +39,9 @@ public final class Context {
   /// The configuration for this run of the pipeline, provided by a configuration JSON file.
   let configuration: Configuration
 
+  /// The optional ranges to process
```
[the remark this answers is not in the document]
DaveEwing: `Selection` is now an `enum`, has an `infinite` case, and is never used as an `Optional`, so this should be less confusing.

DaveEwing:
Pushed changes that respond to _most_ of the comments from @allevato and @ahoppen. (I need to take another pass through the comments to see what I've missed!) I think the changes really are improvements, so thanks you two! I also need to do more testing locally... probably adding more tests - and getting the one disabled test working.

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import SwiftSyntax
+
+/// The selection as given on the command line - an array of offets and lengths
```
DaveEwing, line 16: And I need to take a pass through and update comments....

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,85 @@
...
+public extension Syntax {
+  /// return true if the node is completely inside any range in the selection (or if there's
+  /// no selection)
+  func isInsideSelection(_ selection: Selection?) -> Bool {
+    // no selection means we should process everything
+    guard let selection else { return true }
+
+    return selection.ranges.contains { range in
+      return range.offset <= position && endPosition <= range.end
+    }
```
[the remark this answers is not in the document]
DaveEwing: Actually, I can't. This test is to see if the node is _completely_ covered by an individual range. That is, we only want to apply rules for nodes that are completely inside a range.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
DaveEwing: Well, we have to make sure our state variables are set correctly so that it does that correctly. That the next few lines. 🙂

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import Foundation
+import SwiftSyntax
+
+/// The selection as given on the command line - an array of offets and lengths
+public struct Selection {
+  // TODO: make this a typealias instead
+  public struct Range: Decodable {
+    public let offset: AbsolutePosition
+    public let length: Int
```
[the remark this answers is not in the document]
DaveEwing: Done. Nice that this matches your refactoring in swift-syntax. When first implementing all this there wasn't a good idiom to follow. Now there is. And it's much cleaner. 🙂

Sources/SwiftFormat/API/Selection.swift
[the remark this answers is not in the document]
DaveEwing: As I said elsewhere, this is _so_ much nicer!

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
DaveEwing: Hmm... maybe. Lemme see if I can come up with a scenario where it'll matter.

Sources/SwiftFormat/PrettyPrint/Token.swift
DaveEwing: > Design question
I'd certainly prefer that these stand alone. They don't seem like "sub-enums" to me.

Tests/SwiftFormatTests/PrettyPrint/AccessorTests.swift
DaveEwing: I can move these test cases into `SelectionTests.swift` (which I've renamed locally from `Selections.swift`).

Tests/SwiftFormatTests/PrettyPrint/Selections.swift
DaveEwing: Fixed. The only issue was with idempotency. The utf8 offset in a selection range isn't valid for the formatted text. If I add support for line ranges then we can check for idempotency for formatting that won't do any line wrapping (or unwrapping).

Sources/_SwiftFormatTestSupport/MarkedText.swift
DaveEwing: I'm now using `Range<Int>` as they're built up, but create a `Selection` in the initializer. I may want to revisit this (and not have a `selection` property) if we support line ranges.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
DaveEwing: While this PR is pretty close to being ready, this is still an outstanding question....

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -4010,6 +4067,12 @@ extension Syntax {
 /// For example, comments after binary operators are relocated to be before the operator, which
 /// results in fewer line breaks with the comment closer to the relevant tokens.
 class CommentMovingRewriter: SyntaxRewriter {
+  init(selection: Selection = .infinite) {
+    self.selection = selection
+  }
+
+  var selection: Selection
```
ahoppen, line 4074: Could this be a `private let`?

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
ahoppen: I’m still missing how the whitespace characters that we’re shaving off with this are being printed.

If you have the following

```swift
func foo()   ⏩{}⏪
```

Then I think this only adds `func foo()` to the output stream, right (no whitespace)? But the next token that we print is the `{`, which doesn’t have any leading trivia (spaces are trailing trivia to `)`).

If those examples work because of some trivia re-attribution rule, how about the following?

```swift
func foo() /**/  ⏩{}⏪
```

```swift
func foo() /**/
  ⏩{}⏪
```

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -109,3 +124,18 @@ struct LintFormatOptions: ParsableArguments {
...
+      let pair = $0.components(separatedBy: ":")
+      if pair.count == 2, let start = Int(pair[0]), let end = Int(pair[1]), start <= end {
+        return start ..< end
+      } else {
+        return nil
+      }
+    }
+    self = ranges
+  }
+}
```
ahoppen, line 141: `@retroactive` won’t compile with a Swift 5.10 compiler and swift-format should be able to build with older compilers. The pattern I usually went with, is


```suggestion

extension [Range<Int>] {
  public init?(argument: String) {
    let pairs = argument.components(separatedBy: ",")
    let ranges: [Range<Int>] = pairs.compactMap {
      let pair = $0.components(separatedBy: ":")
      if pair.count == 2, let start = Int(pair[0]), let end = Int(pair[1]), start <= end {
        return start ..< end
      } else {
        return nil
      }
    }
    self = ranges
  }
}

#if compiler(>=6)
extension : @retroactive ExpressibleByArgument {}
#else
extension : ExpressibleByArgument {}
#endif
```

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -94,6 +105,10 @@ struct LintFormatOptions: ParsableArguments {
       throw ValidationError("'--assume-filename' is only valid when reading from stdin")
     }
 
+    if offsets?.isEmpty == false && paths.count > 1 {
```
ahoppen, line 108: It might be me but I always need to think extra hard with the optional to boolean comparisons and prefer coalescing the `nil` value.

```suggestion
    if !(offsets?.isEmpty ?? true) && paths.count > 1 {
```

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -26,6 +26,17 @@ struct LintFormatOptions: ParsableArguments {
       """)
   var configuration: String?
 
+  /// A list of comma-separated "start:end" pairs specifying UTF-8 offsets of the ranges to format.
+  ///
+  /// If not specified, the whole file will be formatted.
+  @Option(
+    name: .long,
+    help: """
+      A list of comma-separated "start:end" pairs specifying UTF-8 offsets of the ranges to format.
```
ahoppen, line 35: Instead of having a comma-separated list, wouldn’t it be nicer if you could specify `--range` or something multiple times to format multiple ranges? Seems like a more idiomatic CLI to me.

Tests/SwiftFormatTests/PrettyPrint/SelectionTests.swift
```diff
@@ -0,0 +1,359 @@
+import SwiftFormat
```
ahoppen, line 1: Could you add a copyright header?

DaveEwing: I can. But it doesn't look like any of the test files have copyright headers. Should that just be fixed all at once (in a separate PR)?

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,63 @@
...
+import Foundation
+import SwiftSyntax
+
+/// The selection as given on the command line - an array of offets and lengths
+public enum Selection {
+  case infinite
+  case ranges([Range<AbsolutePosition>])
+
+  /// Create a selection from an array of utf8 ranges. An empty array means an infinite selection.
+  public init(offsetPairs: [Range<Int>]) {
```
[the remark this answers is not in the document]
DaveEwing: Part of this change was in direct response to you suggesting I use `Range<Int>` here. But I probably misunderstood what you were suggesting. 🤷‍♂️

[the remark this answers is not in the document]
DaveEwing: > I think offsetRanges would be a better name

👍

Sources/SwiftFormat/API/SwiftFormatter.swift
```diff
@@ -21,6 +21,9 @@ public final class SwiftFormatter {
   /// The configuration settings that control the formatter's behavior.
   public let configuration: Configuration
 
+  /// the ranges of text to format
+  public var selection: Selection = .infinite
```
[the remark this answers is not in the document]
DaveEwing: Actually, the property can be deleted.

Tests/SwiftFormatTests/PrettyPrint/SelectionTests.swift
allevato: From what I remember, at the time we started swift-format, I think only Sources had copyright headers but not Tests in other projects, so I followed that pattern.

I'm fine with changing it, but I'd rather we do it all at once.

Sources/SwiftFormat/API/Selection.swift
```diff
@@ -0,0 +1,63 @@
...
+
+
+public extension Syntax {
+  /// return true if the node is _completely_ inside any range in the selection
+  func isInsideSelection(_ selection: Selection) -> Bool {
+    switch selection {
+    case .infinite:
+      return true
+    case .ranges(let ranges):
+      return ranges.contains { return $0.lowerBound <= position && endPosition <= $0.upperBound }
```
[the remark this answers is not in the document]
DaveEwing: This API is only used for full formatting rules which could actually affect the leading/trailing whitespace. Formatting line 2 here works just fine. (It's worth adding some tests around this sort of thing.)

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -26,6 +26,17 @@ struct LintFormatOptions: ParsableArguments {
...
+      """)
+  var offsets: [Range<Int>]?
```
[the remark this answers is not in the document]
allevato: FWIW, `clang-format` also supports this (with a **much** worse command line interface, alternating `--offset N --length N` pairs). I don't have a strong use case for it, but I can see it being useful for tooling that can easily get UTF-8 offsets without converting them to line/column numbers.

I do think the better and more common API would be `--lines=<start-line>-<end-line>`, though. I don't know how much we should care about supporting columns with that syntax, though.

Tests/SwiftFormatTests/PrettyPrint/SelectionTests.swift
ahoppen: Oh, sorry, then I missed that. I was just assuming that swift-format test files have copyright headers like all the other repos I work in but doesn’t seem to be the case. I would suggest that we
- Don’t add the copyright header in this PR, to be consistent
- Create some other PR to add copyright headers to all test files

Sources/swift-format/Subcommands/LintFormatOptions.swift
DaveEwing: I find the extra `!(...)` to be pretty obtuse. The idiom I'm using seems far more common (and readable) to me. 🙃

ahoppen: OK, then it’s just me 😉

Sources/swift-format/Subcommands/LintFormatOptions.swift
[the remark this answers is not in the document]
DaveEwing: Yeah. I basically think there are two interesting use cases (and I don't think I've said this very clearly):
- Editors that call out to `swift-format`. They want to give precise locations that can be given by line/col or offset. How they specify the location really doesn't matter, so getting hung up on it doesn't make sense to me. Users are almost never going to want to give locations this way on the command line.
- Scripts that want to process simple line ranges (ala. git-clang-format). Here they really want to use lines, and not offsets (and not columns!).

There's a third use case which I think is theoretical - Users typing at the command line. If a user were to do this, I think it's highly unlikely that they'd want to specify columns. (Clearly, a user wouldn't ever want to give a character offset!)

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -66,6 +67,19 @@ public class PrettyPrinter {
...
+  /// keep track of where formatting was disabled in the original source
+  ///
+  /// To format a selection, we insert `enableFormatting`/`disableFormatting` tokens into the
+  /// stream when entering/exiting a selection range. Those tokens include utf8 offsets into the
+  /// original source. When enabling formatting, we copy the text between `disabledPosition` and the
+  /// current position to `outputBuffer`. From then on, we continue to format until the next
+  /// `disableFormatting` token.
+  private var disabledPosition: AbsolutePosition? = nil
+  /// true if we're currently formatting
+  private var writingIsEnabled: Bool { disabledPosition == nil }
```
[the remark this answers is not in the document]
DaveEwing: I've moved the check inside `writeRaw()`, as you suggested, so I think having this second property is less useful and I removed it too.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2719,18 +2735,40 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
     return .skipChildren
   }
 
+  private func generateEnableFormattingIfNecessary(_ range: Range<AbsolutePosition>) {
+    if case .infinite = selection { return }
+    if !isInsideSelection && selection.overlapsOrTouches(range) {
+      appendToken(.enableFormatting(range.lowerBound))
+      isInsideSelection = true
+    }
+  }
```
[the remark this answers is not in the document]
DaveEwing: There really is different logic here (I mentioned this in a different comment the other day). `Syntax.isInsideSelection` requires the whole node and all its trivia be inside a range. Here, we're looking at sub-ranges in a node.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2719,18 +2735,40 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
     appendTrailingTrivia(token)
     appendAfterTokensAndTrailingComments(token)
 
     // It doesn't matter what we return here, tokens do not have children.
     return .skipChildren
   }
 
+  private func generateEnableFormattingIfNecessary(_ range: Range<AbsolutePosition>) {
+    if case .infinite = selection { return }
+    if !isInsideSelection && selection.overlapsOrTouches(range) {
```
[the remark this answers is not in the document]
DaveEwing: Nothing happens to the `func` token by itself if the range touches the end of it. However, if the range is `func   ⏩foo() {}⏪`, then you _do_ want the leading whitespace in front of `foo` to get canonicalized. Anyhow, that's my reasoning. 🤷‍♂️

Sources/swift-format/Subcommands/LintFormatOptions.swift
allevato: FWIW, I also find the `== false` version easier to read. 🙂

Sources/swift-format/Subcommands/LintFormatOptions.swift
allevato: I _think_ for array-typed `@Option`s, the [default behavior](https://swiftpackageindex.com/apple/swift-argument-parser/1.4.0/documentation/argumentparser/option/init(name:parsing:help:completion:)-238hg) should already be to append multiple occurrences of the flag into an array.

But since this property is actually a `[Range<Int>]?`, the optional bit might be interfering with the API. Can we drop the optional? We shouldn't need to distinguish between `nil` and an empty array here.

allevato: This also means the `ExpressibleByArgument` part below should probably be updated so that just `Range<Int>` conforms to it, instead of `[Range<Int>]`. Then if we make this property non-optional `[Range<Int>]`, everything else will fall out from there.

DaveEwing: Ah. I hadn't thought of the optional bit being what was getting in the way....

DaveEwing: And it works fine. So simple.

Sources/swift-format/Subcommands/LintFormatOptions.swift
DaveEwing: And this goes away with `offsets` no longer being optional. 😀

Sources/swift-format/Subcommands/LintFormatOptions.swift
DaveEwing: Gah! (Now that I pushed the change.) No, it doesn't work fine. This cause _at least one_ `--offsets` argument to be required, which we definitely don't want. (This is what I was hitting when I went down this road before.) But this is the main remaining issue on the PR, so I'll see if I can figure out a solution....

allevato: You might need to assign the property an explicit default value: `var offsets: [Range<Int>] = []`

DaveEwing: Thanks. Man, I hate programming in C++. Oh... wait.... 🙃

Sources/SwiftFormat/API/SwiftFormatter.swift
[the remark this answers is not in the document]
DaveEwing: I don't like having to regenerate the original source text if we already have it to start (which, of course, we do). And eventually we'll have a transformed syntax tree, and we still need to use the original source along with that. I think passing both makes the most sense.

Sources/SwiftFormat/API/SwiftLinter.swift
```diff
@@ -119,17 +119,18 @@ public final class SwiftLinter {
   /// - Throws: If an unrecoverable error occurs when formatting the code.
   public func lint(
     syntax: SourceFileSyntax,
+    source: String,
```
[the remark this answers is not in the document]
DaveEwing: Same comment here. Since the whitespace linting uses `PrettyPrinter` to compare to the original source, we need to pass it along too.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
DaveEwing: This aspect of the PR isn't working correctly yet. We're basically disabling all rewriting rules when there's a selection. 🙁 Looking into this now.

Sources/swift-format/Subcommands/LintFormatOptions.swift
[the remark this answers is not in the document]
ahoppen: OK, let’ start with `--offset` and we can add `line:colum`-based ranges later.

Sources/SwiftFormat/API/Selection.swift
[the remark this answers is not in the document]
ahoppen: Ah, I see 👍🏽

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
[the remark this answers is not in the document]
ahoppen: That’s a fair argument 👍🏽

ahoppen:
Looking through my comments again, I think there are 3 open comments (noting them here as much for my own sake as everybody else’s)

- [ ] Start in disabled state and explicitly enable formatting in the full-document case: https://github.com/apple/swift-format/pull/708#discussion_r1619262678
- [ ] Enable formatting for all trivia: https://github.com/apple/swift-format/pull/708#discussion_r1619286099
- [ ] I still don’t understand how whitespace that is being stripped will be printed: https://github.com/apple/swift-format/pull/708#discussion_r1619296810

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -68,17 +69,32 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
     super.init(viewMode: .all)
   }
 
   func makeStream(from node: Syntax) -> [Token] {
+    // if we have a selection, then we start outside of it
+    if case .ranges = selection {
+      appendToken(.disableFormatting(AbsolutePosition(utf8Offset: 0)))
+      isInsideSelection = false
+    }
+
```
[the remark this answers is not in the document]
DaveEwing: What's so different about having the pretty printer start in that state from having the token generator set the state? And actually I like that the token generator is solely responsible for managing the state. Moving part of that functionality to the pretty printer doesn't seem like an improvement.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3252,13 +3297,17 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
         appendToken(.comment(Comment(kind: .docLine, text: text), wasEndOfLine: false))
+        generateDisableFormattingIfNecessary(position + piece.sourceLength)
         appendNewlines(.soft)
         isStartOfFile = false
         requiresNextNewline = true
 
       case .docBlockComment(let text):
+        generateEnableFormattingIfNecessary(position ..< position + piece.sourceLength)
         appendToken(.comment(Comment(kind: .docBlock, text: text), wasEndOfLine: false))
+        generateDisableFormattingIfNecessary(position + piece.sourceLength)
```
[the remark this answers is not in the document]
DaveEwing: The idea is that if a range touches any part of some whitespace, then we should format that whitespace in its entirety. Alternatively, you could think of formatting as only moving "visible" things around (which is how I think about it). And if we allow formatting to be turned on and off in the middle of whitespace (and newlines), then we'd have to be much smarter about how we're handling whitespace generation. (This last bit probably applies to your other question about how we're even managing the whitespace. Doing it this way makes it all much easier.)

DaveEwing:
github won't let me comment on [_I’m still missing how the whitespace characters that we’re shaving off with this are being printed_](https://github.com/apple/swift-format/pull/708#discussion_r1619296810), so I'll do it here. As I alluded to [here](https://github.com/apple/swift-format/pull/708#discussion_r1628205066), we strip off trailing whitespace when we copy the original code so that we can let the normal mechanisms do all the normal whitespace generation work its magic instead. This is actually pretty simple. It's a matter of setting just a couple of state variables. This is PrettyPrint.swift, lines 610-617.

DaveEwing:
And one additional comment. Currently all the rules that do "non-whitespace" formatting are disabled when there's a selection. For instance, formatting won't remove semicolons when there's a selection. I think that's fine for a first pass, but I want to get that working too.

My intention, and what the code is currently trying to do but is failing, was to only enable processing of nodes that are fully-contained inside individual ranges. (By "failing", I mean that it just ends up disabling all rules.) Instead, I'm now looking at enabling the rules all the time, and using the existing mechanism to turn printing on/off. That's not working yet because the source ranges of the rewritten nodes no longer match the original source. I've good some work to do to figure out how to get it right....

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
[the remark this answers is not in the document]
ahoppen: My thinking is that adding `--offset 0:<end of file>` will work differently internally than if you don’t pass that where I would expect them to behave the same way. But maybe that’s just me.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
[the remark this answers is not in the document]
ahoppen: Thanks, that’s a great explanation and makes total sense! Might be worth putting in the source code for future reference.

ahoppen:
> github won't let me comment on [_I’m still missing how the whitespace characters that we’re shaving off with this are being printed_](https://github.com/apple/swift-format/pull/708#discussion_r1619296810), so I'll do it here. As I alluded to [here](https://github.com/apple/swift-format/pull/708#discussion_r1628205066), we strip off trailing whitespace when we copy the original code so that we can let the normal mechanisms do all the whitespace generation magic instead. This is actually pretty simple. It's a matter of setting just a couple of state variables. This is PrettyPrint.swift, lines 610-617.

OK, I think I understand what was throwing me off now

I thought that the space and tab trivia from the syntax tree were also added to the token stream but they are not.


Also, you assume that `CharacterSet.whitespace` is equivalent to the `.spaces`, and `.tabs`, `.formfeeds` and `.verticalTabs` cases in `Trivia` right? But that’s not generally true. Two examples:
- [NO-BREAK SPACE](https://unicode-explorer.com/c/00A0) is considered as whitespace in `CharacterSet.whitespace` but it’s parsed as unexpected text in swift-syntax (same for other characters from [Unicode General Category Zs](https://www.compart.com/en/unicode/category/Zs))
- If you have a line comment with a trailing space, we consider that space as part of the comment and not as separate whitespace in the trivia.
Because of this, when the formatted selection ends after a line comment with a trailing whitespace, you would remove it while whole file formatting would not trim it, I think. Similarly, if the formatted selection ends after a non-breaking space, you would remove it while whole file formatting would not.

I think we have two solutions here:
1. Inspect the syntax tree to find characters that swift-syntax has classified as spaces or tabs
2. Accept that there is a discrepancy between whole-file formatting and range-based formatting here. If we do that, I think we should
  - Add a comment explaining the trade-off being made
  - Only trim space, tab, formfeed and vertical tab instead of all characters in `CharacterSet.whitespaces` to align this to swift-syntax as closely as possible.

DaveEwing:
> Also, you assume that CharacterSet.whitespace is equivalent to the .spaces, and .tabs, .formfeeds and .verticalTabs cases in Trivia right? But that’s not generally true. Two examples:
>
> [NO-BREAK SPACE](https://unicode-explorer.com/c/00A0) is considered as whitespace in CharacterSet.whitespace but it’s parsed as unexpected text in swift-syntax (same for other characters from [Unicode General Category Zs](https://www.compart.com/en/unicode/category/Zs))
> If you have a line comment with a trailing space, we consider that space as part of the comment and not as separate whitespace in the trivia.
> Because of this, when the formatted selection ends after a line comment with a trailing whitespace, you would remove it while whole file formatting would not trim it, I think. Similarly, if the formatted selection ends after a non-breaking space, you would remove it while whole file formatting would not.
> I think we have two solutions here:
>
> Inspect the syntax tree to find characters that swift-syntax has classified as spaces or tabs
> Accept that there is a discrepancy between whole-file formatting and range-based formatting here. If we do that, I think we should
> a. Add a comment explaining the trade-off being made
> b. Only trim space, tab, formfeed and vertical tab instead of all characters in CharacterSet.whitespaces to align this to swift-syntax as closely as possible.

Once again, github doesn't let me comment directly. Weird.

In any case, adding a non-breaking space to your source causes swift-format to fail completely. It just outputs an empty file. This is true even before the changes in this PR. I haven't bothered looking into what it's doing.

I've tried cases where a comment ends with a space, or when block comments have a trailing space. I haven't found any differences between the behavior with or without a selection. The spaces are always removed. (And the behavior matches before the behavior before this PR.)

While there might be some edge cases here, they don't seem important to me.

ahoppen (approved):
I agree I am very pedantic here. A lot of it was me trying to understand how things fit together to see if there are any more significant issues lurking somewhere. But seems like there aren’t. Let’s get this merged and we can always fix it later if it’s causing any issues in practice (which I agree would probably be very niece anyway).

allevato:
> Currently all the rules that do "non-whitespace" formatting are disabled when there's a selection. For instance, formatting won't remove semicolons when there's a selection. I think that's fine for a first pass, but I want to get that working too.

@DaveEwing Were you still working on this, or did you want to land this change and then come back to it? I'm fine either way (since it's not immediately obvious how deep the change will be make that work).

I'll go ahead and kick off CI again, and if it comes back clean and you're happy with the current state of the PR, we can get it merged.

allevato (approved):
CI is green, so let me know if you'd like this merged now!

DaveEwing:
> > Currently all the rules that do "non-whitespace" formatting are disabled when there's a selection. For instance, formatting won't remove semicolons when there's a selection. I think that's fine for a first pass, but I want to get that working too.
>
> @DaveEwing Were you still working on this, or did you want to land this change and then come back to it? I'm fine either way (since it's not immediately obvious how deep the change will be make that work).

Part of me would like to clean this up first (maybe not fix it, but just make it more obvious in the code, and to users, that the rules are disabled). But I'd gladly do that in a second PR.

If you're up for it, let's merge, and I'll continue with polish over the next few weeks/months.

allevato:
I'll go ahead and merge then. Since this feature is only on `main`, we have time to iterate on it. Thanks for working on this—it's something I've wanted for a long time!

Sources/swift-format/Subcommands/LintFormatOptions.swift
xedin: @DaveEwing Alex pointed me to this conversion because I'm trying to implement `git swift-format`, we do need `--lines` for that to work because that's the only easily extractable information from `git diff-index`.

PR #710, opened 2024-03-26
Fix comment for multiElementCollectionTrailingCommas

ozumin:
According to the test case below, I thought
```swift
let MyCollection = [1, 2,]
```
would always be
```swift
let MyCollection = [1, 2]
```
regardless of `multiElementCollectionTrailingCommas` configuration.
https://github.com/apple/swift-format/blob/5545cf668ad61840162a51de97e4bc16ad666844/Tests/SwiftFormatTests/PrettyPrint/CommaTests.swift#L108-L144

Please close this PR if I misunderstood the issue 🙏


allevato (approved):
Thanks for fixing that oversight!

PR #729, opened 2024-05-01
Make sure there is a break after an #endif.

### LINKED ISSUE #705: Incorrect formatting for arguments following postfix `#if`
For e.g:

```swift
print(
    ""
    #if true
        .count
    #endif
    , 0
)
```

swift-format produces:

```swift
print(
  ""
    #if true
      .count
    #endif, 0
)
```

which the compiler rejects since the comma may not appear on the same line as the `#endif`.

rdar://124557891

shawnhyam:
You wouldn't normally allow a line break just before the comma in a parameter list, but if the preceding line is an #endif, we have to be sure to include the break or we'll generate invalid code. Fix #705.


allevato:
Wow, I guess I've never used postfix-`#if` in a comma-delimited list before; having to put a newline before the comma is super unfortunate 😞  Thanks for the fix!

PR #732, opened 2024-05-07
Merge adjacent .line and .docLine comments into a single element.

shawnhyam:
There is existing code to do this (for .docLine comments) but it isn't triggering due to the soft breaks in the token stream between every comment. This is a necessary first step for formatting line comments.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -3386,12 +3386,31 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   private func appendToken(_ token: Token) {
     if let last = tokens.last {
       switch (last, token) {
-      case (.comment(let c1, _), .comment(let c2, _))
-      where c1.kind == .docLine && c2.kind == .docLine:
-        var newComment = c1
-        newComment.addText(c2.text)
-        tokens[tokens.count - 1] = .comment(newComment, wasEndOfLine: false)
-        return
+      case (.break(.same, _, .soft(let count, _)), .comment(let c2, _))
+        where count == 1 && (c2.kind == .docLine || c2.kind == .line):
+        // we are search for the pattern of [line comment] - [soft break 1] - [line comment]
+        // where the comment type is the same; these can be merged into a single comment
+        if let nextToLast = tokens.dropLast().last,
+          case let .comment(c1, false) = nextToLast, 
+          c1.kind == c2.kind
+        {
+          var mergedComment = c1
+          mergedComment.addText(c2.text)
+          tokens.removeLast()  // remove the soft break
+          // replace the original comment with the merged one
+          tokens[tokens.count - 1] = .comment(mergedComment, wasEndOfLine: false)
+
+          // need to fix lastBreakIndex because we just removed the last break
+          lastBreakIndex = tokens.lastIndex(where: {
+            switch $0 {
+            case .break: return true
+            default: return false
+            }
+          })
+          canMergeNewlinesIntoLastBreak = false
+
+          return
+        }
 
       // If we see a pair of spaces where one or both are flexible, combine them into a new token
       // with the maximum of their counts.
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3386,12 +3386,31 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
   private func appendToken(_ token: Token) {
     if let last = tokens.last {
       switch (last, token) {
-      case (.comment(let c1, _), .comment(let c2, _))
-      where c1.kind == .docLine && c2.kind == .docLine:
-        var newComment = c1
-        newComment.addText(c2.text)
-        tokens[tokens.count - 1] = .comment(newComment, wasEndOfLine: false)
-        return
+      case (.break(.same, _, .soft(let count, _)), .comment(let c2, _))
```
allevato, line 3389: Since you're not using `count` anywhere else in the body of the statement, it would be simpler to put it directly in the pattern instead of in the `where` clause:

```swift
case (.break(.same, _, .soft(1, _)), .comment(let c2, _))
where c2.kind == .docLine || c2.kind == .line:
```

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3386,12 +3386,31 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+        where count == 1 && (c2.kind == .docLine || c2.kind == .line):
+        // we are search for the pattern of [line comment] - [soft break 1] - [line comment]
```
allevato, line 3391: Should we limit ourselves to just `.same` type breaks? Here's a weird example:

```swift
for x: Int
  // foo
  // bar
in y {}
```

This produces the following token stream around the comments:

```
  [BREAK Kind: continue Size: 1 Length: 107 NL: soft(count: 1, discretionary: true) Idx: 10]
  [COMMENT Line Length: 7 EOL: false Idx: 11]
  // foo
  [BREAK Kind: continue Size: 0 Length: 107 NL: soft(count: 1, discretionary: true) Idx: 12]
  [COMMENT Line Length: 7 EOL: false Idx: 13]
  // bar
  [BREAK Kind: continue Size: 0 Length: 104 NL: soft(count: 1, discretionary: true) Idx: 14]
```

This uses `.continue` breaks, so the logic as written here wouldn't merge those, even though visually they look like a multiline comment.

We'd also want to distinguish that from a similar case:

```
for x: Int  // foo
  // bar
in y {}
```

Presumably we wouldn't want to merge those comments together. I think the logic you have here is fine for that one, because the token stream looks like this:

```
  [SPACE Size: 2 Flexible: true Length: 2 Idx: 9]
  [COMMENT Line Length: 7 EOL: true Idx: 10]
  // foo
  [BREAK Kind: close(mustBreak: false) Size: 0 Length: 0 NL: elective(ignoresDiscretionary: false) Idx: 11]
  [BREAK Kind: continue Size: 1 Length: 107 NL: soft(count: 1, discretionary: true) Idx: 12]
  [COMMENT Line Length: 7 EOL: false Idx: 13]
  // bar
  [BREAK Kind: continue Size: 0 Length: 104 NL: soft(count: 1, discretionary: true) Idx: 14]
```

There are two breaks after `foo` (the first closes the indenting break group around the type annotation, the second is for the newline), so this logic (correctly) wouldn't trigger.

allevato (commented):
I like this generally as a way to move us toward being able to reflow comments as a unit (I saw your earlier draft PR). In addition to the comments below, can you please add some more tests? Comments are probably an area where we're a little bit weaker on test coverage, and it would be good to make sure we don't break anything obvious with this change. You don't have to test that the comments themselves got merged in the data model (since you don't have direct access to the token stream outside of the pretty printer), but tests verifying that the final indentation and line breaks are what you expect would be fine.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
shawnhyam: I changed it to allow merging of `.continue` and `.contextual` breaks, based on your feedback and examining the token stream from an existing codebase.

allevato (approved):
Thanks for working on this!

PR #738, opened 2024-05-10
Early implementation of formatting comments with Markdown.

shawnhyam:
@allevato I wanted to check if this is roughly what you had in mind for comment formatting. Maybe you'd be interested in playing with it a bit. I'm very interested in hearing your thoughts on how you see a feature like this working! A few random thoughts so far:

- Most of the time I really like what it does; and a minority of the time, I really really hate what it does.
- The automatic wrapping and list numbering/formatting is really pleasant.
- Commenting out a block of code will result in that block of code being reflowed and completely mangled.
- Ending lines with `\` to force line breaks gets re-rendered as double space `  `, which my editor trims and everything ends up getting reformatted anyway.
- It seems too aggressive for `.line` comments, but aside from the two issues above it could just be an issue with applying it to an existing codebase. Possibly a separate flag, or maybe individual comment sections could have a way to opt in/out?
- There are also block comments; should they be reformatted? If so, is there a canonical form and/or should this be an option?
- Should end-of-line comments be excluded? They are more difficult to format nicely.
- I noticed some of my projects would really benefit from a different width specification for comments.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration+Default.swift
+++ b/Sources/SwiftFormat/API/Configuration+Default.swift
@@ -39,5 +39,6 @@ extension Configuration {
     self.noAssignmentInExpressions = NoAssignmentInExpressionsConfiguration()
     self.multiElementCollectionTrailingCommas = true
     self.wrapComments = false
+    self.maximumCommentTextWidth = 72
   }
 }
--- a/Sources/SwiftFormat/API/Configuration+Default.swift
+++ b/Sources/SwiftFormat/API/Configuration+Default.swift
@@ -38,5 +38,6 @@ extension Configuration {
     self.spacesAroundRangeFormationOperators = false
     self.noAssignmentInExpressions = NoAssignmentInExpressionsConfiguration()
     self.multiElementCollectionTrailingCommas = true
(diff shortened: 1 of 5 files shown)


allevato:
> Interesting idea, although it might not make much difference in the end. Worth playing around with though.

Did you have something different in mind? Curious what your thoughts were here.

> They get canonicalized to a single form. Lists get a -, emphasis gets *, etc.

That works for me! I was worried lists might get something like `*`, but I guess since Apple prefers `-` for their doc tags, that's what they chose to render lists with.

allevato:
What was the status of this? It was still in Draft mode the last time I checked, so I wasn't sure it was ready for deeper review yet.

shawnhyam:
@allevato I actually wanted to have a discussion to nail down some of the desired behavior around the edge cases of both doc and regular comments. Things like block comments (with and without ASCII art) and indentation level changes, mixtures of different comment types, etc. What is the best place to discuss topics like this?

allevato:
I think this PR would be fine to discuss those, especially if the discussion needs to happen in the context of the changes you're making in the PR.

PR #746, opened 2024-05-28
Handle indented block comments with ASCII art correctly.

shawnhyam:
ASCII art was not being properly identified and stripped out if the block comments were indented. I fixed this and updated some test cases to have indented comments.


allevato:
Sorry for the delay on getting back to this. Thanks for the fix!

Issue #455, opened 2022-12-11
Lots of undocumented configuration options


dabrahams:
The very first one in the `dump-configuration` output, `fileScopedDeclarationPrivacy`, doesn't appear in the [configuration documentation](https://github.com/apple/swift-format/blob/main/Documentation/Configuration.md), and it has what _looks_ like a pretty drastic effect, and doesn't accept what I'm guessing is the most obvious setting `internal`.

allevato:
That option controls which visibility modifier to use (`private` or `fileprivate`) for declarations already using one of those at file-scope. That is, `private` and/or `fileprivate` declarations will be forced to be one or the other. This is to force consistency where—with the exception of `extension` decls, which are excluded from this—the two have the same effective visibility.

`internal` isn't supported since that would change the effective visibility of file-scoped decls.

allevato:
Re: the larger configuration issue, we should just adopt DocC like I want to do for the rules so that the source code is the source of truth and we generate everything from that.

ahoppen:
Tracked in Apple’s issue tracker as rdar://126948300

PR #747, opened 2024-06-01
Create jsonschema for configuration file

zidhuss:
This PR adds a JSON Schema that describes the `.swift-format` configuration file.

When this schema is referenced, It serves as a form of documentation and provides auto-completion and validation for configuration options (in editors that support it).

SCR-20240601-rhbx

Ideally you'd want to generate this from the code as done in #615. However the documentation for the other configuration options isn't automated from the comments leading to out of date documentation (#455).

My aim is to get this schema in the [schema store](https://www.schemastore.org/json/) so that editors automatically associate `.swift-format` files with it.

### DIFF
--- a/Documentation/configuration_schema.json
+++ b/Documentation/configuration_schema.json
@@ -1,0 +1,396 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Swift Format Configuration",
+  "description": "Configuration\nhttps://github.com/apple/swift-format/blob/main/Documentation/Configuration.md#swift-format-configuration",
+  "type": "object",
+  "properties": {
+    "version": {
+      "title": "version",
+      "type": "number",
+      "description": "The version of the configuration file. For now, this should always be 1.",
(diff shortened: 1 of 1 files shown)


ahoppen (commented):
Sorry for the late review @zidhuss. Adding a reference for the `.swift-format` file is great! 

I assume you wrote the schema by hand, right? I think it would be great if we had a script to make sure it’s in sync with Configuration.md. Maybe we could even generate Configuraiton.md based on the JSON schema. Could you be interested in writing such as script (ideally in Swift but really any scripting language would be fine)?

allevato:
It would be great to generate this from the doc comments in the `Configuration` struct (and the other structs that it contains) like you mentioned was done for rules in https://github.com/swiftlang/swift-format/pull/615, and also generate the Markdown from that (or alternatively, just abandon Configuration.md in favor of DocC comments). Otherwise, if we just check in a hand-written version of this file, it will go stale the same way the Markdown has.

PR #749, opened 2024-06-05
Configuration files in config directories (dotfiles, etc)

WuerfelDev:
Searches for configuration files (`swift-format/config.json`) in the config directories (`~/Library/Application Support`, `$XDG_CONFIG_HOME`, `.config`).
That enables a "global" (userwide) configuration that is stored eg in the dotfiles.

The configuration file is only used when no other configuration file was found

### DIFF
--- a/Documentation/Configuration.md
+++ b/Documentation/Configuration.md
@@ -5,7 +5,7 @@ used as a command line tool or as an API.
 
 ## Command Line Configuration
 
-A `swift-format` configuration file is a JSON file with the following
+A `.swift-format` configuration file is a JSON file with the following
 top-level keys and values:
 
 *   `version` _(number)_: The version of the configuration file. For now, this
@@ -92,7 +92,7 @@ top-level keys and values:
 
 An example `.swift-format` configuration file is shown below.
 
-```javascript
+```json
 {
     "version": 1,
     "lineLength": 100,
@@ -117,6 +117,18 @@ You can also run this command to see the list of rules in the default
 `swift-format` configuration:
 
     $ swift-format dump-configuration
+
+## Global Configuration
+
+If no `.swift-format` can be found for the current project/file, the configuration directories
+are searched for a `swift-format/config.json` file. While the filename is different,the
+configuration format stays the same.
+
+Locations that are searched, in this order:
+
+- `~/Library/Application Support/swift-format/config.json`
+- `$XDG_CONFIG_HOME/swift-format/config.json`
+- `~/.config/swift-format/config.json`
 
 ## API Configuration
 
--- a/Sources/swift-format/Frontend/Frontend.swift
+++ b/Sources/swift-format/Frontend/Frontend.swift
@@ -181,7 +181,9 @@ class Frontend {
   ///   it was provided, or by searching in paths inferred by `swiftFilePath` if one exists, or the
   ///   default configuration otherwise. If an error occurred when reading the configuration, a
   ///   diagnostic is emitted and `nil` is returned. If neither `pathOrString` nor `swiftFilePath`
-  ///   were provided, a default `Configuration()` will be returned.
+  ///   were provided, a configuration is searched at the current working directory or upwards the
+  ///   path. Next the configuration is searched for at the OS default config locations as
+  ///   swift-format/config.json. Finally the default `Configuration()` will be returned.
   private func configuration(
     fromPathOrString pathOrString: String?,
     orInferredFromSwiftFileAt swiftFileURL: URL?
@@ -238,6 +240,57 @@ class Frontend {
         diagnosticsEngine.emitError(
           "Unable to read configuration for \(cwd): \(error.localizedDescription)")
         return nil
+      }
+    }
+
+    // Load global configuration file
+    // First URLs are created, then they are queried. First match is loaded
+    var configLocations: [URL] = []
+
+    if #available(macOS 13.0, iOS 16.0, *) {
+      // From "~/Library/Application Support/" directory
+      configLocations.append(URL.applicationSupportDirectory)
+      // From $XDG_CONFIG_HOME directory
+      if let xdgConfig: String = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
+        configLocations.append(URL(filePath: xdgConfig, directoryHint: .isDirectory))
+      }
+      // From "~/.config/" directory
+      var dotconfig: URL = URL.homeDirectory
+      dotconfig.append(component: ".config", directoryHint: .isDirectory)
+      configLocations.append(dotconfig)
+    } else {
+      // From "~/Library/Application Support/" directory
+      var appSupport: URL = FileManager.default.homeDirectoryForCurrentUser
+      appSupport.appendPathComponent("Library", isDirectory: true)
+      appSupport.appendPathComponent("Application Support", isDirectory: true)
+      configLocations.append(appSupport)
+      // From $XDG_CONFIG_HOME directory
+      if let xdgConfig: String = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
+        configLocations.append(URL(fileURLWithPath: xdgConfig))
+      }
+      // From "~/.config/" directory
+      var dotconfig: URL = FileManager.default.homeDirectoryForCurrentUser
+      dotconfig.appendPathComponent(".config")
+      configLocations.append(dotconfig)
+    }
+
+    for var location: URL in configLocations {
+      if #available(macOS 13.0, iOS 16.0, *) {
+        location.append(components: "swift-format", "config.json")
+      } else {
+        location.appendPathComponent("swift-format", isDirectory: true)
+        location.appendPathComponent("config.json", isDirectory: false)
+      }
+      if FileManager.default.fileExists(atPath: location.path) {
+        do {
+          let configuration = try configurationLoader.configuration(at: location)
+          self.checkForUnrecognizedRules(in: configuration)
+          return configuration
+        } catch {
+          diagnosticsEngine.emitError(
+            "Unable to read configuration for \(location.path): \(error.localizedDescription)")
+          return nil
+        }
       }
     }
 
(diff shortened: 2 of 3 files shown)


Documentation/Configuration.md
```diff
@@ -118,6 +118,18 @@ You can also run this command to see the list of rules in the default
...
+
+If no `.swift-format` can be found for the current project/file, the configuration directories
+are searched for a `swift-format/config.json` file. While the filename is different,the
+configuration format stays the same.
+
+Locations that are searched, in this order:
+
+- `~/Library/Application Support/swift-format/config.json`
+- `$XDG_CONFIG_HOME/swift-format/config.json`
+- `~/.config/swift-format/config.json`
```
calebhearth, line 131: I'd love to see $XDG_CONFIG_HOME / ~/.config first here. What's the thinking behind looking first in ~/Library/Application Support?

WuerfelDev: It is the default path for configuration files on MacOS

ahoppen (commented):
Thanks for the PR, @calebhearth. I think adding an ability to have a global `.swift-format` configuration without putting `.swift-format` in `/` will be useful.

A couple comments regarding the search locations:
- I think we should also support `XDG_CONFIG_DIRS` for a system-wide configuration file isntead of just a user-wide configuration file.
- We should also offer default search locations on Windows, the [Build Server Protocol](https://build-server-protocol.github.io/docs/overview/server-discovery#default-locations-for-bsp-connection-files) uses `%LOCALAPPDATA%\bsp\` and `%PROGRAMDATA%\bsp\` and I think we could do the same here.
- The search path discovery for BSP servers is implemented in SourceKit-LSP at https://github.com/swiftlang/sourcekit-lsp/blob/4ca25836226a2b28868a42df3a3336fcc481e1c5/Sources/BuildSystemIntegration/ExternalBuildSystemAdapter.swift#L195-L220. I think you should be able to steal a lot of that (especially note that there are methods in Foundation to get the path for the `Application Support` directories). Also, I think it makes sense to follow that precedent for ranking the XDG-dirs vs `Application Support`.

calebhearth:
@WuerfelDev opened the PR, but I second the thanks!

ahoppen:
Oh, whoops, sorry. @WuerfelDev thank you for opening the PR 🙏🏽

WuerfelDev:
I will make the changes within the next week. However I am unable to test on windows

ahoppen:
> I will make the changes within the next week. However I am unable to test on windows

That’s fine. Just try and do the best you can think of, leaving FIXMEs for things that you know need covering and I can do the rest.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+
+    #if os(Windows)
+      if let localAppData = ProcessInfo.processInfo.environment["LOCALAPPDATA"] {
+        configLocations.append(URL(fileURLWithPath: localAppData))
+      }
+      if let programData = ProcessInfo.processInfo.environment["PROGRAMDATA"] {
+        configLocations.append(URL(fileURLWithPath: programData))
+      }
+    #else
+      if let xdgConfigHome = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
```
WuerfelDev, line 258: I'm a bit unhappy because `$XDG_CONFIG_HOME` is not defined by default on Mac. Do you think we could additionally check for `$HOME/.config/`

calebhearth: The XDG spec suggests falling back to $HOME/.config if $XDG_CONFIG_HOME is unset so I think that's a good idea too

ahoppen: I don’t see any problem with that and don’t think we need to search `$HOME/.config`. On macOS, you can place your config files in `Application Support`.

> The XDG spec suggests falling back to $HOME/.config if $XDG_CONFIG_HOME is unset so I think that's a good idea too. 

Do you have a reference where the XDG spec says this?

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+        configLocations.append(URL(fileURLWithPath: xdgConfigHome))
+      }
+
+      if let libraryUrl = FileManager.default.urls(
+        for: .applicationSupportDirectory, in: .userDomainMask
+      ).first {
+        configLocations.append(libraryUrl)
+      }
```
ahoppen, line 266: Not sure why this would happen but I’d add all directories returned from `urls(for:in:)` to `configLocations`.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+      if let libraryUrl = FileManager.default.urls(
+        for: .applicationSupportDirectory, in: .systemDomainMask
+      ).first {
+        configLocations.append(libraryUrl)
+      }
+    #endif
+
+    for case var location? in configLocations {
+      if #available(macOS 13.0, iOS 16.0, *) {
+        location.append(components: "swift-format", "config.json")
```
ahoppen, line 283: This will never be true on non-Apple platforms. I would prefer to just have one branch here that uses `appendPathComponent`. We do that in many other places as well. Just has lower risk of platforms diverging.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+      } else {
+        location.appendPathComponent("swift-format", isDirectory: true)
+        location.appendPathComponent("config.json", isDirectory: false)
+      }
+      if FileManager.default.fileExists(atPath: location.path) {
+        do {
+          let configuration = try configurationLoader.configuration(at: location)
+          self.checkForUnrecognizedRules(in: configuration)
```
ahoppen, line 291: What do we need this check for?

Sources/swift-format/Frontend/Frontend.swift
calebhearth: https://specifications.freedesktop.org/basedir-spec/latest/#variables

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

ahoppen: Oh, great catch. I think we should follow that advice. @WuerfelDev, could you add that to your PR?

Sources/swift-format/Frontend/Frontend.swift
WuerfelDev: I was a confused about how language features availability can be checked in swift. In the documentation they are listed as available/deprecated by OS and Xcode version but no mention of a swift version which would be arguably better for usage in non-apple environment.
I'll change it back to `appendPathComponent`. Was trying to avoid it if possible because of the note in the [documentation](https://developer.apple.com/documentation/foundation/url/1780239-appendingpathcomponent).

Sources/swift-format/Frontend/Frontend.swift
WuerfelDev: The check was implemented to print diagnostics when unknown rules are used (https://github.com/swiftlang/swift-format/pull/612). It is checked every time that we load a configuration file

Sources/swift-format/Frontend/Frontend.swift
ahoppen: I think we might be able to switch to `append(components:)` if we consistently use [swift-foundation](https://github.com/swiftlang/swift-foundation) instead of [swift-corelibs-foundation](https://github.com/swiftlang/swift-corelibs-foundation) on non-Apple platforms but we don’t right now because we need to support building with Swift 5.10, which doesn’t have the new swift-foundation yet.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+      }
+      if let programData = ProcessInfo.processInfo.environment["PROGRAMDATA"] {
+        configLocations.append(URL(fileURLWithPath: programData))
+      }
+    #else
+      if let xdgConfigHome = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
+        configLocations.append(URL(fileURLWithPath: xdgConfigHome))
+      }
+
+      if let homeLocation = ProcessInfo.processInfo.environment["HOME"] {
```
ahoppen, line 262: We should only look at `$HOME/.config` if `XDG_CONFIG_HOME` is not set.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -241,6 +243,61 @@ class Frontend {
...
+        configLocations.append(URL(fileURLWithPath: homeLocation + "/.config/"))
```
ahoppen, line 263: Could you use `appendingPathComponent` instead of concatenating strings to form paths?

Sources/swift-format/Frontend/Frontend.swift
ahoppen: Ah, I see. 👍🏽

ahoppen (commented):
A couple more comments. Also, could you format your changes using swift-format? Otherwise CI will fail with that issue.

WuerfelDev:
@ahoppen Thank you so much for guiding me through this feature implementation, being new to swift I've learned a bunch of things along the way
I hope everything is fine now, else lmk if more changes are required

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -268,6 +270,64 @@ class Frontend {
       }
     }
 
+    // Load global configuration file
+    // First URLs are created, then they are queried. First match is loaded
+    var configLocations: [URL?] = []
```
ahoppen, line 275: Sorry, I only now noticed: Do we ever add `nil` to this array? I don’t think so and we wouldn’t need to do `case var location?` in the loop below if we don’t allow optionals in here.

PR #750, opened 2024-06-14
Fix `@_expose` attribute argument spacing

kateinoigakukun:
Pretty-print the `@_expose` with the correct spacing. It was formatted as `@_expose(wasm,"foo")` instead of `@_expose(wasm, "foo")`.


allevato (approved):
Thanks! Looks good but there's something (unrelated to this PR) currently broken on Linux CI, so I'll need to wait for that to get resolved first.

allevato:
Would you mind cherry-picking this into the `release/6.0` branch as well?

ahoppen:
I’m merging `main` into `release/6.0` here's o a cherry-pick is not needed. https://github.com/apple/swift-format/pull/751

PR #758, opened 2024-06-25
[6.0] Update links for repositories moved to the swiftlang org on GitHub

ahoppen:


CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ endif()
 find_package(SwiftSyntax CONFIG)
 if(NOT SwiftSyntax_FOUND)
   FetchContent_Declare(Syntax
-    GIT_REPOSITORY https://github.com/apple/swift-syntax
+    GIT_REPOSITORY https://github.com/swiftlang/swift-syntax
     GIT_TAG main)
```
allevato, line 72: Unrelated to this PR, but I just noticed that CMake is referring to the `main` branch here. Does that need to be fixed on the 6.0 branch?

allevato (approved):

CMakeLists.txt
bnbarham: Heh, probably. It wouldn't be used in CI at all. @compnerd what if instead of specifying repository + tag, we just specify a path that's passed down and then rely on update-checkout? It makes it a little more clunky to eg. just build the repo itself, but is maybe better than having yet another place to update versions?

ahoppen: I updated the branch to `release/6.0` for now. But yes, would be nice to not have yet another place that needs version updates.

PR #759, opened 2024-06-25
Split the PrettyPrint class into two pieces.

shawnhyam:
This change breaks up the PrettyPrint class into one piece that is responsible for tracking the state of the pretty printing algorithm - processing the tokens, tracking the break stack, comma delimited region stack, etc.; and another piece that is responsible for assembling the output and tracking the state of the output - current line, column position, and indentation.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -76,12 +76,13 @@ public class PrettyPrinter {
   /// original source. When enabling formatting, we copy the text between `disabledPosition` and the
   /// current position to `outputBuffer`. From then on, we continue to format until the next
   /// `disableFormatting` token.
-  private var disabledPosition: AbsolutePosition? = nil
+  private var disabledPosition: AbsolutePosition? = nil {
+    didSet {
+      outputBuffer.isEnabled = disabledPosition == nil
+    }
+  }
 
-  private var outputBuffer: String = ""
-
-  /// The number of spaces remaining on the current line.
-  private var spaceRemaining: Int
+  private var outputBuffer: PrettyPrintBuffer
 
   /// Keep track of the token lengths.
   private var lengths = [Int]()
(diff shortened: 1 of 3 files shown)


Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -607,21 +528,19 @@ public class PrettyPrinter {
...
-      }
+      outputBuffer.writeVerbatimAfterEnablingFormatting(text)
 
     case .disableFormatting(let newPosition):
       assert(disabledPosition == nil)
       disabledPosition = newPosition
     }
   }
 
+  private func shouldBreak(_ length: Int) -> Bool {
```
allevato, line 539: I wonder if this argument should have a default value of 1? Then the cases above where we're checking if we're already at or past the end of the line can just write `shouldBreak()` instead of `shouldBreak(1)`; the latter feels a little "magical".

Continuing down that line of thinking, maybe an even better name for this would be `canFit`. Then the call sites would read like `canFit(length)` where we're looking for a specific count, or `canFit()` for "can anything fit".

WDYT?

Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
```diff
@@ -0,0 +1,140 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+struct PrettyPrintBuffer {
```
allevato, line 15: Please add a brief doc comment for this type and for the stored properties below.

allevato (commented):
Interesting idea! I think this looks really good; the algorithm has gotten more complex over time as we've needed to add new features to the original, and I like that this moves the much of the "number fudging" (i.e., direct math operations) down to a lower layer and making the core algorithm read a bit more semantically.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
shawnhyam: Makes total sense to me, I've updated with these changes.

Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
shawnhyam: Done!

allevato (approved):

PR #763, opened 2024-06-27
Add PrettyPrintBuffer.swift to CMakeLists.txt

ahoppen:


allevato:
Odd, shouldn't CI have caught this and failed the Windows build?

bnbarham:
Yeah, the issue was that the cross PR test used swiftlang and not apple, then didn't *actually* cross PR test. That should go away as a problem quite soon.

PR #765, opened 2024-07-09
Add lint warning for retroactive conformances

harlanhaskins:
This provides a lint rule whenever an `@retroactive` appears in a source file.

I've marked the rule as opt-out, but I'm also wondering if we have a syntax for opting out of a rule in general.

### DIFF
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -14,6 +14,7 @@ Here's the list of available rules:
 - [AlwaysUseLiteralForEmptyCollectionInit](#AlwaysUseLiteralForEmptyCollectionInit)
 - [AlwaysUseLowerCamelCase](#AlwaysUseLowerCamelCase)
 - [AmbiguousTrailingClosureOverload](#AmbiguousTrailingClosureOverload)
+- [AvoidRetroactiveConformances](#AvoidRetroactiveConformances)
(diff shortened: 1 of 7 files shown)


harlanhaskins:
@swift-ci please test

allevato:
> I've marked the rule as opt-out, but I'm also wondering if we have a syntax for opting out of a rule in general.

You can override the [`isOptIn`](https://github.com/swiftlang/swift-format/blob/6fadf647aec2595c84d0c9273f65e82ee33d3fba/Sources/SwiftFormat/Core/Rule.swift#L26) property from `Rule` if you want to make the rule disabled by default, and when you run generate-swift-format it will use flip the default in the generated configuration. Is that what you had in mind?

harlanhaskins:
Oh, I meant like a comment-like syntax for clients to silence the rule within a file, like

`// swift-format ignore: AvoidRetroactiveConformances`

did we ever get anything like that?

allevato:
Yeah, `// swift-format-ignore: <rulle name>` should work for any visitor-based rule. https://github.com/swiftlang/swift-format/blob/main/Documentation/IgnoringSource.md#ignoring-source-transforming-rules

allevato:
Can you please update the CMake files to include the new `AvoidRetroactiveConformances.swift` file? (Just the `Sources`, The `Tests` don't need it.)

allevato (approved):
Thanks!

PR #767, opened 2024-07-13
Add checks for test functions marked with @Test attribute in relevant Rules

hamtiko:
This pull request updates the `AlwaysUseLowerCamelCase` rule in the swift-format project to support the new [Swift Testing](https://github.com/apple/swift-testing) framework. 
Previously, the rule allowed underscores in test function names based on the old `XCTest` framework criteria. I’ve added a check to also allow underscores for functions marked with the `@Test` attribute.

P.S. I am planning to add check for new testing framework for `NeverForceUnwrap`,  `NeverUseForceTry`  and `NeverUseImplicitlyUnwrappedOptionals` rules as well.

### DIFF
--- a/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
+++ b/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
@@ -13,7 +13,8 @@
 import SwiftSyntax
 
 extension WithAttributesSyntax {
-  /// Indicates whether the node has attribute with the given `name`.
+  /// Indicates whether the node has attribute with the given `name` and `module`.
+  /// The `module` is only considered if the attribute is written as `@Module.Attribute`.
   ///
   /// - Parameter name: The name of the attribute to lookup.
   /// - Parameter module: The module name to lookup the attribute in.
@@ -22,10 +23,11 @@ extension WithAttributesSyntax {
     attributes.contains { attribute in
       let attributeName = attribute.as(AttributeSyntax.self)?.attributeName
       if let identifier = attributeName?.as(IdentifierTypeSyntax.self) {
+        // @Attribute syntax
         return identifier.name.text == name
       }
-      // support @Module.Attribute syntax as well
       if let memberType = attributeName?.as(MemberTypeSyntax.self) {
+        // @Module.Attribute syntax
         return memberType.name.text == name
           && memberType.baseType.as(IdentifierTypeSyntax.self)?.name.text == module
       }
--- a/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
+++ b/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
@@ -1,0 +1,28 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+extension WithAttributesSyntax {
+    /// Indicates whether the node has attribute with the given `name`.
+    ///
+    /// - Parameter name: The name of the attribute to lookup.
+    /// - Returns: True if the node has an attribute with the given `name`, otherwise false.
+    func hasAttribute(_ name: String) -> Bool {
+        attributes.contains { attribute in
+            let attributeName = attribute.as(AttributeSyntax.self)?.attributeName
+            return attributeName?.as(IdentifierTypeSyntax.self)?.name.text == name
+            // support @Module.Attribute syntax as well
+            || attributeName?.as(MemberTypeSyntax.self)?.name.text == name
+        }
+    }
+}
--- a/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
+++ b/Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
@@ -13,16 +13,23 @@
 import SwiftSyntax
 
 extension WithAttributesSyntax {
-    /// Indicates whether the node has attribute with the given `name`.
-    ///
-    /// - Parameter name: The name of the attribute to lookup.
-    /// - Returns: True if the node has an attribute with the given `name`, otherwise false.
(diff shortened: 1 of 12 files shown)


Sources/SwiftFormat/Rules/AlwaysUseLowerCamelCase.swift
```diff
@@ -100,7 +100,11 @@ public final class AlwaysUseLowerCamelCase: SyntaxLintRule {
 
     // We allow underscores in test names, because there's an existing convention of using
     // underscores to separate phrases in very detailed test names.
-    let allowUnderscores = testCaseFuncs.contains(node)
+    let allowUnderscores = testCaseFuncs.contains(node) || node.attributes.contains {
```
allevato, line 103: It would be great to factor this new check out into a convenience method on `WithAttributesSyntax`, which could then be used elsewhere (and on any declaration that can have attributes). What about something like this?

```
extension WithAttributesSyntax {
  func hasAttribute(_ name: String) { ... }
}
```

You could create a new `WithAttributesSyntax+Convenience.swift` file in Sources/SwiftFormat/Core for this (make sure to also add the new file to the CMakeLists.txt file in Sources/SwiftFormat).

If you're feeing really bold, you could also support the fully-qualified version: `@Testing.Test`.

allevato (commented):
> P.S. I am planning to add check for new testing framework for NeverForceUnwrap, NeverUseForceTry and NeverUseImplicitlyUnwrappedOptionals rules as well.

Would you be willing to do that in this PR? I don't think it would make it too big to review and it would be nice to have a single PR that adds both new swift-testing-related features.

Sources/SwiftFormat/Rules/AlwaysUseLowerCamelCase.swift
hamtiko: That's a good idea, will do it together with adding Swift Testing support for `NeverForceUnwrap`,  `NeverUseForceTry` and `NeverUseImplicitlyUnwrappedOptionals` rules to this PR. 
Thanks

hamtiko:
@allevato I have added the Swift Testing checks for `NeverForceUnwrap`, `NeverUseForceTry` and `NeverUseImplicitlyUnwrappedOptionals` rules to this PR and moved the attribute lookup logic to the separate extension as you suggested. Thanks

Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
```diff
@@ -0,0 +1,28 @@
...
+    /// Indicates whether the node has attribute with the given `name`.
+    ///
+    /// - Parameter name: The name of the attribute to lookup.
+    /// - Returns: True if the node has an attribute with the given `name`, otherwise false.
+    func hasAttribute(_ name: String) -> Bool {
+        attributes.contains { attribute in
+            let attributeName = attribute.as(AttributeSyntax.self)?.attributeName
+            return attributeName?.as(IdentifierTypeSyntax.self)?.name.text == name
+            // support @Module.Attribute syntax as well
+            || attributeName?.as(MemberTypeSyntax.self)?.name.text == name
```
allevato, line 25: Since you're not checking the module name here, this would match anything, right? Like `@Foo.Test`, for example.

It might be better to make the function signature `hasAttribute(_ name: String, inModule: String)` and have the caller always provide it, but it would only be checked if it was a `MemberTypeSyntax`. So `hasAttribute("Test", inModule: "Testing")` would still match either `@Test` or `@Testing.Test`, but not `@Foo.Test`.

Tests/SwiftFormatTests/Rules/AlwaysUseLowerCamelCaseTests.swift
```diff
@@ -210,4 +210,24 @@ final class AlwaysUseLowerCamelCaseTests: LintOrFormatRuleTestCase {
...
+      """
+      @Test
+      func function_With_Test_Attribute() {}
+      @Testing.Test("Description for test functions",
+            .tags(.testTag))
+      func function_With_Test_Attribute_And_Args() {}
+      func 1️⃣function_Without_Test_Attribute() {}
+      @objc
+      func 2️⃣function_With_Non_Test_Attribute() {}
+      """,
```
allevato, line 226: Once you make the change above, you can add another example like `@Foo.Test` and verify that it does emit a diagnosis.

allevato (commented):
Looks really good! Just one comment about the attribute checking.

Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
hamtiko: What you think if we will have 2 functions, one will be the current one  and another function like you suggested?

allevato: I think it's better for simplicity to have one function that does both checks. We can't _definitively_ know if `@Test` refers to `@Testing.Test` without the file being type-checked, which we can't do in swift-format. So we're making an assumption that `@Test` by itself refers to the one in `Testing` and nothing else, because of how common we expect that pattern to be. If we're handling that as the common case, then there's no reason to not also handle `@Testing.Test` with the same query—each of the individual rules shouldn't have to make two calls to find the attribute it's looking for.

Tests/SwiftFormatTests/Rules/AlwaysUseLowerCamelCaseTests.swift
hamtiko: Done.

Sources/SwiftFormat/Core/WithAttributesSyntax+Convenience.swift
```diff
@@ -13,16 +13,23 @@
...
-    /// - Parameter name: The name of the attribute to lookup.
-    /// - Returns: True if the node has an attribute with the given `name`, otherwise false.
-    func hasAttribute(_ name: String) -> Bool {
-        attributes.contains { attribute in
-            let attributeName = attribute.as(AttributeSyntax.self)?.attributeName
-            return attributeName?.as(IdentifierTypeSyntax.self)?.name.text == name
-            // support @Module.Attribute syntax as well
-            || attributeName?.as(MemberTypeSyntax.self)?.name.text == name
-        }
+  /// Indicates whether the node has attribute with the given `name`.
```
allevato, line 16: Can you mention in the doc comment that the module name is only considered if present if the attribute is written as `@Module.Name`, and that that's intentional? Otherwise it might look like a bug to someone reading it in the future.

hamtiko: good point, done. 👍

allevato (approved):
Thank you for adding this!

PR #769, opened 2024-07-15
Add a break after `do` when it has typed throws

### LINKED ISSUE #768: Typed throws in do blocks formatted incorrectly
```
struct bar: Error {
}

func foo() {
  do throws(bar) {
    throw bar()
  } catch {
  }
}
```

`do throws(bar)` becomes `dothrows(bar)`

bnbarham:
Pretty-printing of `do` statements with typed throws was skipping the space between the `do` and `throws`. Make sure to add it in.

Fixes #768.


allevato (approved):
Thanks!

PR #776, opened 2024-07-21
Add option to customize number of spaces leading `//` comments

dduan:
Add an option `spacesBeforeEndOfLineComments` that allows customization of number of spaces between the first non-trivial token and a line comment with the `//` prefix. Defaults to 2, the historical default value.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -67,7 +67,7 @@ public struct Configuration: Codable, Equatable {
   /// The maximum length of a line of source code, after which the formatter will break lines.
   public var lineLength: Int
 
-  // Number of spaces that preceeds line comments.
+  /// Number of spaces that precede line comments.
   public var spacesBeforeEndOfLineComments: Int
 
   /// The width of the horizontal tab in spaces.
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -28,6 +28,7 @@ public struct Configuration: Codable, Equatable {
     case version
     case maximumBlankLines
     case lineLength
+    case spacesBeforeEndOfLineComments
     case tabWidth
     case indentation
     case respectsExistingLineBreaks
@@ -65,6 +66,9 @@ public struct Configuration: Codable, Equatable {
 
   /// The maximum length of a line of source code, after which the formatter will break lines.
   public var lineLength: Int
+
+  // Number of spaces that preceeds line comments.
+  public var spacesBeforeEndOfLineComments: Int
 
   /// The width of the horizontal tab in spaces.
   ///
@@ -225,6 +229,9 @@ public struct Configuration: Codable, Equatable {
     self.lineLength =
       try container.decodeIfPresent(Int.self, forKey: .lineLength)
       ?? defaults.lineLength
+    self.spacesBeforeEndOfLineComments =
+      try container.decodeIfPresent(Int.self, forKey: .spacesBeforeEndOfLineComments)
+      ?? defaults.spacesBeforeEndOfLineComments
     self.tabWidth =
       try container.decodeIfPresent(Int.self, forKey: .tabWidth)
       ?? defaults.tabWidth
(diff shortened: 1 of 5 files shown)


Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -66,6 +67,9 @@ public struct Configuration: Codable, Equatable {
   /// The maximum length of a line of source code, after which the formatter will break lines.
   public var lineLength: Int
 
+  // Number of spaces that preceeds line comments.
+  public var spacesBeforeLineComments: Int
```
bnbarham, line 71: Does this apply to both `//` and `/**/` comments? The test case with `/**/` has 1-space in both the input and expected, so wasn't clear from there.

Tests/SwiftFormatTests/PrettyPrint/CommentTests.swift
```diff
@@ -199,6 +200,127 @@ final class CommentTests: PrettyPrintTestCase {
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 45)
   }
 
+  func testLineCommentsWithCustomLeadingSpaces() {
```
bnbarham, line 203: I think it would help to split this test case up - it's quite hard to compare the input/expectation at the moment.

Sources/SwiftFormat/API/Configuration.swift
dduan: I believe `/* */` is referred to as "block comments" regardless of whether it's on the same line or not. In existing unit tests (adjacent to the one I added), there are `/* */` comments with 1 leading space that aren't affected by the formatting (which ensures 2 spaces). I made sure in the documentation, it explicitly says "comments starting with '//'" to avoid this ambiguity.

Tests/SwiftFormatTests/PrettyPrint/CommentTests.swift
dduan: it's a copy of the previous tests. I can try and split of both?

bnbarham: I'm happy to just have the new tests split, feel free to also clean up the copied-from in another PR though.

Sources/SwiftFormat/API/Configuration.swift
bnbarham: Ah I see, thanks. Could you make the comment here a doc comment (`///`) and include that here as well? @allevato is it intended that block comments are treated differently in this case? Not sure I have much of an opinion on it either way.

allevato: The name of this configuration should be explicit that it's just about "**end of line** comments", since that traditionally means `//` comments that specifically follow something else on a line and does not include `/* ... */`, `///`, or `/** ... */` comments, nor `//` comments on their own line. So I'd call this `spacesBeforeEndOfLineComments`.

I think it's correct to only support this for `//` EoL comments. Let's imagine we have a situation where we want 2 spaces before those, but only one space around inline block comments, and we have this contrived example:

```swift
let x = 5 // something
let y = someValue + x /* meters */ / someOtherValue /* seconds */
```
When we add another space before the comment on the first line, I don't think it would be right to change the spacing around the block comments on the second line.

simanerush (commented):
Neat!

Tests/SwiftFormatTests/PrettyPrint/CommentTests.swift
dduan: split!

bnbarham: Personally I'd split them into functions so the failure is super obvious if it happens. Currently you'd have to map the failed input/output back to one of the pairs by looking over them all (since the failure would always be on the same line). You could also add a line number to each test case, but at that point having separate functions seems easier.

Any preference @allevato? I'm also happy to take as is, it's just tests.

Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -66,6 +67,9 @@ public struct Configuration: Codable, Equatable {
   /// The maximum length of a line of source code, after which the formatter will break lines.
   public var lineLength: Int
 
+  // Number of spaces that preceeds line comments.
```
bnbarham, line 70:
```suggestion
  /// Number of spaces that precede line comments.
```

bnbarham (commented):
Thanks for the updates!

dduan:
@allevato thoughts?

allevato (approved):
I'm surprised folks would have a strong enough opinion about this to want to change the default, but whatever 🙂

dduan:
Thanks @allevato! FWIW, I personally don't hold any opinions about any of the options. But getting swift-format to adoption for my team requires some finessing such as this :)

tscholze:
Does anyone know when this version of `swiftformat` will be shipped with Xcode? It seems that the current version of Xcode has not this feature included. Thanks!

ahoppen:
This change will be included in the swift-format bundled with Swift 6.1.

Issue #771, opened 2024-07-18
Ignoring a rule on a variable declaration doesn't work


FranzBusch:
I tried to ignore a rule on a variable declaration like this:

```
struct Bar {}
struct Foo {
  // swift-format-ignore: DontRepeatTypeInStaticProperties
  var fooBar: Bar 
```

However this didn't seem to work and I had to move the ignore to the struct declaration to work.

ahoppen:
Synced to Apple’s issue tracker as rdar://131991753

allevato:
To paraphrase our offline chat here: I suspect it's a bug in the way we visit the nodes. We visit the type/extension and then manually recurse into the member list to emit the violations instead of implementing `visit` methods directly for the members. That would bypass the `swift-format-ignore` logic. (The implementation of the rule preceded the implementation of `ignore`, and we never caught this.)

shawnhyam:
@allevato It is definitely the case that the issue is with how the nodes are visited. I was looking at converting the `DontRepeatTypeInStaticProperties` rule to use the visit function on `VariableDeclSyntax` nodes, which would require knowing the current type name in the visit method. I was going to use the existing `visit` methods for that purpose, but because of nested types it seems necessary to push/pop the current type name using `visit`/`visitPost` pairs. Unfortunately, although everything works fine in the `linter.walk(...)` flow in `LintOrFormatRuleTestCase`, the `visitPost` methods are not called in the `pipeline.lint(...)` flow. Is this known or expected behavior? Or is there a better way to get the name of the enclosing type from a syntax node?

shawnhyam:
Update: I changed the code generator to make sure `visitPost` gets called, but don't use that. I just climb the tree looking for a type name instead.

MahdiBM:
just bumped into this but the problem is the "struct declaration" is out of the package so we can't even suppress this warning?

```swift
@available(*, deprecated, message: "Deprecating conformance to `PostgresDataConvertible`, since it is deprecated.")
extension Data: PostgresDataConvertible {
    public static var postgresDataType: PostgresDataType {
        .bytea
    }
```

```
Sources/PostgresNIO/Data/PostgresData+Bytes.swift:26:23: warning: [DontRepeatTypeInStaticProperties] remove the suffix 'Data' from the name of the variable 'postgresDataType'
```

PR #777, opened 2024-07-23
Upgrade the DontRepeatTypeInStaticProperties rule.

shawnhyam:
- Using the `visit` method to examine all the variable declarations should allow specific decls to be ignored.
- Update the PipelineGenerator to call `visitPost` methods
- Modify the PipelineGenerator to produce output matching the current generated files.

### DIFF
--- a/Sources/SwiftFormat/Core/Pipelines+Generated.swift
+++ b/Sources/SwiftFormat/Core/Pipelines+Generated.swift
@@ -76,7 +76,6 @@ class LintPipeline: SyntaxVisitor {
     visitIfEnabled(AllPublicDeclarationsHaveDocumentation.visit, for: node)
     visitIfEnabled(AlwaysUseLowerCamelCase.visit, for: node)
     visitIfEnabled(BeginDocumentationCommentWithOneLineSummary.visit, for: node)
-    visitIfEnabled(DontRepeatTypeInStaticProperties.visit, for: node)
     visitIfEnabled(NoLeadingUnderscores.visit, for: node)
     visitIfEnabled(TypeNamesShouldBeCapitalized.visit, for: node)
     visitIfEnabled(UseTripleSlashForDocumentationComments.visit, for: node)
@@ -86,7 +85,6 @@ class LintPipeline: SyntaxVisitor {
     onVisitPost(rule: AllPublicDeclarationsHaveDocumentation.self, for: node)
     onVisitPost(rule: AlwaysUseLowerCamelCase.self, for: node)
     onVisitPost(rule: BeginDocumentationCommentWithOneLineSummary.self, for: node)
-    onVisitPost(rule: DontRepeatTypeInStaticProperties.self, for: node)
     onVisitPost(rule: NoLeadingUnderscores.self, for: node)
     onVisitPost(rule: TypeNamesShouldBeCapitalized.self, for: node)
     onVisitPost(rule: UseTripleSlashForDocumentationComments.self, for: node)
@@ -182,7 +180,6 @@ class LintPipeline: SyntaxVisitor {
 
   override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
     visitIfEnabled(BeginDocumentationCommentWithOneLineSummary.visit, for: node)
-    visitIfEnabled(DontRepeatTypeInStaticProperties.visit, for: node)
     visitIfEnabled(FullyIndirectEnum.visit, for: node)
     visitIfEnabled(NoLeadingUnderscores.visit, for: node)
     visitIfEnabled(OneCasePerLine.visit, for: node)
@@ -192,7 +189,6 @@ class LintPipeline: SyntaxVisitor {
   }
   override func visitPost(_ node: EnumDeclSyntax) {
     onVisitPost(rule: BeginDocumentationCommentWithOneLineSummary.self, for: node)
-    onVisitPost(rule: DontRepeatTypeInStaticProperties.self, for: node)
     onVisitPost(rule: FullyIndirectEnum.self, for: node)
     onVisitPost(rule: NoLeadingUnderscores.self, for: node)
     onVisitPost(rule: OneCasePerLine.self, for: node)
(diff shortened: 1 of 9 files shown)


shawnhyam:
Should be a partial fix for #771, since it upgrades one rule only. Interested in feedback on if there is a better way to do this before upgrading any other rules.

Sources/SwiftFormat/Core/Pipelines+Generated.swift
```diff
@@ -76,17 +76,16 @@ class LintPipeline: SyntaxVisitor {
...
     visitIfEnabled(BeginDocumentationCommentWithOneLineSummary.visit, for: node)
-    visitIfEnabled(DontRepeatTypeInStaticProperties.visit, for: node)
     visitIfEnabled(NoLeadingUnderscores.visit, for: node)
     visitIfEnabled(TypeNamesShouldBeCapitalized.visit, for: node)
     visitIfEnabled(UseTripleSlashForDocumentationComments.visit, for: node)
     return .visitChildren
   }
   override func visitPost(_ node: ClassDeclSyntax) {
     onVisitPost(rule: AllPublicDeclarationsHaveDocumentation.self, for: node)
+    visitPost(AlwaysUseLowerCamelCase.visitPost, for: node)
```
ahoppen, line 86: Is this an unrelated change? And should this be `onVisitPost`. I don’t actually know but `visitPost` sticks out to me here.

Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
```diff
@@ -110,3 +67,38 @@ extension Finding.Message {
     "remove the suffix '\(type)' from the name of the variable '\(name)'"
   }
 }
+
+extension Syntax {
+  /// Returns the name of the immediately enclosing type of this node if there is one,
+  /// otherwise nil.
+  fileprivate func typeName() -> String? {
+    switch Syntax(self).as(SyntaxEnum.self) {
+    case .classDecl(let node):
```
ahoppen, line 76: You should also check for `.actorDecl`.

Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
```diff
@@ -110,3 +67,38 @@ extension Finding.Message {
     "remove the suffix '\(type)' from the name of the variable '\(name)'"
   }
 }
+
+extension Syntax {
+  /// Returns the name of the immediately enclosing type of this node if there is one,
+  /// otherwise nil.
+  fileprivate func typeName() -> String? {
```
ahoppen, line 74: This could be a computed property. Seems more swifty to me.

```suggestion
  fileprivate var typeName: String? {
```

Sources/generate-swift-format/PipelineGenerator.swift
```diff
@@ -93,12 +94,20 @@ final class PipelineGenerator: FileGenerator {
         handle.write(
           """
             override func visitPost(_ node: \(nodeType)) {
+          
           """
         )
-        for ruleName in lintRules.sorted() {
+       for ruleName in lintRules {
```
ahoppen, line 100: Unintentionally removed an indentation whitespace here?

Sources/SwiftFormat/Core/Pipelines+Generated.swift
shawnhyam: My original implementation used `visitPost`, but there was a problem because the generated pipeline doesn't pass on `visitPost` calls to rules that override those methods. `onVisitPost` only does some cleanup of the pipeline. So the code would work as a standalone visitor, but not as part of the pipeline. I ended up changing the implementation to not use `visitPost`, but I did keep the fix to the code generator in.

ahoppen: OK, I’m not too familiar with how the entire pipeline works but I trust you on this 😉

ahoppen (approved):

ahoppen:
@shawnhyam Instead of merging `main` into your branch, could you rebase your branch? I find it creates a nicer git history.

shawnhyam:
@allevato I was wondering if you had any thoughts on this, especially the changes to the pipeline generator (so that `visitPost` gets called).

Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
```diff
@@ -110,3 +67,40 @@ extension Finding.Message {
     "remove the suffix '\(type)' from the name of the variable '\(name)'"
   }
 }
+
+extension Syntax {
+  /// Returns the name of the immediately enclosing type of this node if there is one,
+  /// otherwise nil.
+  fileprivate var typeName: String? {
```
allevato, line 74: The name of this property feels a little misleading; when I saw you accessing the `typeName` of a `VariableDeclSyntax` node above, it read to me like it was accessing the type of the *variable*. Can we call it `containingDeclName` so that the nature of the traversal is clear at the call site?

Sources/SwiftFormat/Rules/NeverForceUnwrap.swift
```diff
@@ -16,6 +16,7 @@ import SwiftSyntax
 ///
 /// This rule does not apply to test code, defined as code which:
 ///   * Contains the line `import XCTest`
+///   * The function is marked with `@Test` attribute
```
allevato, line 19: Is this noise from another PR that was merged in?

Sources/SwiftFormat/Core/Pipelines+Generated.swift
allevato: Can we drop this and the code generator changes then, if they're not being used?

Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
```diff
@@ -22,69 +22,26 @@ import SwiftSyntax
...
+      guard let identifierPattern = binding.pattern.as(IdentifierPatternSyntax.self) else {
+        continue
+      }
 
-        let varName = identifierPattern.identifier.text
-        if varName.contains(bareTypeName) {
-          diagnose(.removeTypeFromName(name: varName, type: bareTypeName), on: identifierPattern)
-        }
+      let varName = identifierPattern.identifier.text
+      if varName.contains(bareTypeName) {
```
allevato, line 39: Now that I'm getting another look at this, I wonder if we should update this to be `hasSuffix`, since the diagnostic explicitly talks about suffixes.

Sources/SwiftFormat/Rules/NeverForceUnwrap.swift
shawnhyam: > Is this noise from another PR that was merged in?

This was because I ran `generate-swift-format`, and there have been some manual edits to the generated files. So I made the code edits necessary to keep the generated file the same.

Sources/SwiftFormat/Core/Pipelines+Generated.swift
shawnhyam: @allevato I could split it out into another change, but I still want to put it in because AFAICT it's a bug in the pipeline generator (`visitPost` is never called) and I have other changes in progress that depend on that visitor method being called. Should I go ahead and split this into two separate changes?

Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
shawnhyam: Good question...  we wouldn't catch plurals (which we don't mention or have tests for), but there are likely false positives with the current logic too.
```
public actor Cookie {
  // cases that are caught with .contains but not with .hasSuffix
  static let chocolateChipCookies: Cookie
  static let rawCookieDough: Cookie
}
```

allevato:
Thanks for fixing this!

PR #784, opened 2024-07-28
Add option to add a newline between 2 adjacent attributes

### LINKED ISSUE #773: Rule for putting annotations into separate lines
We often put various annotations such as `@available` or `@inlinable` in separate lines before methods or types. A rule to enforce that formatting would be great.

```
@usableFromInline
@SomeMacro
struct Foo {
}
```

dduan:
Add an option that inserts hard line breaks between adjacent attributes.

Closes #773

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -34,6 +34,7 @@ public struct Configuration: Codable, Equatable {
     case lineBreakBeforeControlFlowKeywords
     case lineBreakBeforeEachArgument
     case lineBreakBeforeEachGenericRequirement
+    case lineBreakBetweenDeclarationAttributes
     case prioritizeKeepingFunctionOutputTogether
     case indentConditionalCompilationBlocks
     case lineBreakAroundMultilineExpressionChainComponents
@@ -110,6 +111,9 @@ public struct Configuration: Codable, Equatable {
   /// list to be laid out vertically. If false (the default), requirements will be laid out
   /// horizontally first, with line breaks only being fired when the line length would be exceeded.
   public var lineBreakBeforeEachGenericRequirement: Bool
+
+  /// If true, a line break will be added between adjacent attributes.
+  public var lineBreakBetweenDeclarationAttributes: Bool
 
   /// Determines if function-like declaration outputs should be prioritized to be together with the
   /// function signature right (closing) parenthesis.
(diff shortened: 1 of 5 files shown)


dduan:
converted draft to investigate whether extra spaces between attributes should be remove and what that means for the naming/implementation of this option.

allevato:
I think we would only want this behavior for _declaration_ attributes, because it would be heavy-handed on type attributes. For example, how does your current implementation handle this case?

```swift
func f(body: @MainActor @Sendable () -> Void) {}
```

We should definitely keep those horizontally flowing, IMO.

Can you add some tests for that case, and also some tests that use attributes in conditional `#if ...` blocks, just to make sure we don't break anything there?

dduan:
makes sense. I'll give it a go!

dduan:
@allevato I took your suggestion and expanded the unit test to cover a lot more node types.

Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -34,6 +34,7 @@ public struct Configuration: Codable, Equatable {
     case lineBreakBeforeControlFlowKeywords
     case lineBreakBeforeEachArgument
     case lineBreakBeforeEachGenericRequirement
+    case lineBreakBetweenAttributes
```
allevato, line 37: Can we update the name of this to be `lineBreakBetweenDeclarationAttributes` to match the new behavior?

Tests/SwiftFormatTests/PrettyPrint/AttributeTests.swift
```diff
@@ -468,4 +468,103 @@ final class AttributeTests: PrettyPrintTestCase {
 
     assertPrettyPrintEqual(input: input, expected: expected, linelength: 100)
   }
+
+  func testLineBreakBetweenAttributes() {
```
allevato, line 472: Can you add a test that includes attributes surrounded by `#if`?

I would expect

```swift
@foo @foo2
#if FLAG
  @bar @baz
#endif
@quux func f() {}
```

to become

```swift
@foo
@foo2
#if FLAG
  @bar
  @baz
#endif
@quux
func f() {}
```

dduan: yup

dduan:
@allevato thanks for the feedbacks! I've addressed them.

allevato:
Thanks!

PR #788, opened 2024-08-02
Modify assertLint test helper to run via pipeline only.

shawnhyam:
Standalone lint visitors don't check for swift-format-ignore markers, so any test cases that use those markers will return different results between standalone vs full pipeline.


allevato:
This makes me wonder if we even need the standalone-run mode anymore. Since we now double-check everything using the pipeline, it might be better to just have that be the single mode we care about, because the rule never runs in a standalone mode otherwise.

Could you make that change instead? Then test authors don't have to pick what mode they're interested in, which is cleaner.

shawnhyam:
Sure! From my point of view that's better -- I found it pretty confusing when I was getting different behavior between the two runs.

allevato (approved):

PR #791, opened 2024-08-08
Add support for re-indenting block comments.

shawnhyam:
Re-indent all lines of a block comment with the current indentation level. The comment must not have anything other than whitespace preceding it on the line, and all lines must have at least that level of whitespace (or be empty).

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/Comment.swift
+++ b/Sources/SwiftFormat/PrettyPrint/Comment.swift
@@ -58,9 +58,12 @@ struct Comment {
   let kind: Kind
   var text: [String]
   var length: Int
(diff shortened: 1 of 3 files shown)


allevato (approved):
Clearly my efforts to divert people away from using `/* ... */` comments by not formatting them well have been in vein 😛

thomasvl:
@shawnhyam - I think you might want to look at adding a test case with some blank lines in the multiline comments, i.e. –

```swift
        /* Comment

           Comment after blank */
```

```swift
        /*
           Comment

           Comment after blank
        */
```

It looks like this is inserting the leading whitespace on the blank lines rather than leaving the blank lines completely blank.

PR #792, opened 2024-08-13
Wrap multiline string literals to line length.

### LINKED ISSUE #724: Split string literals into multiple lines
Formatting a long string literal single line even though it exceeds the line limit. We should break it up into multiple lines, probably into a multi-line string literal

```swift
"some very long string literal that just goes on and on and one without having any real comment but it reaches past the end of the line limit"
```

Should become

```swift
"""
some very long string literal that just goes on and on and one without having \
any real comment but it reaches past the end of the line limit
"""
```

thunderseethe:
Modify the pretty printer to insert escaped newlines "\\\n" when a line would exceed configured line length. This only affects multi-line strings and not single-line strings. Previously string segements in multiline literals were treated as one large `.syntax` token. They are now broken up into multiple tokens to allow opportunities to line break.

A string segement `"the quick fox"` will now be comprised of the tokens:
```swift
[.syntax("the"), .syntax(" "), .break(.escaped), .syntax("quick"), .syntax(" "), .break(.escaped), .syntax("fox")]
```
we now break up the segment based on whitespace. After each series of whitespace, we insert the new `.escaped` break.
A couple of odd things to note here:
-  A new NewlineBehavior: `.escaped`
- Tokens include both the whitespace and the line break
- We emit whitespace as a `.syntax` and not a `.space`

`.escaped` is very similar to an `.elective` line break. It will only break when it exceed the line length, otherwise it will be printed as `size` spaces. Unlike `.elective`, when it is printed it will print as an escaped newline "\\\n". On the implementation side we also have to calculate it's length differently than we do for `.elective`.

Because all whitespace in a string literal is significant (unlike whitespace in other swift syntax), it is emitted alongside the break. This is because, unlike other breaks, we always want to print the whitespace in a string literal. The choice to print the whitespace prior to the line break is cosmetic.

Whitespace is emitted as a `.syntax` to avoid special casing between `.break`s and `.space`s. The pretty printer will ignore spaces (or change them) around line breaks to make code look better. However this doesn't work for string literals where the whitespace is significant. Emitting `.syntax` token ensures all whitespace in the literal shows up. 

For example, the literal:
```swift
"""
Lorem ipsum odor amet, consectetuer adipiscing elit. Vehicula aptent per convallis fermentum mus. Rutrum ultrices nisl etiam et volutpat dictum pulvinar elementum felis.
"""
```
will be formatted as:
```swift
"""
Lorem ipsum odor amet, consectetuer adipiscing \
elit. Vehicula aptent per convallis fermentum mus. \
Rutrum ultrices nisl etiam et volutpat dictum \
pulvinar elementum felis.
"""
```

Wrapping will reflow escaped newlines already present in the string literal. For example, with a line length of 30:

```swift
"""
multi-line \
string.
"""
```

will be formatted as:

```swift
"""
multi-line string.
"""
```

However hard line breaks are always respected, since they appear in the string literal itself not just the syntax. So:

```swift
"""
multi-line
string
"""
```

will not be reformatted.

Breaks are only inserted after whitespace, so words longer than line length will not be wrapped (line length 30):

```swift
"""
long word: 01914cc5-a274-7df4-b88b-5e167f3f5662
"""
```
will be formatted as
```swift
"""
long word: \
01914cc5-a274-7df4-b88b-5e167f3f5662
"""
```
even though the second line exceeds 30 characters.

Fixes #724

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -214,7 +214,6 @@ public class PrettyPrinter {
     switch token {
     case .contextualBreakingStart:
       activeBreakingContexts.append(ActiveBreakingContext(lineNumber: outputBuffer.lineNumber))
-
       // Discard the last finished breaking context to keep it from effecting breaks inside of the
       // new context. The discarded context has already either had an impact on the contextual break
       // after it or there was no relevant contextual break, so it's safe to discard.
@@ -414,7 +413,7 @@ public class PrettyPrinter {
 
       var overrideBreakingSuppressed = false
       switch newline {
-      case .elective: break
+      case .elective, .escaped: break
       case .soft(_, let discretionary):
         // A discretionary newline (i.e. from the source) should create a line break even if the
         // rules for breaking are disabled.
@@ -429,6 +428,10 @@ public class PrettyPrinter {
       let suppressBreaking = isBreakingSuppressed && !overrideBreakingSuppressed
       if !suppressBreaking && (!canFit(length) || mustBreak) {
         currentLineIsContinuation = isContinuationIfBreakFires
+        if case .escaped = newline {
+          outputBuffer.enqueueSpaces(size)
+          outputBuffer.write("\\")
+        }
         outputBuffer.writeNewlines(newline)
         lastBreak = true
       } else {
@@ -594,19 +597,51 @@ public class PrettyPrinter {
       // Break lengths are equal to its size plus the token or group following it. Calculate the
       // length of any prior break tokens.
       case .break(_, let size, let newline):
-        if let index = delimIndexStack.last, case .break = tokens[index] {
-          lengths[index] += total
+        if let index = delimIndexStack.last, case .break(_, _, let lastNewline) = tokens[index] {
+          /// If the last break and this break are both `.escaped` we add an extra 1 to the total for the last `.escaped` break.
+          /// This is to handle situations where adding the `\` for an escaped line break would put us over the line length.
+          /// For example, consider the token sequence:
+          /// `[.syntax("this fits"), .break(.escaped), .syntax("this fits in line length"), .break(.escaped)]`
+          /// The naive layout of these tokens will incorrectly print as:
+          ///  """
+          ///  this fits this fits in line length \
+          ///  """
+          ///  which will be too long because of the '\' character. Instead we have to print it as:
+          ///  """
+          ///  this fits \
+          ///  this fits in line length
+          ///  """
+          ///
+          /// While not prematurely inserting a line in situations where a hard line break is occurring, such as:
+          ///
+          /// `[.syntax("some text"), .break(.escaped), .syntax("this is exactly the right length"), .break(.hard)]`
+          ///
+          /// We want this to print as:
+          /// """
+          /// some text this is exactly the right length
+          /// """
+          /// and not:
+          /// """
+          /// some text \
+          /// this is exactly the right length
+          /// """
+          if case .escaped = newline, case .escaped = lastNewline {
+            lengths[index] += total + 1
+          } else {
+            lengths[index] += total
+          }
           delimIndexStack.removeLast()
         }
         lengths.append(-total)
         delimIndexStack.append(i)
 
-        if case .elective = newline {
-          total += size
-        } else {
-          // `size` is never used in this case, because the break always fires. Use `maxLineLength`
-          // to ensure enclosing groups are large enough to force preceding breaks to fire.
-          total += maxLineLength
+        switch newline {
+        case .elective, .escaped:
+            total += size
+        default:
+            // `size` is never used in this case, because the break always fires. Use `maxLineLength`
+            // to ensure enclosing groups are large enough to force preceding breaks to fire.
+            total += maxLineLength
         }
 
       // Space tokens have a length equal to its size.
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -15,7 +15,7 @@ import SwiftOperators
 import SwiftSyntax
 
 fileprivate extension AccessorBlockSyntax {
-  /// Assuming that the accessor only contains an implicit getter (i.e. no 
+  /// Assuming that the accessor only contains an implicit getter (i.e. no
   /// `get` or `set`), return the code block items in that getter.
   var getterCodeBlockItems: CodeBlockItemListSyntax {
     guard case .getter(let codeBlockItemList) = self.accessors else {
@@ -1437,9 +1437,9 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
   override func visit(_ node: ReturnClauseSyntax) -> SyntaxVisitorContinueKind {
     if node.parent?.is(FunctionTypeSyntax.self) ?? false {
-      // `FunctionTypeSyntax` used to not use `ReturnClauseSyntax` and had 
-      // slightly different formatting behavior than the normal 
-      // `ReturnClauseSyntax`. To maintain the previous formatting behavior, 
+      // `FunctionTypeSyntax` used to not use `ReturnClauseSyntax` and had
+      // slightly different formatting behavior than the normal
+      // `ReturnClauseSyntax`. To maintain the previous formatting behavior,
       // add a special case.
       before(node.arrow, tokens: .break)
       before(node.type.firstToken(viewMode: .sourceAccurate), tokens: .break)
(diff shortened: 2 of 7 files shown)


Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2462,37 +2462,80 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+      if let stringLiteralExpr = stringLiteralParent() {
         return pendingMultilineStringBreakKinds[stringLiteralExpr, default: .same]
       } else {
         return .same
       }
     }
 
+    let isMultiLineString = stringLiteralParent()?.openingQuote.tokenKind == .multilineStringQuote
+    // Replace each instance of consecutive whitespace in a string segment with an `.escaped` break.
+    func emitMultilineSegmentTextTokens(segment: Substring) {
```
allevato, line 2481: This is kind of a big function to nest locally. Can we move it outside as a helper method?

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2462,37 +2462,80 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+      var currentWord = [Character]()
```
allevato, line 2482: Can we build up `Unicode.Scalar`s instead of `Character`s here? That should produce the same result but without incurring the performance penalty of doing grapheme clustering.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -594,19 +597,51 @@ public class PrettyPrinter {
...
+          /// `[.syntax("some text"), .break(.escaped), .syntax("this is exactly the right length"), .break(.hard)]`
+          ///
+          /// We want this to print as:
+          /// """
+          /// some text this is exactly the right length
+          /// """
+          /// and not:
+          /// """
+          /// some text \
+          /// this is exactly the right length
```
allevato, line 626: Since backslashes are used to _suppress_ newlines, I _think_ I agree with this choice. I'll admit it's somewhat opinionated though. I don't know if other folks would want to be able to choose where in a string they break it. If we made this configurable, we could allow that, but the problem with doing that is that once a string is broken in this way, it can never be automatically unbroken, so shifting indentation could cause a string to split awkwardly.

@ahoppen @bnbarham Do either of you have opinions here?

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -2460,39 +2460,85 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
...
+    }
+
+    // Only one of these will actually do anything based on whether our last char was whitespace or not.
+    emitWord()
+    emitBreak()
+  }
+
   override func visit(_ node: StringSegmentSyntax) -> SyntaxVisitorContinueKind {
     // Looks up the correct break kind based on prior context.
+    func stringLiteralParent() -> StringLiteralExprSyntax? {
```
allevato, line 2504: This could just be a local variable rather than a nested function, I think?

Tests/SwiftFormatTests/PrettyPrint/ClosureExprTests.swift
```diff
@@ -251,12 +251,11 @@ final class ClosureExprTests: PrettyPrintTestCase {
       let a = funcCall() { [weak self] (a: Int) in
         return a + 1
       }
-      let a = funcCall() {
-        [weak self, weak a = self.b] (a: Int) in
+      let a = funcCall() { [weak self, weak a=self.b] (a: Int) in
```
allevato, line 254: I think you need to rebase your local checkout or make sure that you have the latest dependencies; these shouldn't change based on https://github.com/swiftlang/swift-format/pull/790 and the corresponding swift-syntax PR.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
allevato: Looking back at the code, I don't remember why it was a nested function to begin with. It doesn't look like it depends on any state that would change elsewhere during execution.

I think it would be fine to simplify it.

allevato (approved):
This looks reasonable to me. I'll go ahead and kick off CI, but let's wait for other folks to comment on it as well.

grynspan:
Something to take into consideration is that ASCII spaces between words are not universal across scripts. Most (all?) Latin-based scripts use them, but Chinese and Japanese writing rarely do, and some scripts use alternate characters (interpuncts, etc.) instead of ASCII spaces.

allevato:
> Something to take into consideration is that ASCII spaces between words are not universal across scripts. Most (all?) Latin-based scripts use them, but Chinese and Japanese writing rarely do, and some scripts use alternate characters (interpuncts, etc.) instead of ASCII spaces.

Good point. At a minimum, [the logic on this line](https://github.com/swiftlang/swift-format/pull/792/files#diff-c6c2e2a43fac11c6c1cda074a1e1b01693672cc219a2259994b2630c15e628dfR2479) should, at a minimum, preserve the original scalar detected by `isWhiteSpace`, rather than replacing it with a hardcoded ASCII space.

We could consider other kinds of characters to break on later; unfortunately, we didn't include any of the Unicode breaking properties in [`Unicode.Scalar.Properties`](https://developer.apple.com/documentation/swift/unicode/scalar/properties-swift.struct) from what I can tell.

shawnhyam:
> One idea would be to only reflow strings where **every** line ends in a `\`. That would likely exclude cases like the ones above, but it might also exclude a lot of valid cases (like two very long lines of English prose). I'm definitely open to better ideas.

The case of preformatted code (or any preformatted text) came to mind for me -- this is possibly one solution, although I'm not convinced I would want it to reformat these blocks at all. I guess if I was used to it, it might be fine. But my feeling is this should be controlled by an option so it could at least be turned off if it's creating problems.

allevato:
Code in multi-line strings is the use case that stands out as most compelling to address to me.

I'd love to see any examples that someone might have where they have non-code text that is split across multiple lines terminated by `\` where reflowing those lines arbitrarily would harm readability.

I guess one example would be other kinds of structured text:

```swift
let x = """
  value1: \(value1), \
  value2: \(value2), \
  value3: \(value3), \
  value4: \(value4)
  """
```
which could reflow like this, depending on line width:
```swift
let x = """
  value1: \(value1), value2: \(value2), value3: \
  \(value3), value4: \(value4)
  """
```
I think the argument could be made that the second form is worse.

It should be fairly straightforward to make something like `// swift-format-ignore: MultilineStrings` work, but it would be even _better_ if we can do the right thing in most cases without intervention.

grynspan:
Well now we're getting dangerously close to including swift-syntax as a dependency just so we can format code-in-block-strings. 🙃

allevato:
swift-evolution proposal to add descriptive labels to multiline strings, à la Markdown code fences?

```swift
let code = """swift
  func \(name)() -> \(result) {
  }
  """
```
🤔 

_Of course, my preference would be to do something good here without modifying the language_ 😅

grynspan:
I mean, why not? Pitch the triple-backtick syntax! _Do it!_

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
ahoppen: I personally would not automatically re-flow string literals. Most likely a user already put thought into the re-flow. If you do want to re-flow, I personally think it’s not too complicated to change the string literal to be on a single line again and then let swift-format break it again.

bnbarham: > Most likely a user already put thought into the re-flow

The main case that this isn't true is what @allevato mentioned, ie. format causes a split, now you indent for some other reason (nest in an if/whatever), and now never automatically reflow.

> I personally think it’s not too complicated to change the string literal to be on a single line again

Could have a refactoring for this pretty easily too.

-----

I'm somewhat split in my opinion as I can see benefits in both. I definitely think we need to have an option (probably defaulted to off) if we do reflow though. Maybe `// swift-format-ignore` is enough for cases when it's on?

allevato: After chatting offline, we decided on the above and a couple other changes:

1. Keep the default mode as `never`, to preserve existing behavior (don't break or reflow strings)
2. Even when reflowing is `onlyLinesOverLength` or `always`, only do it for regular multiline strings (`"""`), never for raw multiline strings (`#"""`, etc.)

The rationale for (2) is that it's highly likely that raw multiline strings are used for things like structured data, code, etc., whereas regular multiline strings are more likely to be simple prose.

That's not always the case, of course; swift-syntax and swift-format both have a mixture of code snippets in regular and raw multiline strings. But if we wanted to change the default behavior for this later, we could update those so that they format correctly in each case.

@ahoppen @bnbarham Do you think this is a reasonable landing spot for the time being?

bnbarham: Seems reasonable to me! Also CC @shawnhyam

PR #797, opened 2024-08-16
Add NoEmptyLinesOpeningClosingBraces rule

mateusrodriguesxyz:
This rule removes empty lines after opening braces and before closing braces:

```swift
// original

struct P {
            
    let x: Int
  
    let y: Int
          
}

// formatted

struct P {  
    let x: Int
  
    let y: Int   
}
```

That's similar to [vertical_whitespace_opening_braces](https://realm.github.io/SwiftLint/vertical_whitespace_opening_braces.html) and [vertical_whitespace_closing_braces](https://realm.github.io/SwiftLint/vertical_whitespace_opening_braces.html).

It's a style used in official swift projects such as **swift-syntax** and **sourcekit-lsp**.

### DIFF
--- a/Sources/SwiftFormat/Core/RuleRegistry+Generated.swift
+++ b/Sources/SwiftFormat/Core/RuleRegistry+Generated.swift
@@ -33,6 +33,7 @@
     "NoAssignmentInExpressions": true,
     "NoBlockComments": true,
     "NoCasesWithOnlyFallthrough": true,
+    "NoEmptyLinesOpeningClosingBraces": false,
     "NoEmptyTrailingClosureParentheses": true,
     "NoLabelsInCasePatterns": true,
     "NoLeadingUnderscores": false,
--- a/Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
+++ b/Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
@@ -1,0 +1,131 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+/// Empty lines are forbidden after opening braces and before closing braces.
+///
+/// Lint: Empty lines after opening braces and before closing braces yield a lint error.
+///
+/// Format: Empty lines after opening braces and before closing braces will be removed.
+@_spi(Rules)
+public final class NoEmptyLinesOpeningClosingBraces: SyntaxFormatRule {
+  public override class var isOptIn: Bool { return true }
+  
+  public override func visit(_ node: AccessorBlockSyntax) -> AccessorBlockSyntax {
+    var result = node
+    switch node.accessors {
+      case .accessors(let accessors):
+        result.accessors = .init(rewritten(accessors))
+      case .getter(let getter):
+        result.accessors = .init(rewritten(getter))
+    }
+    result.rightBrace = rewritten(node.rightBrace)
+    return result
+  }
+  
+  public override func visit(_ node: CodeBlockSyntax) -> CodeBlockSyntax {
+    var result = node
+    result.statements = rewritten(node.statements)
+    result.rightBrace = rewritten(node.rightBrace)
+    return result
+  }
+  
+  public override func visit(_ node: MemberBlockSyntax) -> MemberBlockSyntax {
+    var result = node
+    result.members = rewritten(node.members)
+    result.rightBrace = rewritten(node.rightBrace)
+    return result
+  }
+  
+  public override func visit(_ node: ClosureExprSyntax) -> ExprSyntax {
+    var result = node
+    result.statements = rewritten(node.statements)
+    result.rightBrace = rewritten(node.rightBrace)
+    return ExprSyntax(result)
+  }
+  
+  public override func visit(_ node: SwitchExprSyntax) -> ExprSyntax {
+    var result = node
+    result.cases = rewritten(node.cases)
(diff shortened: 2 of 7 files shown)


Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
```diff
@@ -0,0 +1,130 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+@_spi(Rules)
+public final class NoEmptyLinesOpeningClosingBraces: SyntaxFormatRule {
```
allevato, line 16: It's too bad we don't have a convenient way to just visit anything that conforms to `BracedSyntax` here.

Can you add `PrecedenceGroupDeclSyntax` and `SwitchExprSyntax`, which look like the only two other `BracedSyntax` node types that are missing here?

Sources/SwiftFormat/Core/RuleRegistry+Generated.swift
```diff
@@ -33,6 +33,7 @@
     "NoAssignmentInExpressions": true,
     "NoBlockComments": true,
     "NoCasesWithOnlyFallthrough": true,
+    "NoEmptyLinesOpeningClosingBraces": true,
```
allevato, line 36: This will change the behavior of the default configuration (and also any existing configuration that doesn't explicitly reference the new rule). Is that something you intend to do with this PR?

Adding `public override class var isOptIn: Bool { return true }` would be safer for now. Then, we should probably come up with a policy for how we want to handle changes to the default configuration in the future.

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
```diff
@@ -0,0 +1,130 @@
...
+        diagnose(.removeEmptyLinesAfter(count), on: first, anchor: .leadingTrivia(0))
+        result[index] = first.with(\.leadingTrivia, trimmedLeadingTrivia)
+      }
+    }
+    return rewrite(result).as(C.self)!
+  }
+}
+
+extension Trivia {
+  func trimmingSuperfluousNewlines() -> Trivia {
```
allevato, line 76: It looks like this function is never used, only the one with the underscore below is. Delete this and remove the underscore from the other one?

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
allevato, line 16: Can you please add a doc comment for this rule using a similar format to the other rules? `RuleDocumentation.md` should be generated from that when you run `generate-swift-format`; you don't need to update it manually.

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
```diff
@@ -0,0 +1,130 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
```
allevato, line 5: Nit: Update the ending copyright date to 2024 here.

allevato (commented):
My initial reaction to this was that I would like to keep rules involving whitespace inside the pretty printer (at least that was my plan when I got around to implementing this rule one day, which I've wanted to do for a while, so thanks for saving me the trouble!). But, that might end up being more complex looking at the logic there, so maybe this is fine the way it is—it removes the blank lines before the pretty printer ever sees them, which does the job.

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
ahoppen: To visit anything conforming to `BracedSyntax` you can do the following

```swift
public override func visitAny(_ node: Syntax) -> Syntax? {
  if let braced = node.as(BracedSyntax.self) {
    // do things
  } else {
    super.visitAny(node)
  }
}
```

allevato: In general that would work for a regular visitor, but `generate-swift-format` scans the rules to figure out which concrete nodes they visit so that it can generate the pipeline visitor.

It might not be difficult to add `visitAny` support to the pipeline, but it would be better to do it separately from this PR.

ahoppen: 👍🏽 Just wanted to put this pattern out there because it’s not terribly obvious.

Sources/SwiftFormat/Core/RuleRegistry+Generated.swift
mateusrodriguesxyz: Sorry, It wasn’t my intention.

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
mateusrodriguesxyz: Yeah, I forget this 😅

Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
mateusrodriguesxyz: I've wished more than once that `override func f(value: some T)` in a class would override any concrete method that matches.

mateusrodriguesxyz:
@allevato I've updated the PR addressing your comments.

allevato:
Windows CI is failing because of the new source file. Can you please add `Rules/NoEmptyLinesOpeningClosingBraces.swift` to [Sources/SwiftFormat/CMakeLists.txt](https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/CMakeLists.txt)?

mateusrodriguesxyz:
Done!

allevato (approved):
Thanks!

PR #798, opened 2024-08-23
Fix extraction of trailing comments only when necessary during afterToken addition

### LINKED ISSUE #795: `multiElementCollectionTrailingCommas` adds comma to line comment when element is a ternary operator
```swift
[
  1,
  true ? 1 : 2 // some comment
]
```

Gets formatted to 

```swift
[
  1,
  true ? 1 : 2  // some comment,
]
```

Adding the comma to the comment instead of after `2`.

rdar://133862349

TTOzzi:
Resolve #795


allevato (approved):
Thanks!

PR #800, opened 2024-08-24
Fix to arrange attributeList for attributes starting with ifConfig

### LINKED ISSUE #799: Attributes merged without space
The following code:
```
#if os(macOS)
@available(macOS, unavailable)
@_spi(Foo)
#endif
public let myVar = "Test"
```

Is rewritten as:
```
#if os(macOS)
  @available(macOS, unavailable)@_spi(Foo)
#endif
public let myVar = "Test"
```

The `#if` here is the important part, we don't reflow at all otherwise. Ideally we shouldn't join at all, but if we do, we should at least add a space.

(mirrors rdar://133869717)

TTOzzi:
Resolve #799 

It seems that when attributes start with `#if`, `attributes.dropLast` is empty, preventing the arrangement of attributeList inside IfConfigDeclSyntax. 
Fixed to allow arranging attribute list inside IfConfigDeclSyntax when AttributeListSyntax starts with `#if`.


allevato (approved):
Thanks!

PR #802, opened 2024-09-04
Fix infinite loop on Windows caused by checking path against "/".

### LINKED ISSUE #694: Formatting not working on Windows
The formatting does not seem to work properly when compiled on Windows. The compilation went smoothly, however, when used as CLI as well as together with vscode formatter, the behavior is incorrect:
- Running `swift-format` command from bash/powershell hangs without any output and never completes. 
- Running _Format document With_ with _apple-swift-format_ empties content of the editor the formatting is run on.

Tested on
- Windows 11 (10.0.22631 Build 22631)
- Swift 5.9.2
- swift-format 509.0.0
- apple-swift-format 1.4.3

shawnhyam:
Should fix #694.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -415,7 +415,7 @@ public struct Configuration: Codable, Equatable {
       if FileManager.default.isReadableFile(atPath: candidateFile.path) {
         return candidateFile
       }
-    } while candidateDirectory.path != "/"
+    } while candidateDirectory.pathComponents.count > 1
 
     return nil
   }
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -410,12 +410,17 @@ public struct Configuration: Codable, Equatable {
       candidateDirectory.appendPathComponent("placeholder")
     }
     repeat {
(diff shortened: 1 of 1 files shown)


allevato (commented):
I don't know what a root-only `file:` URL like `C:/` on Windows looks like and I don't have a good way to test it. @compnerd does this look reasonable to you?

compnerd:
I think that it depends on what you are using. Particularly with swift-foundation, it _should_ be `/[A-Z]:/`, unless you do the correct thing and use `withFileSystemRepresentation`, in which case it would be `[A-Z]:\`.

Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -415,7 +415,7 @@ public struct Configuration: Codable, Equatable {
       if FileManager.default.isReadableFile(atPath: candidateFile.path) {
         return candidateFile
       }
-    } while candidateDirectory.path != "/"
+    } while candidateDirectory.pathComponents.count > 1
```
compnerd, line 418: The concern that I have with this on Windows is that this would return `true` for `C:` and `\`, both of which are relative paths.

compnerd (approved):
This seems generic enough to work. It is unfortunate that Foundation doesn't have a `isRoot` property.

allevato (approved):

PR #804, opened 2024-09-05
Add a configuration to preserve whitespace on lines containing only whitespace

### LINKED ISSUE #801: Don't trim whitespace-only lines
The only thing holding me back from adopting swift-format is the inability to allow trailing whitespace in whitespace-only lines. This is my favorite Xcode feature!

image

Can we change PrettyPrinter to keep the surrounding context's indentation for empty lines, or would that be a breaking change? Can we add something that's based on a configuration to PrettyPrinter? I would love to get the ball rolling if someone can point me to the right place to implement this.

TTOzzi:
Resolve #801 

I've implemented an option to preserve whitespace on lines that consist solely of whitespace.
I believe this option will be useful for many people, but I'm not sure if I've implemented it in the correct way 🤔 
If the implementation direction is not correct, please feel free to let me know. Thank you 🙇

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -441,7 +441,7 @@ public class PrettyPrinter {
           outputBuffer.enqueueSpaces(size)
           outputBuffer.write("\\")
         }
-        outputBuffer.writeNewlines(newline)
+        outputBuffer.writeNewlines(newline, shouldIndentBlankLines: configuration.indentBlankLines)
         lastBreak = true
       } else {
         if outputBuffer.isAtStartOfLine {
@@ -458,6 +458,10 @@ public class PrettyPrinter {
 
     // Print out the number of spaces according to the size, and adjust spaceRemaining.
     case .space(let size, _):
+      if configuration.indentBlankLines, outputBuffer.isAtStartOfLine {
+        // An empty string write is needed to add line-leading indentation that matches the current indentation on a line that contains only whitespaces.
+        outputBuffer.write("")
+      }
       outputBuffer.enqueueSpaces(size)
 
     // Print any indentation required, followed by the text content of the syntax token.
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
@@ -70,8 +70,11 @@ struct PrettyPrintBuffer {
   /// subtract the previously written newlines during the second call so that we end up with the
   /// correct number overall.
   ///
-  /// - Parameter newlines: The number and type of newlines to write.
-  mutating func writeNewlines(_ newlines: NewlineBehavior) {
+  /// - Parameters:
+  ///   - newlines: The number and type of newlines to write.
+  ///   - shouldIndentBlankLines: A Boolean value indicating whether to insert spaces
+  ///     for blank lines based on the current indentation level.
+  mutating func writeNewlines(_ newlines: NewlineBehavior, shouldIndentBlankLines: Bool) {
     let numberToPrint: Int
     switch newlines {
     case .elective:
@@ -86,7 +89,13 @@ struct PrettyPrintBuffer {
     }
 
     guard numberToPrint > 0 else { return }
-    writeRaw(String(repeating: "\n", count: numberToPrint))
+    for number in 0..<numberToPrint {
+      if shouldIndentBlankLines, number >= 1 {
+        writeRaw(currentIndentation.indentation())
+      }
+      writeRaw("\n")
+    }
+
     lineNumber += numberToPrint
     isAtStartOfLine = true
     consecutiveNewlineCount += numberToPrint
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -3521,6 +3521,12 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
         leadingIndent = nil
 
       case .newlines(let count), .carriageReturns(let count), .carriageReturnLineFeeds(let count):
+        if config.indentBlankLines,
+          let leadingIndent, leadingIndent.count > 0
+        {
+          requiresNextNewline = true
+        }
+
         leadingIndent = .spaces(0)
         guard !isStartOfFile else { break }
 
@@ -3557,6 +3563,7 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
       case .spaces(let n):
         guard leadingIndent == .spaces(0) else { break }
         leadingIndent = .spaces(n)
+
       case .tabs(let n):
         guard leadingIndent == .spaces(0) else { break }
         leadingIndent = .tabs(n)
--- a/Tests/SwiftFormatTests/PrettyPrint/IndentBlankLinesTests.swift
+++ b/Tests/SwiftFormatTests/PrettyPrint/IndentBlankLinesTests.swift
@@ -1,0 +1,294 @@
+import SwiftFormat
+
+final class IndentBlankLinesTests: PrettyPrintTestCase {
+  func testIndentBlankLinesEnabled() {
+    let input =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+      \u{0020}\u{0020}
+        func bar() -> Int {
+          return 2
+        }
+      }
+      """
+
+    let expected =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+      \u{0020}\u{0020}
+        func bar() -> Int {
+          return 2
+        }
+      }
+
+      """
+    var config = Configuration.forTesting
+    config.indentBlankLines = true
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 80, configuration: config)
+  }
+
+  func testIndentBlankLinesDisabled() {
+    let input =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+      \u{0020}\u{0020}
+        func bar() -> Int {
+          return 2
+        }
+      }
+      """
+
+    let expected =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+
+        func bar() -> Int {
+          return 2
+        }
+      }
+
+      """
+    var config = Configuration.forTesting
+    config.indentBlankLines = false
+    assertPrettyPrintEqual(input: input, expected: expected, linelength: 80, configuration: config)
+  }
+
+  func testLineWithMoreWhitespacesThanIndentation() {
+    let input =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+      \u{0020}\u{0020}\u{0020}\u{0020}\u{0020}
+        func bar() -> Int {
+          return 2
+        }
+      }
+      """
+
+    let expected =
+      """
+      class A {
+        func foo() -> Int {
+          return 1
+        }
+      \u{0020}\u{0020}
+        func bar() -> Int {
+          return 2
+        }
+      }
+
+      """
+    var config = Configuration.forTesting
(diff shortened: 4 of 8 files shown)


Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -449,7 +449,12 @@ public class PrettyPrinter {
 
     // Print out the number of spaces according to the size, and adjust spaceRemaining.
     case .space(let size, _):
-      outputBuffer.enqueueSpaces(size)
+      if configuration.allowWhitespaceOnlyLines, outputBuffer.isAtStartOfLine {
+        let currentIndentationSpaceSize = outputBuffer.currentIndentation.indentation().count
+        outputBuffer.write(String(repeating: " ", count: size - currentIndentationSpaceSize))
+      } else {
```
TTOzzi, line 455: The current implementation preserves whitespace exactly as it is.

Input (Output is the same as the input)
```swift
class A {
  func foo() -> Int {
    return 1
  }
     // 5 spaces
  func bar() -> Int {
    return 2
  }
}
```



Would it be better to modify it so that only the current level of indentation is preserved, like this?

```suggestion
      if configuration.allowWhitespaceOnlyLines, outputBuffer.isAtStartOfLine {
        let currentIndentationSpaceSize = outputBuffer.currentIndentation.indentation().count
        outputBuffer.write(String())
      } else {
```

Output will be changed to

```swift
class A {
  func foo() -> Int {
    return 1
  }
  // 2 spaces(same as indentation)
  func bar() -> Int {
    return 2
  }
}
```

allevato: Since people want this to match the behavior in Xcode (which I personally think is undesirable, but...), then the number of spaces should be changed to match the expected indentation if someone were to type on that line.

For example, if I have this:

```swift
func f() {}
     // <- 5 spaces
func g() {}
```

Then I wrap it in a `struct`, once I type the closing `}`, Xcode will re-indent the whole thing. When that happens, the line has changed to this:

```swift
struct S {
  func f() {}
  // <- 2 spaces
  func g() {}
}
```

Tests/SwiftFormatTests/PrettyPrint/AllowWhitespaceOnlyLinesTests.swift
```diff
@@ -0,0 +1,99 @@
+import SwiftFormat
+
+final class AllowWhitespaceOnlyLinesTests: PrettyPrintTestCase {
+  func testAllowWhitespaceOnlyLinesEnabled() {
+    let input =
+      """
```
allevato, line 6: In this file, can you replace any trailing whitespace with literal `\u{0020}` escape sequences? That makes it clear to the reader what's being tested, and we don't risk the spaces being lost if someone's editor strips trailing spaces.

TTOzzi: Oh, that part was exactly what I was struggling with, and I didn’t know there was such a great solution! Thank you!

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
TTOzzi: To completely match the behavior with Xcode, we need to create spaces according to the indentation even when there are no existing spaces. For now, I've modified it so that the indentation is adjusted only when spaces are present. 

Do you think we need to generate spaces to match the indentation even when there are no spaces?

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -449,7 +449,11 @@ public class PrettyPrinter {
 
     // Print out the number of spaces according to the size, and adjust spaceRemaining.
     case .space(let size, _):
-      outputBuffer.enqueueSpaces(size)
+      if configuration.allowWhitespaceOnlyLines, outputBuffer.isAtStartOfLine {
+        outputBuffer.write(String())
```
allevato, line 453: Can you leave a brief comment here explaining why we do this? To a random observer, asking it to write an empty string feels odd, but it's because we need the buffer to still do its other logic (flushing the pending indentation, etc.).

I'd also just write it as `outputBuffer.write("")`; same effect but easier to read/scan.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
allevato: It looks like Xcode does insert spaces there, even when there aren't any.

@ahoppen @bnbarham Do you all have any strong opinions here? I've never liked this behavior in Xcode anyway (I always tell it to strip whitespace-only lines) but I could see myself arguing for either direction here. (And I'd rather not make it another separate knob on top of the one already being added here.)

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
TTOzzi: I've added a comment and made the modification!

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
bnbarham: I too would prefer not having an additional option for this, but I can definitely see some people wanting "ident all the things including blank lines" and others wanting "keep my trailing whitespace but don't add indentation to blank lines". So I'm not sure there's a good way around it :(.

Though TBH I don't fully understand the use case for either of these - might ask that in the issue, they do say "this is my favorite feature".

TTOzzi: That sounds good to me! 
Once we've finalized the implementation direction from the ongoing discussion, I'll update the configuration name as you suggested to make it clearer.

TTOzzi: I renamed the configuration to `indentBlankLines` for better clarity and updated the comments accordingly.

Additionally, I modified it so that when `indentBlankLines` is set to true, it adds spaces that match the current indent, even if there are no spaces in the blank lines.

**input**
```swift
struct S {
  func f() {}
// <- no space
  func g() {}
}
```

**output**
```swift
struct S {
  func f() {}
  // <- 2 spaces
  func g() {}
}
```

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -458,7 +458,12 @@ public class PrettyPrinter {
 
     // Print out the number of spaces according to the size, and adjust spaceRemaining.
     case .space(let size, _):
-      outputBuffer.enqueueSpaces(size)
+      if configuration.indentBlankLines, outputBuffer.isAtStartOfLine {
+        // An empty string write is needed to add line-leading indentation that matches the current indentation on a line that contains only whitespaces.
+        outputBuffer.write("")
```
ahoppen, line 463: Don’t we also need to enqueue the spaces when `configuration.indentBlankLines` is set?

Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
```diff
@@ -86,7 +89,15 @@ struct PrettyPrintBuffer {
     }
 
     guard numberToPrint > 0 else { return }
-    writeRaw(String(repeating: "\n", count: numberToPrint))
+    (0..<numberToPrint).forEach { number in
```
ahoppen, line 92:
```suggestion
    for number in 0..<numberToPrint {
```

is preferred.

Tests/SwiftFormatTests/PrettyPrint/IndentBlankLinesTests.swift
ahoppen, line 1: Could you add a test for two consecutive empty lines?

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3556,7 +3556,12 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
 
       case .spaces(let n):
         guard leadingIndent == .spaces(0) else { break }
-        leadingIndent = .spaces(n)
+        if config.indentBlankLines, trivia.count > index + 1, trivia[index + 1].isNewline {
+          appendToken(.space(size: n))
+          requiresNextNewline = true
```
ahoppen, line 3561: Does this mean that superfluous newline are no longer discarded if `config.indentBlankLines` is true? That seems surprising to me. Eg. 

```swift
func foo() {}


func bar() {}
```

should still be formatted to only have one newline.

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
TTOzzi: Oh right, it seems we should enqueue this consistently, regardless of `configuration.indentBlankLines`.

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
TTOzzi: Looking at it again, this logic doesn’t seem appropriate 🤔 
In the initial implementation, this was added along with logic to retain user-entered spaces by inserting a `space` token, but it doesn’t seem necessary with the new implementation approach. 
I’ll keep the logic that adds spaces as `leadingIndent` and, when processing newlines, add a soft newline if there’s a `leadingIndent`, ensuring that the actual output doesn’t exceed the configured limit lines.

Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
```diff
@@ -86,7 +89,15 @@ struct PrettyPrintBuffer {
...
 
     guard numberToPrint > 0 else { return }
-    writeRaw(String(repeating: "\n", count: numberToPrint))
+    for number in 0..<numberToPrint {
+      if shouldIndentBlankLines, number >= 1 {
+        writeRaw(currentIndentation.indentation())
+        writeRaw("\n")
+      } else {
+        writeRaw("\n")
+      }
```
ahoppen, line 98:
```suggestion
      if shouldIndentBlankLines, number >= 1 {
        writeRaw(currentIndentation.indentation())
      }
      writeRaw("\n")
```

Would be slightly simpler

Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
```diff
@@ -3521,6 +3521,11 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
         leadingIndent = nil
 
       case .newlines(let count), .carriageReturns(let count), .carriageReturnLineFeeds(let count):
+        if config.indentBlankLines,
+           let _leadingIndent = leadingIndent, _leadingIndent.count > 0 {
```
ahoppen, line 3525:
```suggestion
        if config.indentBlankLines, let leadingIndent, leadingIndent.count > 0 {
```

ahoppen (commented):
Thanks. Just two minor comments, otherwise LGTM.

TTOzzi:
> Thanks. Just two minor comments, otherwise LGTM.

I've applied all the comments you suggested. Thank you for the kind review 🙇‍♂️

ahoppen (approved):
Thank you for implementing this, @TTOzzi. Looks good.

TTOzzi:
Oops, sorry. The format check failed 😓 
I’ve fixed it and force-pushed.

ahoppen:
No worries. Already re-triggered CI.

TTOzzi:
> No worries. Already re-triggered CI.

Oh no, it failed on a different file this time 😨
I did the same formatting with `curl -s https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/check-swift-format.sh | bash` as in CI, is there any other way to fix this?

ahoppen:
You should be able to just cd to the swift-format directory and then run `swift format -r -i .`

TTOzzi:
> You should be able to just cd to the swift-format directory and then run `swift format -r -i .`

I'm done! Thank you!

PR #805, opened 2024-09-10
Fix link to configuration in rule documentation file

mattia:
The link in the `RuleDocumentation.md` file to the configuration was broken because we need to provide a relative URL path


ahoppen (approved):
Thank you!

PR #821, opened 2024-09-29
Create a GitHub action to create a release

ahoppen:
This should replace the script that I have locally to test swift-format for a release and at the same time expand the configurations that are being tested.

A sample run can be seen here: https://github.com/ahoppen/swift-format/actions/runs/11089874502

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -1,0 +1,108 @@
+name: Publish Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      prerelease:
+        type: boolean
+        description: "Prerelease"
+        # Whether to create a prerelease or proper release
+        default: true
+        required: true
+      swift_format_version:
+        type: string
+        default: 601.0.0
+        description: "swift-format version"
+        # The version of swift-format to tag. If this is a prerelease, `-prerelease-<date>` is added to this version.
+        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,108 @@
...
+
+jobs:
+  check_triggering_actor:
+    name: Check user is allowed to create release
+    # Only a single user should be allowed to create releases to avoid two people triggering the creation of a release
+    # at the same time. If the release manager changes between users, update this condition.
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          if [[ "${{ github.triggering_actor }}" != "ahoppen" ]]; then
```
shahmishal, line 33: We should create a team for swift-format, so it's not blocked on one person.

shahmishal (commented):
Overall looks good, just one suggestion on using team to avoid being blocked on one person. You might want to add @bnbarham.

.github/workflows/publish_release.yml
ahoppen: My thought was that it’s pretty easy to modify the workflow to shift responsibilities. That way the responsibility transfer is explicit.

shahmishal (approved):

allevato (approved):

PR #824, opened 2024-09-30
[CI] Add soundness check

shahmishal:


allevato:
I was just getting ready to ask about how we could integrate a formatting check. Really happy to see a common workflow for this 😄 

The repo definitely _won't_ pass the formatting check today (since we haven't been able to enforce it easily, I didn't want to add more burden to the contribution workflow). I'd be happy to take some time to figure out what my preferred format configuration should be and apply it throughout, though.

@ahoppen @bnbarham If you have logistical reasons that it should match swift-syntax's format configuration, I can live with that, but I'd like to do something even more opinionated here. WDYT?

.github/workflows/pull_request.yml
```diff
@@ -8,3 +8,15 @@ jobs:
   tests:
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+  soundness:
```
ahoppen, line 11: Can you consistently use 2 space for indentation?

.github/workflows/pull_request.yml
```diff
@@ -8,3 +8,15 @@ jobs:
...
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
+    with:
+      api_breakage_check_enabled: true
+      broken_symlink_check_enabled: true
+      docs_check_enabled: true
+      format_check_enabled: true
+      license_header_check_enabled: false
+      license_header_check_project_name: "Swift.org"
+      shell_check_enabled: true
+      unacceptable_language_check_enabled: true
```
ahoppen, line 22: All of these are enabled in the workflow by default, right? So we could omit all the `: true` lines, right?

.github/workflows/pull_request.yml
```diff
@@ -8,3 +8,15 @@ jobs:
...
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+  soundness:
+    name: Soundness
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
+    with:
+      api_breakage_check_enabled: true
+      broken_symlink_check_enabled: true
+      docs_check_enabled: true
+      format_check_enabled: true
```
shahmishal, line 18:
```suggestion
      format_check_enabled: false
```

For now disable the format check until the repo is updated.

PR #837, opened 2024-10-05
Fix relative symlinks when using --follow-symlinks

### LINKED ISSUE #836: The option `--follow-symlinks` can't follow relative symlinks
## Description

The subcommands `lint` and `format` have the option `--follow-symlinks`. However, although it can follow absolute symlinks, it can't follow relative symlinks.

## Steps to reproduce

```console
$ mkdir symlink-test
$ cd symlink-test/
$ echo 'let foo:Int=0' > .foo.swift
$ ln -s .foo.swift bar.swift
$ swift format dump-configuration > .swift-format
$ cd ..
$ swift format lint -rs --follow-symlinks symlink-test/
<unknown>: error: Unable to open .foo.swift: file is not readable or does not exist
```

If the current working directory is the symlink's directory, it works.

```console
$ cd symlink-test
$ swift format lint -rs --follow-symlinks .
.foo.swift:1:9: warning: [Spacing] add 1 space
.foo.swift:1:12: warning: [Spacing] add 1 space
.foo.swift:1:13: warning: [Spacing] add 1 space
```

It's because the following line is using `relativeTo: workingDirectory`.

https://github.com/swiftlang/swift-format/blob/8068ff75e23084bfdc0706a70fa1bc926c9ca71e/Sources/SwiftFormat/Utilities/FileIterator.swift#L152

## Environment

- swift-format (bundled in Swift 6.0.1 toolchain)
- Debian 12 (aarch64)

kkebo:
fixes #836 

Changes:

- Add a relative symlink to `FileIteratorTests`
  - This causes test failures if the following fixes are not applied
- Fix relative symlinks when using `--follow-symlinks`
- Eliminate duplicated files with `standardizedFileURL` when comparing them in `FileIterator.next()`

### DIFF
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -97,12 +97,12 @@ public struct FileIterator: Sequence, IteratorProtocol {
           output = next
         }
       }
-      if let out = output, visited.contains(out.absoluteURL.standardized.standardizedFileURL.path) {
+      if let out = output, visited.contains(out.standardizedFileURL.path) {
         output = nil
       }
     }
     if let out = output {
-      visited.insert(out.absoluteURL.standardized.standardizedFileURL.path)
+      visited.insert(out.standardizedFileURL.path)
     }
     return output
   }
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -135,7 +135,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
         else {
           break
         }
-        path = URL(fileURLWithPath: destination, isDirectory: false, relativeTo: item).path
+        path = URL(fileURLWithPath: destination, relativeTo: item).path
         fallthrough
 
       case .typeRegular:
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -97,12 +97,12 @@ public struct FileIterator: Sequence, IteratorProtocol {
           output = next
         }
       }
-      if let out = output, visited.contains(out.absoluteURL.standardized.path) {
+      if let out = output, visited.contains(out.absoluteURL.standardized.standardizedFileURL.path) {
         output = nil
       }
     }
     if let out = output {
-      visited.insert(out.absoluteURL.standardized.path)
+      visited.insert(out.absoluteURL.standardized.standardizedFileURL.path)
     }
     return output
   }
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -79,7 +79,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
           else {
             break
           }
-          next = URL(fileURLWithPath: destination)
+          next = URL(fileURLWithPath: destination, relativeTo: next)
           fallthrough
 
         case .typeDirectory:
@@ -135,7 +135,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
         else {
           break
         }
-        path = destination
+        path = URL(fileURLWithPath: destination, isDirectory: false, relativeTo: item).path
         fallthrough
 
       case .typeRegular:
--- a/Tests/SwiftFormatTests/Utilities/FileIteratorTests.swift
+++ b/Tests/SwiftFormatTests/Utilities/FileIteratorTests.swift
@@ -18,6 +18,7 @@ final class FileIteratorTests: XCTestCase {
     try touch("project/.hidden.swift")
     try touch("project/.build/generated.swift")
     try symlink("project/link.swift", to: "project/.hidden.swift")
+    try symlink("project/rellink.swift", relativeTo: ".hidden.swift")
   }
 
   override func tearDownWithError() throws {
@@ -55,7 +56,10 @@ final class FileIteratorTests: XCTestCase {
     // passed to the iterator. This is meant to avoid situations where a symlink could be hidden by
     // shell expansion; for example, if the user writes `swift-format --no-follow-symlinks *`, if
     // the current directory contains a symlink, they would probably *not* expect it to be followed.
-    let seen = allFilesSeen(iteratingOver: [tmpURL("project/link.swift")], followSymlinks: false)
+    let seen = allFilesSeen(
+      iteratingOver: [tmpURL("project/link.swift"), tmpURL("project/rellink.swift")],
+      followSymlinks: false
+    )
     XCTAssertTrue(seen.isEmpty)
   }
 }
@@ -81,11 +85,19 @@ extension FileIteratorTests {
     }
   }
 
-  /// Create a symlink between files or directories in the test's temporary space.
+  /// Create a absolute symlink between files or directories in the test's temporary space.
   private func symlink(_ source: String, to target: String) throws {
     try FileManager.default.createSymbolicLink(
       at: tmpURL(source),
       withDestinationURL: tmpURL(target)
+    )
+  }
+
+  /// Create a relative symlink between files or directories in the test's temporary space.
+  private func symlink(_ source: String, relativeTo target: String) throws {
+    try FileManager.default.createSymbolicLink(
+      atPath: tmpURL(source).path,
+      withDestinationPath: target
     )
   }
 


kkebo:
It seemed that `testFollowSymlinks` failed in macOS. I'll investigate it.

https://ci.swift.org/job/swift-syntax-PR-macOS/4601/console

```
Test Case '-[SwiftFormatTests.FileIteratorTests testFollowSymlinks]' started.
/Users/ec2-user/jenkins/workspace/swift-syntax-PR-macOS/branch-main/swift-format/Tests/SwiftFormatTests/Utilities/FileIteratorTests.swift:37: error: -[SwiftFormatTests.FileIteratorTests testFollowSymlinks] : XCTAssertEqual failed: ("4") is not equal to ("3")
Test Case '-[SwiftFormatTests.FileIteratorTests testFollowSymlinks]' failed (0.699 seconds).
```

kkebo:
The above error did not occur on macOS Sequoia. So I will prepare macOS Ventura. However, it will take time.

kkebo:
I have successfully reproduced the issue on macOS 13.6.9 on EC2.

When the test failed, the local variable [`seen`](https://github.com/kkebo/swift-format/blob/f824817fc015c15e6e3762391bf9d82e5cd1e572/Tests/SwiftFormatTests/Utilities/FileIteratorTests.swift#L36) looked like this:

```swift
[
    "/private/var/folders/5r/59x37fyj13l2hjhq53g287_80000gn/T/TemporaryItems/NSIRD_xctest_Jjw9up/project/real2.swift",
    "/private/var/folders/5r/59x37fyj13l2hjhq53g287_80000gn/T/TemporaryItems/NSIRD_xctest_Jjw9up/project/.hidden.swift",
    "/var/folders/5r/59x37fyj13l2hjhq53g287_80000gn/T/TemporaryItems/NSIRD_xctest_Jjw9up/project/.hidden.swift",
    "/private/var/folders/5r/59x37fyj13l2hjhq53g287_80000gn/T/TemporaryItems/NSIRD_xctest_Jjw9up/project/real1.swift"
]
```

So it seems that this test failure occurs if a relative symbolic link is linked to a file that its path contains symbolic links.

kkebo:
Fixed: ef73806f30f24736156221665543e73bff8afff8

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -97,12 +97,12 @@ public struct FileIterator: Sequence, IteratorProtocol {
           output = next
         }
       }
-      if let out = output, visited.contains(out.absoluteURL.standardized.path) {
+      if let out = output, visited.contains(out.absoluteURL.standardized.standardizedFileURL.path) {
```
ahoppen, line 100: Do we still need `.standardized` or is this included in `standardizedFileURL`?

Same below.

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -135,7 +135,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
         else {
           break
         }
-        path = destination
+        path = URL(fileURLWithPath: destination, isDirectory: false, relativeTo: item).path
```
ahoppen, line 138: The symlink could point to a directory, right? I would just remove the `isDirectory` bit so that URL can infer that information from the file system.

ahoppen (commented):
Thanks for the PR and tracking down the failure, @kkebo. Looks good to me, just two very minor comments.

Sources/SwiftFormat/Utilities/FileIterator.swift
kkebo: I just fixed in 5f90fb3d8646b74946bd9917c72d49d8acdeba16, but I don't think this change is meaningful because [the existing code](https://github.com/swiftlang/swift-format/blob/06ca47b4769b9b59028d71b6c3b0ab4360edb35f/Sources/SwiftFormat/Utilities/FileIterator.swift#L152) always treats it as `isDirectory: false`.

Sources/SwiftFormat/Utilities/FileIterator.swift
ahoppen: Oh, interesting. Could you add a test case for this so we don’t carelessly remove `standardized` in the future?

kkebo: Sorry, I couldn't add a test case because it doesn't matter in swift-format.

I found that the code example above was bad. The reason that `URL(filePath: "..").standardized.path` was `"/home/kebo"` was that the given URL was relative.

[`URL.standardized`](https://github.com/swiftlang/swift-foundation/blob/3f9171072dc842c04c77ed2fa578d1b0cd123a04/Sources/FoundationEssentials/URL/URL.swift#L1763-L1781) follows the following steps:

1. Create `components` from `_parseInfo` (but `_parseInfo` does not contain the `baseURL` info)
1. Assign `components.percentEncodedPath.removingDotSegments` to `components.percentEncodedPath`
1. Return `components.url(relativeTo: baseURL)!`

In step 2, `".."` will be `""` because `URLComponents` doesn't have the `baseURL` info and `removingDotSegments` will never append dots. [This test case](https://github.com/swiftlang/swift-foundation/blob/3f9171072dc842c04c77ed2fa578d1b0cd123a04/Tests/FoundationEssentialsTests/StringTests.swift#L424) also indicates that. Because of that, `removingDotSegments` can't go up above the `baseURL`. So in step 3, it will return a `URL` just like `URL(filePath: "", relativeTo: baseURL)`, so its `.path` equals to `baseURL.path`. Therefore, `URL(filePath: "..").standardized.path` became `"/home/kebo"`.

However, if the given URL is absolute, `removingDotSegments` can go up above the base URL because all information including the base URL is included in `URLComponents`. In addition, `baseURL` will be `nil`. So `URL(filePath: "..").absoluteURL.standardized.path` will be `"/home"`, and it's the same as `URL(filePath: "..").standardizedFileURL.path`.

In swift-format's `FileIterator`, `standardized` always follows `absoluteURL`. So I can assure that what I said [before](https://github.com/swiftlang/swift-format/pull/837#discussion_r1803669289) does not occur in swift-format.

----

So I looked through `URL.standardizedFileURL` once again.

- [`URL.standardizedFileURL`](https://github.com/swiftlang/swift-foundation/blob/3f9171072dc842c04c77ed2fa578d1b0cd123a04/Sources/FoundationEssentials/URL/URL.swift#L1789-L1799)
- [`String.standardizingPath`](https://github.com/swiftlang/swift-corelibs-foundation/blob/cb6b2fa46992e04ae3f77fae30e2d86cebd2b440/Sources/Foundation/NSPathUtilities.swift#L318-L329)
- [`String.expandingTildeInPath`](https://github.com/swiftlang/swift-corelibs-foundation/blob/cb6b2fa46992e04ae3f77fae30e2d86cebd2b440/Sources/Foundation/NSPathUtilities.swift#L284-L303)
- [`String.resolvingSymlinksInPath`](https://github.com/swiftlang/swift-corelibs-foundation/blob/cb6b2fa46992e04ae3f77fae30e2d86cebd2b440/Sources/Foundation/NSPathUtilities.swift#L331-L364)

`URL.standardizedFileURL` seems to be `expandingTildeInPath` + `resolvingSymlinksInPath` + removing the `/var/automount` prefix + making an absolute URL, and `resolvingSymlinksInPath` seems to include `URL.standardized` in addition to resolve symlinks and removing the `/private` prefix.

So I think it is safe to replace `.absoluteURL.standardized` with `.standardizedFileURL` at least on macOS and other Unix-like OSes. How about doing this instead of adding a test case for `.standardized`?

I have already tested the following diff can pass the all tests on macOS and Linux.

```diff
diff --git a/Sources/SwiftFormat/Utilities/FileIterator.swift b/Sources/SwiftFormat/Utilities/FileIterator.swift
index 8189f18..9d1a1d2 100644
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -97,12 +97,12 @@ public struct FileIterator: Sequence, IteratorProtocol {
           output = next
         }
       }
-      if let out = output, visited.contains(out.absoluteURL.standardized.standardizedFileURL.path) {
+      if let out = output, visited.contains(out.standardizedFileURL.path) {
         output = nil
       }
     }
     if let out = output {
-      visited.insert(out.absoluteURL.standardized.standardizedFileURL.path)
+      visited.insert(out.standardizedFileURL.path)
     }
     return output
   }
```

Sorry for a very long post.

ahoppen: Thanks for the detailed explanation, very nice read. What you suggest sounds like the best solution, let’s do that.

kkebo: I pushed b9065b18d8435618635e0f619e62a97e7dee55fb.

ahoppen (approved):

allevato (approved):

PR #839, opened 2024-10-05
Update README.md to make it platform agnostic

0xTim:
Make the README platform agnostic

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ then you can get it in three different ways:
 
 ### Included in the Swift Toolchain
 
-Swift 6 (included with Xcode 16) and above include swift-format in the toolchain. You can run `swift-format` from anywhere on the system using `swift format` (notice the space instead of dash). To find the path at which `swift-format` is installed on macOS, run `xcrun --find swift-format`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,9 +63,9 @@ For example, if you are using Xcode 13.3 (Swift 5.6), you will need
 If you are mainly interested in using swift-format (rather than developing it),
 then you can get it in three different ways:
 
-### Included in Xcode
+### Included in the Swift Toolchain
 
-Xcode 16 and above include swift-format in the toolchain. You can run `swift-format` from anywhere on the system using `swift format` (notice the space instead of dash). To find the path at which `swift-format` is installed, run `xcrun --find swift-format`.
+Swift 6 (included with Xcode 16) and above include swift-format in the toolchain. You can run `swift-format` from anywhere on the system using `swift format` (notice the space instead of dash). To find the path at which `swift-format` is installed on macOS, run `xcrun --find swift-format`.
```
allevato, line 68: "on macOS" might be better expressed as "in Xcode".

@ahoppen Thoughts on wording here from Apple's end?

bnbarham: Is the last sentence needed if the section is now generically "in the Swift Toolchain"?

ahoppen: I think we should keep it because usually people don’t know the path of the toolchain inside Xcode. And I agree that `in Xcode` makes more sense.

0xTim: Updated!

ahoppen (approved):

PR #841, opened 2024-10-07
Support building swift-format using SwiftPM for toolchain builds on Windows

ahoppen:
This allows us to also run tests of swift-format on Windows.


Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
```diff
@@ -86,7 +86,9 @@ public final class NoEmptyLinesOpeningClosingBraces: SyntaxFormatRule {
       let (trimmedLeadingTrivia, count) = first.leadingTrivia.trimmingSuperfluousNewlines()
       if trimmedLeadingTrivia.sourceLength != first.leadingTriviaLength {
         diagnose(.removeEmptyLinesAfter(count), on: first, anchor: .leadingTrivia(0))
-        result[index] = first.with(\.leadingTrivia, trimmedLeadingTrivia)
+        var first = first
+        first.leadingTrivia = trimmedLeadingTrivia
+        result[index] = first
```
ahoppen, line 91: I haven’t had a chance to reduce the failure but I think test execution crashes with the previous code due to https://github.com/swiftlang/swift/issues/67676.

allevato: This is fine anyway, I've come to mildly disfavor the bespoke `with` methods and just use normal value type mutation for operations like this.

allevato (approved):

PR #850, opened 2024-10-11
[6.0.2] Fix infinite searching for .swift-format file on Windows and Linux

ahoppen:
- **Explanation**: Deleting path components from a URL on Windows will never end up with a path `/` because Windows has drive letters. On Linux, we end up with an empty path instead of `/` because of https://github.com/swiftlang/swift-foundation/issues/980.
- **Scope**: Directory traversal in the search for a `.swift-format` file
- **Risk**: Low, the existing logic is still in place, we just increase the number of directories that we consider root and all of those checks should be safe
- **Testing**: Add test cases, manually verified that we terminate the directory traversal on Linux and Windows
- **Issue**: https://github.com/swiftlang/swift-format/issues/847 / https://github.com/swiftlang/swift-format/issues/694 / rdar://137738047 / rdar://126948386
- **Reviewer**:   @allevato on https://github.com/swiftlang/swift-format/pull/802 and https://github.com/swiftlang/swift-format/pull/848

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -328,7 +328,7 @@ public struct Configuration: Codable, Equatable {
       if FileManager.default.isReadableFile(atPath: candidateFile.path) {
         return candidateFile
       }
-    } while candidateDirectory.path != "/"
+    } while !candidateDirectory.isRoot
 
     return nil
   }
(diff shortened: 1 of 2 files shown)


bnbarham (approved):

ahoppen:
Testing via https://github.com/swiftlang/swift-syntax/pull/2756

PR #853, opened 2024-10-14
Remove duplicated dependencies in Package.swift

ahoppen:
Not sure how I missed this.


Package.swift
```diff
@@ -120,9 +116,7 @@ var targets: [Target] = [
       .product(name: "Markdown", package: "swift-markdown"),
       .product(name: "SwiftOperators", package: "swift-syntax"),
       .product(name: "SwiftParser", package: "swift-syntax"),
```
allevato, line 118: Should these two also be moved into `swiftSyntaxDependencies`?

ahoppen: Indeed.

allevato (approved):

PR #859, opened 2024-10-20
[CI] Add support for post merge testing

shahmishal:


.github/workflows/post_merge.yml
```diff
@@ -0,0 +1,16 @@
+name: Post merge
+
+on:
+  push:
+    branches: [main]
+
```
ahoppen, line 6: Would it be better to just add 

```yml
  push:
    branches: [main]
```

To `pull_reqeust.yml` and rename the workflow’s name from `Pull request` to `Test`? That way the pull request tests and the post-merge tests are always the same. I am a little worried that they will drift apart otherwise.

Issue #862, opened 2024-10-23
`FileIterator` returns broken URLs when the working directory is `/`


kkebo:
## Description

When you use the `-r` option and `workingDirectory.path` is `"/"`, `String(path.dropFirst(workingDirectory.path.count + 1))` will remove the first character of the file path, so `FileIterator` returns broken URLs and swift-format can't read any file.

https://github.com/swiftlang/swift-format/blob/3ca1a182f2990fb34403fc521b6bd3b839555527/Sources/SwiftFormat/Utilities/FileIterator.swift#L149

## How to reproduce the issue

For instance, on wasm32-unknown-wasi, `URL(fileURLWithPath: ".")` will be `/`, so this is the problem.

And also, the issue can be reproducible on normal platforms like this:

```console
$ pwd
/
$ ls -l | grep *.swift
.rw-r--r--. root root  12 B  Thu Oct 24 02:39:39 2024 test.swift                                                                                                     
$ swift format lint -rs .
<unknown>: error: Unable to open est.swift: file is not readable or does not exist
```

My environment:

- Linux aarch64
- Swift 6.0 development snapshot toolchain (6.0-snapshot-2024-10-08)
  ```
  Swift version 6.0.3-dev (LLVM 81a2745a5c40b5f, Swift 8f95e04bb5d9d9a)
  Target: aarch64-unknown-linux-gnu
  ```
- swift-format bundled with the Swift toolchain

ahoppen:
Synced to Apple’s issue tracker as rdar://138492445

kkebo:
Resolved by #865

PR #865, opened 2024-10-25
Fix support for `FileIterator` when working directory is `/`

macshome:
Added an additional check that tests to see if the working directory is root to solve #862.

### Simple Fix
Simple check that looks to see if the working directory is root.
```swift
 let relativePath =
          path.hasPrefix(workingDirectory.path) && FileManager.default.currentDirectoryPath != "/"
          ? String(path.dropFirst(workingDirectory.path.count + 1))
          : path
        output =
          URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: workingDirectory)
``` 
#
### Testing

I couldn't find a non-intrusive way to simulate changing the working directory of the unit tests. This path remains uncovered as a result of that.

Manual testing with the debug build. The behavior of #862 is resolved.

**Works from `/tmp`**
```shell
/tmp % pwd
/tmp

/tmp % ls -l testsf/*.swift 
-rw-r--r--@ 1 josh.wisenbaker  wheel  308 Nov  4  2022 testsf/bp.swift

/tmp % <trimmed>/Products/Debug/swift-format lint -rs .
bp.swift:1:31: warning: [GroupNumericLiterals] group every 3 digits in this decimal literal using a '_' separator
``` 

**Now works from `/` as well**
```shell 
% cd /           
/ % pwd                 
/

/ % ls -l /tmp/testsf/*.swift
-rw-r--r--@ 1 josh.wisenbaker  wheel  308 Nov  4  2022 /tmp/testsf/bp.swift

/ % <trimmed>/Products/Debug/swift-format lint -rs /tmp/testsf
/private/tmp/testsf/bp.swift:1:31: warning: [GroupNumericLiterals] group every 3 digits in this decimal literal using a '_' separator
```

I'll need to spin up a linux VM to test if we get relative paths when running from root.

### DIFF
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -12,6 +12,10 @@
 
 import Foundation
 
+#if os(Windows)
+import WinSDK
+#endif
+
 /// Iterator for looping over lists of files and directories. Directories are automatically
 /// traversed recursively, and we check for files with a ".swift" extension.
 @_spi(Internal)
@@ -32,7 +36,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
 
   /// The current working directory of the process, which is used to relativize URLs of files found
   /// during iteration.
-  private let workingDirectory = URL(fileURLWithPath: ".")
+  private let workingDirectory: URL
 
   /// Keep track of the current directory we're recursing through.
   private var currentDirectory = URL(fileURLWithPath: "")
@@ -46,8 +50,13 @@ public struct FileIterator: Sequence, IteratorProtocol {
   /// Create a new file iterator over the given list of file URLs.
   ///
   /// The given URLs may be files or directories. If they are directories, the iterator will recurse
-  /// into them.
-  public init(urls: [URL], followSymlinks: Bool) {
+  /// into them. Symlinks are never followed on Windows platforms as Foundation doesn't support it.
+  /// - Parameters:
+  ///   - urls: `Array` of files or directories to iterate.
+  ///   - followSymlinks: `Bool` to indicate if symbolic links should be followed when iterating.
+  ///   - workingDirectory: `URL` that indicates the current working directory. Used for testing.
+  public init(urls: [URL], followSymlinks: Bool, workingDirectory: URL = URL(fileURLWithPath: ".")) {
+    self.workingDirectory = workingDirectory
     self.urls = urls
     self.urlIterator = self.urls.makeIterator()
     self.followSymlinks = followSymlinks
@@ -144,12 +153,13 @@ public struct FileIterator: Sequence, IteratorProtocol {
         // if the user passes paths that are relative to the current working directory, they will
         // be displayed as relative paths. Otherwise, they will still be displayed as absolute
         // paths.
-        let relativePath =
-          path.hasPrefix(workingDirectory.path)
-          ? String(path.dropFirst(workingDirectory.path.count + 1))
-          : path
-        output =
-          URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: workingDirectory)
+        let relativePath: String
+        if !workingDirectory.isRoot, path.hasPrefix(workingDirectory.path) {
+          relativePath = String(path.dropFirst(workingDirectory.path.count).drop(while: { $0 == "/" || $0 == #"\"# }))
+        } else {
+          relativePath = path
+        }
+        output = URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: workingDirectory)
 
       default:
         break
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -16,3 +16,4 @@ jobs:
     with:
       license_header_check_enabled: false
       license_header_check_project_name: "Swift.org"
+      api_breakage_check_allowlist_path: "api-breakages.txt"
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -486,17 +486,3 @@ public struct NoAssignmentInExpressionsConfiguration: Codable, Equatable {
 
   public init() {}
 }
-
-fileprivate extension URL {
-  var isRoot: Bool {
-    #if os(Windows)
-    // FIXME: We should call into Windows' native check to check if this path is a root once https://github.com/swiftlang/swift-foundation/issues/976 is fixed.
-    // https://github.com/swiftlang/swift-format/issues/844
-    return self.pathComponents.count <= 1
-    #else
-    // On Linux, we may end up with an string for the path due to https://github.com/swiftlang/swift-foundation/issues/980
-    // TODO: Remove the check for "" once https://github.com/swiftlang/swift-foundation/issues/980 is fixed.
-    return self.path == "/" || self.path == ""
-    #endif
-  }
-}
--- a/Sources/SwiftFormat/CMakeLists.txt
+++ b/Sources/SwiftFormat/CMakeLists.txt
@@ -97,7 +97,8 @@ add_library(SwiftFormat
   Rules/UseTripleSlashForDocumentationComments.swift
   Rules/UseWhereClausesInForLoops.swift
   Rules/ValidateDocumentationComments.swift
-  Utilities/FileIterator.swift)
+  Utilities/FileIterator.swift
+  Utilities/URL+isRoot.swift)
 target_link_libraries(SwiftFormat PUBLIC
   SwiftMarkdown::Markdown
   SwiftSyntax::SwiftSyntax
--- a/Sources/SwiftFormat/Utilities/URL+isRoot.swift
+++ b/Sources/SwiftFormat/Utilities/URL+isRoot.swift
@@ -1,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension URL {
+  @_spi(Testing) public var isRoot: Bool {
+    #if os(Windows)
+    // FIXME: We should call into Windows' native check to check if this path is a root once https://github.com/swiftlang/swift-foundation/issues/976 is fixed.
+    // https://github.com/swiftlang/swift-format/issues/844
+    var pathComponents = self.pathComponents
+    if pathComponents.first == "/" {
+      // Canonicalize `/C:/` to `C:/`.
+      pathComponents = Array(pathComponents.dropFirst())
+    }
+    return pathComponents.count <= 1
+    #else
+    // On Linux, we may end up with an string for the path due to https://github.com/swiftlang/swift-foundation/issues/980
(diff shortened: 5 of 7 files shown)


Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -145,7 +145,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
         // be displayed as relative paths. Otherwise, they will still be displayed as absolute
         // paths.
         let relativePath =
-          path.hasPrefix(workingDirectory.path)
+          path.hasPrefix(workingDirectory.path) && path.first != "/"
```
ahoppen, line 148: Doesn’t this mean that we never relativize paths on Unix because all paths will start with a `/`? Don’t you want to check that `workingDirectory` is `/`? I might be misunderstanding your change though.

ahoppen (commented):
Thanks for opening a PR and addressing this issue @macshome.

To test this, what do you think about making `workingDirectory` an argument to `FileIterator.init` and just default it to `URL(fileURLWithPath: ".")`? That way you could set the working directory and we can also craft tests for Windows to ensure we don’t have any issues there.

macshome:
> Thanks for opening a PR and addressing this issue @macshome.
> 
> To test this, what do you think about making `workingDirectory` an argument to `FileIterator.init` and just default it to `URL(fileURLWithPath: ".")`? That way you could set the working directory and we can also craft tests for Windows to ensure we don’t have any issues there.

That's what I had thought about for setting a working directory, but that value is a constant in the code...

```swift
private let workingDirectory = URL(fileURLWithPath: ".")
```

Sources/SwiftFormat/Utilities/FileIterator.swift
macshome: Hmm... You seem to be correct here. As far as I can tell the reason for using relative paths is just for nice diagnostic output.

```
        // We attempt to relativize the URLs based on the current working directory, not the
        // directory being iterated over, so that they can be displayed better in diagnostics. Thus,
        // if the user passes paths that are relative to the current working directory, they will
        // be displayed as relative paths. Otherwise, they will still be displayed as absolute
        // paths.
```

And with `workingDirectory` being a constant of `.` this may take a bit more work. If we did change `workingDirectory` to be passed in then it would also allow for easier testing. I'll work on it a bit more and see what I can come up with.

allevato:
> That's what I had thought about for setting a working directory, but that value is a constant in the code...
> 
> ```swift
> private let workingDirectory = URL(fileURLWithPath: ".")
> ```

If you pass it in the initializer, you could make that the default value but then let tests specify whatever they want there. It's a member property, so it can certainly be changed by the initializer.

macshome:
> If you pass it in the initializer, you could make that the default value but then let tests specify whatever they want there. It's a member property, so it can certainly be changed by the initializer.

I'll take a trip down this path and see about getting it covered with tests so it doesn't break again.

macshome:
@allevato I updated the tests to use `[URL]` instead of `[String]` and that let me then actually change the working directory to test that we don't clip names when running from `/` and that relative names are also used properly when applicable.

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -145,7 +145,7 @@ public struct FileIterator: Sequence, IteratorProtocol {
         // be displayed as relative paths. Otherwise, they will still be displayed as absolute
         // paths.
         let relativePath =
-          path.hasPrefix(workingDirectory.path)
+          path.hasPrefix(workingDirectory.path) && FileManager.default.currentDirectoryPath != "/"
```
allevato, line 148: @compnerd What will this do on Windows? Probably not what's intended, I think?

macshome: That was my concern, I don't know how Foundation will deal with that there.

macshome: Actually it might be OK.

 https://github.com/swiftlang/swift-foundation/blob/ba000ae07c106149c2710bb9cb8c6993e20734b2/Sources/FoundationEssentials/FileManager/FileManager%2BDirectories.swift#L467

ahoppen: `FileManager.default.currentDirectoryPath` will never be `/` on Windows because Windows always has a drive letter, eg. `C:\`. Checking if a URL is a root path on Windows is not quite trivial because you also need to account for networks shares eg. `\\your-server\blah` and the only good way of doing it issuing the `PathCchIsRoot` Windows API (eg. https://github.com/swiftlang/swift-tools-support-core/blob/5b130e04cc939373c4713b91704b0c47ceb36170/Sources/TSCBasic/Path.swift#L491). 

But maybe it’s not really an issue on Windows anyway, depending on how test output looks like. This is why I suggested making `workingDirectory` changeable from tests and then we can just check.

macshome: I don't know that it is an issue on Windows either. Most people seem to have discovered it when using WASM. Are there Windows and Linux runners for CI so that we can see if this breaks anything? Otherwise I can install a toolchain on Windows here and check.

I can still make `workingDirectory` changeable from tests, but at this point I'm worried about the validity of the entire approach on Windows.

macshome: I suppose I could also implement the `isRoot` property from the swift-tools-support-core as well in `FileIterator`. That way it would work on Windows as well.

macshome: These tests do fail on Windows when trying to change the working directory. I'll update everything and get it working.

ahoppen: If you don’t have a Windows machine to test this on, just write code that works to the best of you knowledge and some test cases for it. We have Windows PR testing (we just need to approve runs) and I’m happy to get the Windows code over the finish line if there are minor issues.

ahoppen: Oh, that’s great. Let me know if you have any questions.

macshome: Will do. Hoping to get back to this today.

ahoppen: Let me know when you think the PR is ready for review again.

macshome:
Converting to draft while I work out the Windows stuff.

macshome:
@ahoppen OK, I'm throwing in the towel on getting things to work as I expect on Windows. In the meantime I skipped the test for iterating in root on Windows. If someone who is better versed in Windows Foundation wants to take a swing at it that would be great.

All tests pass on macOS and Windows now, I've run `swift format` on the files I changed.

### Issues I'm running into on Windows
After digging around in the Foundation sources it seems like `FileManager` should be doing all the right things on Windows. I made a simple CLI tool to make the calls and validate that.
```swift
// The Swift Programming Language
// https://docs.swift.org/swift-book
import WinSDK
import Foundation

var pwd = FileManager.default.currentDirectoryPath
print(pwd) //Output: C:\Users\macsh\Swift\pwd\pwd

var isRoot = pwd.withCString(encodedAs: UTF16.self, PathCchIsRoot)
print(isRoot) //Output: false

FileManager.default.changeCurrentDirectoryPath("/")
pwd = FileManager.default.currentDirectoryPath
isRoot = pwd.withCString(encodedAs: UTF16.self, PathCchIsRoot)

print(pwd) //Output: C:\
print(isRoot) //Output: true
``` 

That looks great! So I made a simple extension to `URL` and added a `isRoot` property to check for. That also works on Windows, but when I run the tests I end up with a bunch of extra `/` in the paths. Instead of the nice path strings from above I end up with things like `/C:/Users/macsh/AppData/Local/Temp/TemporaryItems/(A document being saved by swiftformatPackageTestxctest 24` which are clearly incorrect for a number of reasons. I think that it's getting tripped up by the test runner, but I don't know enough on the Windows tooling to figure it out in a timely manner.

compnerd:
Without checking the implementation, the thing that comes to mind is incorrect API usage. `url.path` will generally give you a Unix style path (`/C:/`) vs. `url.withUnsafeFileSystemRepresentation { String(cString: $0!) }` will give you a proper FSR (`C:\`). It could be that the `CustomStringConvertible` implementation goes through the latter. (As I've not verified that this is the case, this might just be entirely incorrect and is really pure speculation).

ahoppen:
Thanks for your work on this change @macshome. I looked at the getting the tests passing on Windows as well and these are the changes I came up with. https://github.com/ahoppen/swift-format/commit/fe056af49ac6ad03aca89ddca6a2214a6f284b74


Do you want to incorporate them into your PR?

macshome:
> Without checking the implementation, the thing that comes to mind is incorrect API usage. `url.path` will generally give you a Unix style path (`/C:/`) vs. `url.withUnsafeFileSystemRepresentation { String(cString: $0!) }` will give you a proper FSR (`C:\`). It could be that the `CustomStringConvertible` implementation goes through the latter. (As I've not verified that this is the case, this might just be entirely incorrect and is really pure speculation).

Yeah there are a lot of different callers flying around and I had the thought that maybe something was just taking a misstep somewhere. I'll take another look and see if I can reduce that particular issue more.

macshome:
> Thanks for your work on this change @macshome. I looked at the getting the tests passing on Windows as well and these are the changes I came up with. [ahoppen@fe056af](https://github.com/ahoppen/swift-format/commit/fe056af49ac6ad03aca89ddca6a2214a6f284b74)
> 
> Do you want to incorporate them into your PR?

Sure I can pull those changes in and then update my PR.

macshome:
@ahoppen I synced up with `main` and pulled in your changes to update the PR.

macshome:
> @macshome Instead of merging main into your branch, could you rebase your branch? Otherwise LGTM.

Rebase onto your branch or onto main?

ahoppen:
I meant to rebase onto `main` so that your branch doesn’t contain any merge commits.

macshome:
Since I had been clicking the update button in GitHub along the way, there are a few merges to try to get rid of. It's probably easiest to just make a clean branch from main and then rebase the changes onto it. I'll need to create a new PR, but at least we can get the changes in.

ahoppen:
You can rebase your changes and then force-push your branch to update this PR. There’s no need to create a new branch or open a new PR.

macshome:
OK, I'll give that a spin. I figured a force-push was a non-starter.

macshome:
Alright, I was on PTO for a wedding for a while (Not mine, but it was out of town) and I will get back to cleaning up the PR now.

macshome:
@ahoppen Phew. git, am I right?

Hopefully the diff is now ready to test.

ahoppen:
It looks like the git history got a little confused and there’s still a merge conflict. Could you rebase your commits so there’s no merge conflict and maybe also squash your commits into one?

macshome:
I rebased interactively, fixed the conflict, and squashed into one commit. Each commit still shows on it's own also, I don't think that's an issue, but let me know if you want me to try and resolve that with a new PR linked to this one.

ahoppen:
Not entirely sure what went on with your branch but I just rebased and squashed the commits. Maybe things got more complicated because this PR is from the `main` branch of your fork and thus it wasn’t easy for you to rebase on top of swift-format’s `main` branch.

ahoppen (approved):

ahoppen:
I can take care of the CI failures

macshome:
> Not entirely sure what went on with your branch but I just rebased and squashed the commits. Maybe things got more complicated because this PR is from the `main` branch of your fork and thus it wasn’t easy for you to rebase on top of swift-format’s `main` branch.

Yeah, it ended up being more changes than I expected. I'll make sure to be clean and regular on the next one. 

Thanks for all the help!

PR #866, opened 2024-10-25
Prepare for integer generics with new generic argument node

Azoy:
With https://github.com/swiftlang/swift-syntax/pull/2859, there will be a new generic argument node that you need to switch over.

### DIFF
--- a/Sources/SwiftFormat/Rules/UseShorthandTypeNames.swift
+++ b/Sources/SwiftFormat/Rules/UseShorthandTypeNames.swift
@@ -47,12 +47,10 @@ public final class UseShorthandTypeNames: SyntaxFormatRule {
 
     switch node.name.text {
     case "Array":
-      guard let argument = genericArgumentList.firstAndOnly,
-            case .type(let typeArgument) = argument else {
+      guard case .type(let typeArgument) = genericArgumentList.firstAndOnly else {
         newNode = nil
         break
       }
-
       newNode = shorthandArrayType(
         element: typeArgument,
         leadingTrivia: leadingTrivia,
@@ -60,9 +58,8 @@ public final class UseShorthandTypeNames: SyntaxFormatRule {
       )
 
     case "Dictionary":
-      guard let arguments = exactlyTwoChildren(of: genericArgumentList),
-            case .type(let type0Argument) = arguments.0.argument,
-            caes .type(let type1Argument) = arguments.1.argument else {
+      guard case (.type(let type0Argument), .type(let type1Argument)) = exactlyTwoChildren(of: genericArgumentList)
+      else {
         newNode = nil
         break
       }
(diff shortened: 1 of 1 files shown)


Sources/SwiftFormat/Rules/UseShorthandTypeNames.swift
```diff
@@ -47,24 +47,28 @@ public final class UseShorthandTypeNames: SyntaxFormatRule {
 
     switch node.name.text {
     case "Array":
-      guard let typeArgument = genericArgumentList.firstAndOnly else {
+      guard let argument = genericArgumentList.firstAndOnly,
+            case .type(let typeArgument) = argument else {
```
ahoppen, line 51: I think you can fuse this into 

```suggestion
      guard case .type(let typeArgument) = genericArgumentList.firstAndOnly {
```

Same comment in most matching cases below

Sources/SwiftFormat/Rules/UseShorthandTypeNames.swift
```diff
@@ -47,24 +47,28 @@ public final class UseShorthandTypeNames: SyntaxFormatRule {
...
+        element: typeArgument,
         leadingTrivia: leadingTrivia,
         trailingTrivia: trailingTrivia
       )
 
     case "Dictionary":
-      guard let typeArguments = exactlyTwoChildren(of: genericArgumentList) else {
+      guard let arguments = exactlyTwoChildren(of: genericArgumentList),
+            case .type(let type0Argument) = arguments.0.argument,
+            caes .type(let type1Argument) = arguments.1.argument else {
```
ahoppen, line 65: Similar here, I think this can be 
 
```suggestion
      guard case (.type(let type0Argument), .type(let type1Argument)) = exactlyTwoChildren(of: genericArgumentList) else {
```

If I’m mistaken, this is good as-is.

Same one more time below.

ahoppen (commented):
Ignore the CI failures for now, we need to figure out how to support cross-PR testing, swift-format will get tested as part of your test invocation from https://github.com/swiftlang/swift-syntax/pull/2859. We just need to coordinate merging the PRs because bypassing the CI on this PR requires force-merge capabilities. 

But please address the format issues found by https://github.com/swiftlang/swift-format/actions/runs/11524662613/job/32085325124?pr=866

allevato (approved):
You'll probably also need to make some changes in `TokenStreamCreator` to make sure that parameter declarations like `struct S<let T: Int>` (or however it's spelled) get correct whitespace/line-wrapping behavior.

Unfortunately we don't have a good way for the formatter to pass experimental features down to the parser, so that might have to wait. For now, this seems fine (with @ahoppen 's suggested changes) to prevent CI failures from the breaking API change, though.

Issue #870, opened 2024-11-05
Ability To Turn Off Formatting For Subdirectory


samdeane:
I have a `.swift-format` file at the root of my repo, but some third-party code in a sub-directory. I'd like to be able to disable formatting for the subdirectory.

There are a couple of ways that this could be accomplished

- a setting in `.swift-format` with a pattern or patterns for paths to be excluded from formatting
- detecting the presence of a file `.no-swift-format` and disabling formatting anywhere below that file in the tree

The former case is useful, but would not be sufficient in the situation where you've a submodule with some third-party code, and that submodule has its own `.swift-format` file. In that situation the module's file would take precedence as it is closest, but you'd want to be able to override it locally without having to make changes to it. 

Detecting a `.no-swift-format` file at the same level would be efficient in this situation, and the file could be added to `.gitignore` for the module, so that creating it locally would not cause upstream problems.

samdeane:
See also [this thread](https://swift-open-source.slack.com/archives/C02PV8T5HQD/p1730608790736399?thread_ts=1730478877.221959&cid=C02PV8T5HQD)

ahoppen:
Synced to Apple’s issue tracker as rdar://139269456

samdeane:
Have added an (untested) draft implementation at #873. Am happy to try to clean it up and submit for proper review.

samdeane:
Now tested and submitted for review.

jasikpark:
Prior art exists in the swift-lang/github-workflows repo: https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/scripts/check-swift-format.sh

Allows nested folder ignores.

PR #873, opened 2024-11-07
Ability To Turn Off Formatting For Subdirectory

samdeane:
Draft implementation of #870.

This implementation adds a check for a `.swift-format-ignore` file, in the same places that a `.swift-format` file could be.

If present, this file potentially contains gitignore-style rules describing which files to ignore. 

Currently the only rule supported is `*`, to ignore everything in the containing directory and any subdirectories. Later, full pattern-matching rules could be implemented, if desired. For now, we simply check that the content of the file is `*`, to ease any future migration problems.

The presence of `.swift-format-ignore` is checked first at any given directory level. Even if a `.swift-format` file also exists at the same level, the ignore file takes precedence.

### Motivation

The use case which motivated this change is a scenario where my workspace has a `.swift-format` file, and I've turned on format-on-save, but I also have submodules containing third-party code.

If I make a small edit to a file in a submodule, it is formatted using the global workspace settings. This generates a large diff unrelated to my edit, as the submodule is hand-formatted using different rules.

I want to be able to disable formatting for specific submodules on my local machine, whilst retaining formatting for the rest of the workspace.


samdeane:
This implementation completely suppresses processing of files if `addDisabled` is true (or `.no-swift-format` is present).

On reflection, this isn't the correct behaviour I think, as it leads to no output when applied to one or more source files.

A better approach would be to output the input files unchanged. I will switch this back to a draft and fix the implementation.

allevato (commented):
Thanks for looking into this! I think it's a useful feature to have. Some high-level comments:

* I think the name of the file should be `.swift-format-ignore`, which aligns nicely with the name used in comments that the formatter supports, and also similar tools like `.gitignore` and `.prettierignore`.

* Lowering this down into the `SwiftFormatter` and `SwiftLinter` APIs feels like the wrong approach, because it means the tool is still doing all the work of iterating over and parsing all the files that are going to be ignored anyway.

For the second point, raising this check into the frontend seems more appropriate. Maybe `FileIterator` should be responsible for looking for the `.swift-format-ignore` file in the directory or its parents, and if one is found, it doesn't recurse into the directory further? That may end up being simpler to implement (you don't have to touch the configuration at all).

It means that somebody using the `SwiftFormatter`/`SwiftLinter` APIs directly wouldn't get that logic, but I think that's fine; those APIs are meant to be used by clients who want to format something directly and supplying a configuration in memory, and they're not interested in the search logic used by the frontend.

samdeane:
> I think the name of the file should be .swift-format-ignore

Yes, makes sense 👍

samdeane:
> For the second point, raising this check into the frontend seems more appropriate. Maybe FileIterator should be responsible for looking for the .swift-format-ignore

I think I was fixated on the idea that the ignoring mode needed to be configurable from within a `.swift-format`, as well as by the presence of a `.swift-format-ignore` file.  Having done it though, I struggle to see the utility of that setting.

I'll try your plan instead.

samdeane:
Ok, I got rid of the setting and moved the check into FileIterator's directory handling. 

It works well for input directories, and for subdirectories reached via recursion, but doesn't work for input files (e.g: in the case where individual file paths are specified on the command line directly). 

For input files, we need to perform a check for `.swift-format-ignore` in the same directory as the file. Putting that into the iterator's next() feels wrong as it's not needed in most cases. My inclination is to filter the input urls in `init`, and perform this test on any that are files.

allevato:
> For input files, we need to perform a check for `.swift-format-ignore` in the same directory as the file. Putting that into the iterator's next() feels wrong as it's not needed in most cases. My inclination is to filter the input urls in `init`, and perform this test on any that are files.

That makes the most sense to me.

samdeane:
Ok, 'tis a little closer now.

One question is what it should do if `--configuration <file>` is passed in. 

The README still says that the filesystem won't be searched; which is true for `.swift-format`, but no longer true for `.swift-format-ignore`. 

The path of least resistance is to clarify the README, but that depends whether that's the behaviour we want.

Alternatively I'll need to add a flag to FileIterator to indicate whether or not it should check for the ignore file. Which is perfectly doable, of course.

allevato:
> One question is what it should do if `--configuration <file>` is passed in.
> 
> The README still says that the filesystem won't be searched; which is true for `.swift-format`, but no longer true for `.swift-format-ignore`.

The implication of that statement is that the filesystem won't be searched for the configuration (i.e., `.swift-format` file) specifically. We don't need it to apply to the `.swift-format-ignore` file.

If we want a way to ignore the ignore file, we could add something like a `--force` option to the command line interface, but I don't think we need it today.

README.md
```diff
@@ -218,6 +222,9 @@ See [Documentation/Configuration.md](Documentation/Configuration.md) for a
 description of the configuration file format and the settings that are
 available.
 
+Note that the contents of any `.swift-format-ignore`  are ignored; the
+file can be empty.
```
allevato, line 226: Thinking about this further, we may eventually want to allow users to specify patterns here like other tools' ignore files. So we should make sure that the behavior that we support today would upgrade cleanly to something more elaborate in the future. If we allow the file to be empty today and then decide that we want to allow patterns, that becomes an awkward path because an empty file ought to mean "ignore nothing" in that world.

The most obvious thing that comes to mind is for us to currently require that the contents of the `.swift-format-ignore` file are exactly `*` (modulo whitespace). Then anyone who wants to ignore a directory could put that in their file and it would continue to work later if we start expanding it to more complex patterns. When we find the file, we could open it and error out if the file contains anything other than `*`, telling users that that's the only supported option today.

So I would remove this paragraph and change the one you added further up to read something like this instead:

> If it finds a file named `.swift-format-ignore`, its contents will determine which files in that directory should be ignored by swift-format. Currently, the only supported option is `*` (ignores all files).

samdeane:
> If we want a way to ignore the ignore file, we could add something like a --force option to the command line interface, but I don't think we need it today.

Yeah, I was thinking the same. I'll reword the docs.

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -169,3 +179,23 @@ private func fileType(at url: URL) -> FileAttributeType? {
...
+private func inputShouldBeProcessed(at url: URL) -> Bool {
+  guard fileType(at: url) != .typeDirectory else {
+    return true
+  }
+      
+  var containingDirectory = url.absoluteURL.standardized
+  repeat {
+    containingDirectory.deleteLastPathComponent()
+    let candidateFile = containingDirectory.appendingPathComponent(FileIterator.ignoreFileName)
+    if FileManager.default.isReadableFile(atPath: candidateFile.path) {
```
allevato, line 196: More than just checking for the presence of the file, I think we go the extra step of validating that it only contains `*` so that users aren't surprised later when we add more functionality and it subtly changes the behavior based on those contents. If it contains anything else, we should throw an error saying it's the only supported form (and turn that into a nice human-readable error message when we exit).

andre-richter: If we want to upgrade to gitignore like sytanx in the future, should we make the default value `**/*`?

allevato: AFAIK, `*` is sufficient because according to the [documentation](https://git-scm.com/docs/gitignore):

* "An asterisk '`*`' matches anything except a slash."
* "If there is a separator at the beginning or middle (or both) of the pattern, [...]. Otherwise the pattern may also match at any level below the `.gitignore` level."
* "If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories."

samdeane: Sorry, got distracted, but am looking at this now.

One complication is that FileIterator currently doesn't throw, and instead relies on the caller to generate diagnostics (eg for non-existent paths). 

In order to fully validate the ignore file inside FileIterator we'd either need to be able to throw, or be able to emit diagnostics, or we'd need to spot an invalid ignore file and return its path (or a special pseudo-path), then spot that path in the front end and emit the diagnostic there. All of which is doable but a bit ugly.

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,41 @@
...
+
+  func testInvalidIgnoreFile() throws {
+    let url = Bundle.module.url(forResource: "invalid", withExtension: "", subdirectory: "Ignore Files")
+    XCTAssertNotNil(url)
+    XCTAssertThrowsError(try IgnoreFile(forDirectory: url!))
+    XCTAssertThrowsError(try IgnoreFile(for: url!.appending(path:"file.swift")))
+  }
+
+  func testEmptyIgnoreFile() throws {
+    let url = Bundle.module.url(forResource: "empty", withExtension: "", subdirectory: "Ignore Files")
```
ahoppen, line 28: I am personally not a fan of test resources, especially if they are pretty small like for these test cases because it means that you can’t reason about the test case by just looking at the test case anymore but need to inspect the resource files. Could you change the tests to write these ignore files to a temporary directory from a string literal instead?

Edit: If we have an initializer of `IgnoreFile` that takes a string, we might not  need to write the ignore files to the file system in some of these tests.

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,93 @@
...
+  func shouldProcess(_ url: URL) -> Bool {
+    return false
+  }
+
+  /// Returns true if the name of the given URL matches 
+  /// the standard ignore file name.
+  public static func isStandardIgnoreFile(_ url: URL) -> Bool {
+    return url.lastPathComponent == fileName
+  }
+}
```
ahoppen, line 93: Could you add trailing newlines to these files? I think swift-format will also complain about incorrect formatting if there is no trailing newline.

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,93 @@
...
+    case invalidContent(URL)
+  }
+
+  /// Create an instance from the contents of the file at the given URL.
+  /// Throws an error if the file content can't be read, or is not valid.
+  public init(contentsOf url: URL) throws {
+    let content = try String(contentsOf: url, encoding: .utf8)
+      guard content.trimmingCharacters(in: .whitespacesAndNewlines) == "*" else {
+        throw Error.invalidContent(url)
+      }
```
ahoppen, line 39: This formatting looks off to me. Could you run swift-format on the swift-format sources themselves?

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,93 @@
...
+
+  /// Create an instance for the given directory, if a valid
+  /// ignore file with the standard name is found in that directory.
+  /// Returns nil if no ignore file is found.
+  /// Throws an error if an invalid ignore file is found.
+  ///
+  /// Note that this initializer does not search parent directories for ignore files.
+  public convenience init?(forDirectory directory: URL) throws {
+    let url = directory.appendingPathComponent(IgnoreFile.fileName)
+    guard FileManager.default.isReadableFile(atPath: url.path) else {
```
ahoppen, line 50: Instead of doing two file system accesses to determine if the file is readable and then reading its contents, what do you think about changing `init(contentsOf:)` to take the file contents as a string? Then we could read the file contents here once and pass the contents to the `init(contentsOf:)` initializer (which would need a name matching its new behavior).

It also means that we wouldn’t need to write the ignore files to disk in the tests anymore.

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,93 @@
...
+  /// in the directory tree.
+  ///
+  /// Note that we start the search from the given URL's **container**,
+  /// not the URL itself; the URL passed in is expected to be for a file.
+  /// If you pass a directory URL, the search will not include the contents
+  /// of that directory.
+  public convenience init?(for url: URL) throws {
+      var containingDirectory = url.absoluteURL.standardized
+      repeat {
+        containingDirectory.deleteLastPathComponent()
```
ahoppen, line 71: I think we infinite loop if this is started on `/` because the parent directory of `/` is `/..`, which wouldn’t be considered a root. Anyway, would be good to have a test for that.

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -83,6 +83,21 @@ public struct FileIterator: Sequence, IteratorProtocol {
...
+              continue
+            }
+          } catch IgnoreFile.Error.invalidContent(let url) {
+            // we hit an invalid ignore file
+            // we skip the directory, but return the path of the ignore file
+            // so that we can report an error
+            output = url
+          } catch {
+            // we hit another unexpected error; just skip the directory
+            continue
```
ahoppen, line 98: I don’t think that silently skipping a directory from formatting is the right answer here because it gives a false confidence that source files are formatted correctly. I think we should either expose the error to the user or not skip the directory.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -165,6 +165,13 @@ class Frontend {
   /// Read and prepare the file at the given path for processing, optionally synchronizing
   /// diagnostic output.
   private func openAndPrepareFile(at url: URL) -> FileToProcess? {
+    guard !IgnoreFile.isStandardIgnoreFile(url) else {
+      diagnosticsEngine.emitError(
```
ahoppen, line 169: Could you add a test case that ensures this diagnostic gets produced?

Sources/SwiftFormat/Core/IgnoreFile.swift
samdeane: Really the only reason for doing two file accesses is to distinguish between the file not existing and other error cases; for clarity it seemed cleaner to do an explicit check. 

We can do that by catching and screening the thrown exception instead if you're that concerned about a double file system access.

Sources/swift-format/Frontend/Frontend.swift
samdeane: Do you have existing code to test the front end output over a directory?
If you do, then sure.
If not, I think adding that code is outside the scope of this change and would be better off as a follow-up issue.

Sources/SwiftFormat/Utilities/FileIterator.swift
samdeane: Ok, I've made it just carry on. It's tricky to expose the error here due to the design of FileIterator.

Sources/SwiftFormat/Core/IgnoreFile.swift
samdeane: In use it ought to only be passed a file URL (ie not root itself), but I've added an initial guard to make sure we're not given root. Could be an assertion instead I guess.

Sources/SwiftFormat/Core/IgnoreFile.swift
samdeane: I've added a String init and eliminated the second access, fwiw.

samdeane:
Actually I've realised that the current design won't work as written, because when FileIterator encounters a directory, it collapses all subdirectories below it, and therefore will not check for the presence of nested .swift-format-ignore files.

FileIterator will need to be rewritten to check for an ignore file when entering a directory, and to pass that instance down to any subdirectories recursively.

samdeane:
I'm happy to do this, but there are a couple ofdesign questions:

- If we flow through the directory structure, inheriting the ignore settings from the parent directory, how should we treat sym links to other directories? Should they inherit their settings from their actual parent, or their logical parent in the context of the traversal?
- We we're traversing a file tree and looking for `.swift-format-ignore` files, I wonder if we should also be looking for `.swift-format` files, and producing a (URL, Config) tuple for each element. It would be a lot more efficient than search for a nested config whilst processing every file.

samdeane:
Slightly annoyingly, my original naive implementation was simple and worked fine, since it wasn't the front end doing the work. 

I don't think it would be unreasonable to go with that simpler approach as the first iteration, and then refine it subsequently -- perfect being the enemy of good and all that. That said, I'm happy to look at reworking FileIterator instead.

samdeane:
In other news, the current main branch will not build with my version of Swift:

`swift-driver version: 1.115.1 Apple Swift version 6.0.3 (swiftlang-6.0.3.1.9 clang-1600.0.30.1)
Target: arm64-apple-macosx15.0`

```
/Users/sam/Developer/Projects/swift-format/Sources/SwiftFormat/Rules/UseShorthandTypeNames.swift:50:78: error: type 'TypeSyntax?' has no member 'type'
 48 |     switch node.name.text {
 49 |     case "Array":
 50 |       guard case .type(let typeArgument) = genericArgumentList.firstAndOnly?.argument else {
    |                                                                              `- error: type 'TypeSyntax?' has no member 'type'
 51 |         newNode = nil
 52 |         break
```

(and a bunch of other errors)

samdeane:
fwiw git bisect tells me that the problematic commit is c3b0c9f

allevato:
Doing this in the frontend is still the right approach; the API layer shouldn't care about `.swift-format-ignore` files just as it doesn't care about `.swift-format` configuration files. Implementing it in the API layer as an incremental step would be putting us into a worse state than not having it, design-wise.

> If we flow through the directory structure, inheriting the ignore settings from the parent directory, how should we treat sym links to other directories? Should they inherit their settings from their actual parent, or their logical parent in the context of the traversal?

I think the answer has to be the actual (physical) parent. Otherwise you could end up with different ignore states for the same file depending on how you got to it, which wouldn't be desirable. This does make the implementation slightly more complicated, but I think it would be workable to have `FileIterator` maintain a cache of some sort to keep track of the ignore-specs found in each directory it visits. When you visit a directory you look in the cache to see if it or any of its parents have loaded their ignore-specs, and in most cases, this will be fast because you got to that directory from its immediate parent. In the case where you follow a symlink into a different directory tree, you might need to look up all the way toward the root to find the ignore-specs (or know that they don't exist), but that would be rarer.

> We we're traversing a file tree and looking for .swift-format-ignore files, I wonder if we should also be looking for .swift-format files, and producing a (URL, Config) tuple for each element. It would be a lot more efficient than search for a nested config whilst processing every file.

This sounds like a nice optimization! We flatten the iterator into an array in the frontend to do parallel processing so we don't want to have the full `Configuration` stored with each file URL (it would be identical across all files in the majority of cases), but returning a pair with the URL of the file to format and the URL of the relevant configuration would still be an improvement, I think. Then we can still leverage the existing configuration cache to load them on-demand, without the search each time.

The same caveat about which parent to use applies here as above. For symlinks, we'd want the configuration we choose to be the one from the physical parent, not the logical parent, because we don't want files to be formatted differently based on how you got there.

> In other news, the current main branch (https://github.com/swiftlang/swift-format/commit/ed010288eb9c16b6280198aa2c31ac5d9d622271, without my changes) will not build with my version of Swift:

Is your swift-syntax checkout out of date? The change you referenced later was made to support an API change merged a few days ago in swift-syntax.

Sources/swift-format/Frontend/Frontend.swift
theangelperalta: I'm definitely invested in seeing this finished. Is there a convention for the new directory where this new test would be included (`./Tests/swift-formatTests`)? It seems like  we should rename the `swift-format` directory in `Sources` to match [`swift-package-manager` ](https://github.com/swiftlang/swift-package-manager/tree/main/Sources/Commands)structure `Sources/Commands`. @ahoppen, what are your thoughts on this?

allevato: We used to have a `swift-formatTests` for the frontend, but unfortunately Windows doesn't support test targets that depend on a binary target. 

Maybe we should factor out the entire frontend into a separate library target and just have the binary target be nothing but a `@main` that calls it, but I definitely don't want to make that a prerequisite for this change. It's unfortunate that this can't just be made to work on Windows.

theangelperalta: @allevato Thanks for the context on this. How should we precede with this change?

ahoppen: Let’s continue without having a test for it for now. Factoring out the entire frontend into a separate library is out-of-scope for this PR.

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,109 @@
...
+  /// Note that we start the search from the given URL's **container**,
+  /// not the URL itself; the URL passed in is expected to be for a file.
+  /// If you pass a directory URL, the search will not include the contents
+  /// of that directory.
+  public convenience init?(for url: URL) throws {
+    guard !url.isRoot else {
+      return nil
+    }
+
+    var containingDirectory = url.absoluteURL.standardized
```
ahoppen, line 91: Do we need to standardize the URL here? Say we have a directory structure like

```
/
  LibA
    LibA.swift
  LibB
    Externals
      .swift-format-ignore = "*"
      LibA -> /LibA
```

Then, when running swift-format on `/LibB/Externals/LibA/LibA.swift`, I would expect formatting to be skipped because of `/LibB/Externals/.swift-format-ignore`. But with the current logic, we would standardize `/LibB/Externals/LibA/LibA.swift` to `/LibA/LibA.swift` in here and thus not find the `.swift-format-ignore` file.

Sources/SwiftFormat/Core/IgnoreFile.swift
```diff
@@ -0,0 +1,109 @@
...
+    repeat {
+      containingDirectory.deleteLastPathComponent()
+      let url = containingDirectory.appendingPathComponent(IgnoreFile.standardFileName)
+      if FileManager.default.isReadableFile(atPath: url.path) {
+        try self.init(contentsOf: url)
+        return
+      }
```
ahoppen, line 98: We should be able to skip this file system access and just check if `self.init(contentsOf:)` returns a non-nil value. Just trying to avoid race conditions where the file exists here but is gone by the time that `self.init(contentsOf:)` tries to read it.

```suggestion
      if let ignoreFile = try self.init(contentsOf: url) {
        self = ignoreFile
        return
      }
```

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -92,6 +92,20 @@ public struct FileIterator: Sequence, IteratorProtocol {
...
+              continue
+            }
+          } catch IgnoreFile.Error.invalidFile(let url, _) {
+            // we hit an invalid ignore file
+            // we return the path of the ignore file so that we can report an error
+            // and process the directory as normal
+            output = url
+          } catch {
+            // we hit another unexpected error; process the directory as normal
+          }
```
ahoppen, line 107: Why do we need to set `output = url` in the `IgnoreFile.Error.invalidFile` case but not in the generic error case?

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -179,3 +193,34 @@ private func fileType(at url: URL) -> FileAttributeType? {
...
+private func inputShouldBeProcessed(at url: URL) -> Bool {
+  guard fileType(at: url) != .typeDirectory else {
+    return true
+  }
+
+  let ignoreFile = try? IgnoreFile(for: url)
+  return ignoreFile?.shouldProcess(url) ?? true
+}
+
+fileprivate extension URL {
```
ahoppen, line 214: Could we use `isRoot` from `URL+isRoot.swift` instead of defining a new implementation of it here?

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,126 @@
+@_spi(Internal) import SwiftFormat
```
ahoppen, line 1: Could you add a copyright header to this test file?

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -165,6 +165,13 @@ class Frontend {
   /// Read and prepare the file at the given path for processing, optionally synchronizing
   /// diagnostic output.
   private func openAndPrepareFile(at url: URL) -> FileToProcess? {
+    guard url.lastPathComponent != IgnoreFile.standardFileName else {
```
ahoppen, line 168: This way of reporting errors from invalid `.swift-format-ignore` files seems a little roundabout to me. What do you think about gathering errors in `FileIterator` inside a `var diagnostics: [Diagnostic]` member. Then inside `Frontend.processURLs` where we construct the `FileIterator`, we can consult the diagnostics produced by it and report them. That way we should also be able to test that `FileIterator` correctly emits these diagnostics.

Tests/SwiftFormatTests/Resources/Ignore Files/nested/.swift-format-ignore
ahoppen, line 1: Is this file still needed?

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,126 @@
...
+    case file(String, String)
+    case directory(String, [TestTree])
+  }
+
+  override func tearDown() {
+    // Clean up any test tree after each test.
+    if let testTreeURL {
+      // try? FileManager.default.removeItem(at: testTreeURL)
+    }
+  }
```
ahoppen, line 18: No longer needed?

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,126 @@
...
+    if let testTreeURL {
+      try? FileManager.default.removeItem(at: testTreeURL)
+    }
+    let tempDir = FileManager.default.temporaryDirectory
+    let tempURL = tempDir.appendingPathComponent(UUID().uuidString)
+    try FileManager.default.createDirectory(at: tempURL, withIntermediateDirectories: true)
+    try writeTree(tree, to: tempURL)
+    testTreeURL = tempURL
+    return tempURL
+  }
```
ahoppen, line 34: What do you think about making this a `withTempTree` function that deletes the temporary directory after it executes the tests? Otherwise these trees will be left as artifacts. Ie something like

```suggestion
  private func withTempTree(_ tree: TestTree, _ body: (URL) throws -> Void) throws {
    let tempURL = FileManager.default.temporaryDirectory.appendingPathComponent("swift-format-test-scratch").appendingPathComponent(UUID().uuidString)
    defer {
      try? FileManager.default.removeItem(at: tempURL)
    }
    try FileManager.default.createDirectory(at: tempURL, withIntermediateDirectories: true) // I would move this to `writeTree`
    try writeTree(tree, to: tempURL)
    body(tempURL)
  }
```

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,126 @@
...
+
+  /// Write a file or directory tree to the given root URL.
+  func writeTree(_ tree: TestTree, to root: URL) throws {
+    switch tree {
+    case let .file(name, contents):
+      print("Writing file \(name) to \(root)")
```
ahoppen, line 40: Leftover debug output?

Tests/SwiftFormatTests/Core/IgnoreFileTests.swift
```diff
@@ -0,0 +1,126 @@
...
+    let tempDir = FileManager.default.temporaryDirectory
+    let tempURL = tempDir.appendingPathComponent(UUID().uuidString)
+    try FileManager.default.createDirectory(at: tempURL, withIntermediateDirectories: true)
+    try writeTree(tree, to: tempURL)
+    testTreeURL = tempURL
+    return tempURL
+  }
+
+  /// Write a file or directory tree to the given root URL.
+  func writeTree(_ tree: TestTree, to root: URL) throws {
```
ahoppen, line 37: I would make this private

```suggestion
  private func writeTree(_ tree: TestTree, to root: URL) throws {
```

ahoppen (commented):
Sorry for the delay since my last review and thanks a lot for continuing to push this forward. I like the overall design and like the new tests that create the test files in a temporary directory. I left a few comments inline.

bkhouri:
I posted an alternative change under #1156.

samdeane:
> I posted an alternative change under #1156.

Sounds good. I dropped this in the end, not because I didn't need it, but because the level of review feedback was getting into perfect-is-the-enemy-of-good territory for me, and honestly, I didn't have the time. My initial change worked, but wasn't in the best place architecturally. Moving it elsewhere highlighted a different weakness in the architecture, and I got to the point where I felt that the scope had crept too far.

To be clear, it's 100% the prerogative of the repo owners to dictate what goes in, and I don't have a problem with that.

PR #876, opened 2024-11-08
Add `--enable-experimental-feature` to enable those features in the parser.

### LINKED ISSUE #875: Add cli options for passing experimental features to swift-syntax
See title, when trying to format code using value generics, I had to patch the formatter to pass `.valueGenerics` to swift-syntax instead of being able to use a CLI flag.

allevato:
Also add a couple small tests for value generics to exercise the capability in tests, and I took the opportunity to clean up some error handling with the addition of this new user-facing error when an experimental feature isn't recognized.

I'm passing the experimental features through as strings deeper than I'd like, but trying to resolve them to the `Parser.ExperimentalFeatures` type higher in the stack causes more problems since that type is behind an `@_spi`. To prevent SPI leakage in swift-format's public interfaces (and thus require those interfaces to also have their own `@_spi`-restricted overloads), just leave them as a `Set<String>` until we call `Parser.parse` directly.

This PR requires https://github.com/swiftlang/swift-syntax/pull/2895.

Fixes #875.

### DIFF
--- a/Sources/SwiftFormat/Core/Parsing.swift
+++ b/Sources/SwiftFormat/Core/Parsing.swift
@@ -13,7 +13,7 @@
 import Foundation
 import SwiftDiagnostics
 import SwiftOperators
-import SwiftParser
+@_spi(ExperimentalLanguageFeatures) import SwiftParser
 import SwiftParserDiagnostics
 import SwiftSyntax
 
@@ -25,10 +25,14 @@ import SwiftSyntax
 ///
 /// - Parameters:
 ///   - source: The Swift source code to be formatted.
+///   - operatorTable: The operator table to use for sequence folding.
 ///   - url: A file URL denoting the filename/path that should be assumed for this syntax tree,
 ///     which is associated with any diagnostics emitted during formatting. If this is nil, a
 ///     dummy value will be used.
-///   - operatorTable: The operator table to use for sequence folding.
+///   - experimentalFeatures: The set of experimental features that should be enabled in the parser.
+///     These names must be from the set of parser-recognized experimental language features in
+///     `SwiftParser`'s `Parser.ExperimentalFeatures` enum, which match the spelling defined in the
+///     compiler's `Features.def` file.
 ///   - parsingDiagnosticHandler: An optional callback that will be notified if there are any
 ///     errors when parsing the source code.
 /// - Throws: If an unrecoverable error occurs when formatting the code.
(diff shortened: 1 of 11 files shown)


Sources/SwiftFormat/Core/Parsing.swift
```diff
@@ -25,22 +25,36 @@ import SwiftSyntax
...
+    guard let featureValue = Parser.ExperimentalFeatures(name: featureName) else {
+      throw SwiftFormatError.unrecognizedExperimentalFeature(featureName)
+    }
+    experimentalFeaturesSet.formUnion(featureValue)
+  }
+  var source = source
+  let sourceFile = source.withUTF8 { sourceBytes in
+    operatorTable.foldAll(Parser.parse(source: sourceBytes, experimentalFeatures: experimentalFeaturesSet)) { _ in }.as(
+      SourceFileSyntax.self
+    )!
```
ahoppen, line 56:
```suggestion
    operatorTable.foldAll(Parser.parse(source: sourceBytes, experimentalFeatures: experimentalFeaturesSet)) { _ in }
    .as(SourceFileSyntax.self)!
```

Reads better, I think.

Tests/SwiftFormatTests/PrettyPrint/ValueGenericsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+      > {
+        static let bar = n
+      }
+
+      """
+    assertPrettyPrintEqual(
+      input: input,
+      expected: expected,
+      linelength: 20,
+      experimentalFeatures: ["valueGenerics"]
```
ahoppen, line 16: Should we use `ExperimentalFeatures` enum cases here instead of spelling out the feature names?

Sources/SwiftFormat/Core/Parsing.swift
allevato: Done.

Tests/SwiftFormatTests/PrettyPrint/ValueGenericsTests.swift
allevato: Yeah, I guess that's fine for tests since we don't have to be worried about SPI leakage there. Updated.

Sources/_SwiftFormatTestSupport/Parsing.swift
```diff
@@ -0,0 +1,34 @@
...
+/// Parses the given source string and returns the corresponding `SourceFileSyntax` node.
+///
+/// - Parameters:
+///   - source: The source text to parse.
+///   - experimentalFeatures: The set of experimental features that should be enabled in the parser.
+@_spi(Testing)
+public func parseForTesting(
+  source: String,
+  experimentalFeatures: Parser.ExperimentalFeatures
+) -> SourceFileSyntax {
```
ahoppen, line 26: Do we need this still? I think there’s a `Parser.parse` function that does exactly this now that we don’t need to map experimental features anymore.

allevato: Since the `Parser.parse` overload that takes experimental features only takes a UTF-8 pointer, I mainly kept it to tidy up the handful of call sites in tests where we would otherwise have to repeat the "make a mutable copy of the string, convert it to a UTF-8 pointer, and then parse it" dance.

ahoppen: Oh, I see. I would make it an extension of `Parser` then.

allevato: Done.

Sources/swift-format/Subcommands/LintFormatOptions.swift
```diff
@@ -98,6 +98,15 @@ struct LintFormatOptions: ParsableArguments {
   )
   var followSymlinks: Bool = false
 
+  @Option(
```
rauhul, line 101: Sorry I didn't see hit PR go by, using `String` here is not great because the user has no idea what options are valid. Instead using a custom type and conforming it to `ExpressibleByArgument` and `CaseIterable` will allow `ArgumentParser` to provide a list of valid options in the help cli and error output.

allevato: As mentioned in the PR description, we would need `Parser.ExperimentalFeatures` to become non-SPI for that to be feasible, first. Otherwise, we can't pass the values down through the various API layers because they're `public` methods and thus would also need to be hidden behind their own SPI.

rauhul: Ah thats unfortunate, but good to know!

Issue #882, opened 2024-11-23
PrettyPrinter reports wrong line numbers


bkolb:
I have a file with a copyright header. This leads to the reporting of wrong line numbers in rules like `EndOfLineComment`

Here is an example where the reporting works as expected:
```
final class A {
    @Test func b() throws {
        doSomethingInAFunctionWithAVeryLongName()  1️⃣// Here we have a very long comment that should not be here because it is far too long
    }
}
```

Changing the input to the following will fail:
```
// Copyright (C) 2024 My Coorp. All rights reserved.
//
// This document is the property of My Coorp.
// It is considered confidential and proprietary.
//
// This document may not be reproduced or transmitted in any form,
// in whole or in part, without the express written permission of
// My Coorp.

final class A {
    @Test func b() throws {
        doSomethingInAFunctionWithAVeryLongName()  1️⃣// Here we have a very long comment that should not be here because it is far too long
    }
}
```

The error is here is reported in location (line:col 5:52) and should be in (line:col 12:52)

bkolb:
My suspicion is, that the pretty printer is thrown off by multiline comments. In that case, it seems to add only one line for an entire comment block.

PR #883, opened 2024-11-23
PrettyPrinter reports wrong line LineNumbersTests

bkolb:
This is to reproduce issue #882.

I am not sure if the fix is done in the right place. With my change every string is checked for new lines, which might be not necessary and comes at a slight cost. Maybe there is another, better place to fix this.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
@@ -118,7 +118,20 @@ struct PrettyPrintBuffer {
     writeRaw(text)
     consecutiveNewlineCount = 0
     pendingSpaces = 0
-    column += text.count
+
+    // In case of comments, we may get a multi-line string.
+    // To account for that case, we need to correct the lineNumber count.
+    // The new column is only the position within the last line.
+    let lines = text.split(separator: "\n")
+    lineNumber += lines.count - 1
+    if lines.count > 1 {
+      // in case we have inserted new lines, we need to reset the column
+      column = lines.last?.count ?? 0
+    } else {
+      // in case it is an end of line comment or a single line comment,
+      // we just add to the current column
+      column += lines.last?.count ?? 0
+    }
   }
 
   /// Request that the given number of spaces be printed out before the next text token.
--- a/Tests/SwiftFormatTests/PrettyPrint/LineNumbersTests.swift
+++ b/Tests/SwiftFormatTests/PrettyPrint/LineNumbersTests.swift
@@ -1,0 +1,84 @@
+import SwiftFormat
+import _SwiftFormatTestSupport
+
+final class LineNumbersTests: PrettyPrintTestCase {
+  func testLineNumbers() {
+    let input =
+      """
+      final class A {
+        @Test func b() throws {
+          doSomethingInAFunctionWithAVeryLongName()  1️⃣// Here we have a very long comment that should not be here because it is far too long
+        }
+      }
+      """
+
+    let expected =
+      """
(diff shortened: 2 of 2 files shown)


Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
```diff
@@ -118,7 +118,13 @@ struct PrettyPrintBuffer {
...
     consecutiveNewlineCount = 0
     pendingSpaces = 0
-    column += text.count
+
+    // In case of comments, we may get a multi-line string.
+    // To account for that case, we need to correct the lineNumber count.
+    // The new column is only the position within the last line.
+    let lines = text.split(separator: "\n")
+    lineNumber += lines.count - 1
+    column += lines.last?.count ?? 0
```
ahoppen, line 127: If there is a newline in the text, should we set `column` to `lines.last.count` instead of adding it? A newline effectively resets the column to 0 (see eg. `writeNewlines`).

ahoppen (commented):
Thanks for the PR @bkolb! The fix looks good to me, I just have one small comment.

Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
bkolb: Good point.
If there is 
* 0 new lines, we need to add
* 1 or more new lines, we need to set it to lines.last.count.

bkolb: I pushed an update.

ahoppen (approved):
Looks good. Thanks!

ahoppen:
@bkolb Could you run swift-format on your changes? The Windows failures will be fixed by https://github.com/swiftlang/github-workflows/pull/70.

bkolb:
> @bkolb Could you run swift-format on your changes? The Windows failures will be fixed by [swiftlang/github-workflows#70](https://github.com/swiftlang/github-workflows/pull/70).

Done

allevato (approved):
Thanks for digging into this! Turns out the problem was nowhere near where I was guessing it might be.

bkolb:
No problem :-) Thx for your reviews.

I don't understand the build failure on windows. Can you give me a hint?

ahoppen:
The Windows test failure is an infrastructure issue and will be fixed by https://github.com/swiftlang/github-workflows/pull/70.

ahoppen:
@bkolb This PR regressed swift-format’s performance by ~7% (https://github.com/swiftlang/swift-format/issues/894). Could you check if there’s a way to avoid that performance regression?

allevato:
That `split` call is doing a lot of work that gets thrown away. We don't actually use the substrings themselves, just the number of them and the length of the last one, so we have allocations on the critical printing path that we don't need. Replacing that with something that just counts the number of `\n`s and the length of the final segment would probably yield a good improvement.

ahoppen:
https://github.com/swiftlang/swift-format/issues/894 contains instruction how to reproduce the measurement.

macshome:
Changing the code to:
```swift
let lines = text.count { $0 == "\n" }
    lineNumber += lines
    guard lines > 0, let range = text.range(of: "\n", options: .backwards) else {
      // Handle case where no newline exists
      column += text.count
      return
    }
    let lastLine = text[range.upperBound...]
    column = lastLine.count
```

Brings the performance measurement to be much closer to the original. There may be a better way to get the last line though.

@bkolb I can open a PR for #894 if you are busy.

bkolb:
Thank you! I was just about to take a look. You have been faster :-)

Issue #844, opened 2024-10-10
Call into `PathCchIsRoot` to check if a URL points to a root directory 


ahoppen:
Once https://github.com/swiftlang/swift-foundation/issues/976 is fixed, we should use `PathCchIsRoot` to determine if a URL points to the root of a file system instead of counting the number of path components

ahoppen:
Synced to Apple’s issue tracker as rdar://137714365

PR #892, opened 2024-12-09
PathCchIsRoot Updated isRoot extension.

macshome:
Now that https://github.com/swiftlang/swift-foundation/issues/976 and https://github.com/swiftlang/swift-foundation/issues/980 are fixed we can clean this code up a bit by removing the empty path check on Linux and by using the native `PathCchIsRoot` on Windows.

The `isRoot` extension now uses native Windows and macOS API and has the Linux tests cleaned up.

- macOS root is determined by comparing the path to `NSOpenStepRootDirectory()`
- Windows root is determined by a call to `PathCchIsRoot`
- Linux root is determined by a comparison to `"/"`
- Non-filesystem type URLs always return false as they can't be a filesystem root.

Fixes: #844 

All tests pass on macOS and Windows 11.

### DIFF
--- a/Sources/SwiftFormat/Utilities/URL+isRoot.swift
+++ b/Sources/SwiftFormat/Utilities/URL+isRoot.swift
@@ -12,10 +12,32 @@
 
 import Foundation
 
+#if os(Windows)
+import WinSDK
+#endif
+
 extension URL {
+  /// Returns a `Bool` to indicate if the given `URL` leads to the root of a filesystem.
+  /// A non-filesystem type `URL` will always return false.
   @_spi(Testing) public var isRoot: Bool {
+    guard isFileURL else { return false }
+
+    #if compiler(>=6.1)
     #if os(Windows)
-    // FIXME: We should call into Windows' native check to check if this path is a root once https://github.com/swiftlang/swift-foundation/issues/976 is fixed.
+    let filePath = self.withUnsafeFileSystemRepresentation { pointer in
+      guard let pointer else {
+        return ""
+      }
+      return String(cString: pointer)
+    }
+    return filePath.withCString(encodedAs: UTF16.self, PathCchIsRoot)
+    #else  // os(Windows)
+    return self.path == "/"
+    #endif  // os(Windows)
+    #else  // compiler(>=6.1)
+
+    #if os(Windows)
+    // This is needed as the fixes from #844 aren't in the Swift 6.0 toolchain.
     // https://github.com/swiftlang/swift-format/issues/844
     var pathComponents = self.pathComponents
     if pathComponents.first == "/" {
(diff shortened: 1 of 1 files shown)


ahoppen:
We can’t switch this unconditionally yet because those fixes aren’t in Swift 6.0 toolchains. I think we need to guard this by `#if compiler(>=6.1)` and fall back to the old behavior if not.

macshome:
OK, I gated the new behavior behind a check for >=6.1. Below that the old code is still in place. The Mac behavior isn't gated as it most certainly isn't new. 😄

Sources/SwiftFormat/Utilities/URL+isRoot.swift
```diff
@@ -12,21 +12,43 @@
...
+    guard isFileURL else { return false }
+
+    #if os(macOS)
+    return self.path == NSOpenStepRootDirectory()
+    #endif
+
+    #if compiler(>=6.1)
     #if os(Windows)
-    // FIXME: We should call into Windows' native check to check if this path is a root once https://github.com/swiftlang/swift-foundation/issues/976 is fixed.
+    return self.path.withCString(encodedAs: UTF16.self, PathCchIsRoot)
```
ahoppen, line 31: `path` will be the URL path components. We should use `filePath` or `withUnsafeFileSystemRepresentation` instead.

Sources/SwiftFormat/Utilities/URL+isRoot.swift
```diff
@@ -12,21 +12,43 @@
...
+    #elseif os(Linux)
```
ahoppen, line 32: This should be `#else`. Otherwise this function doesn’t have a return value eg. on platforms that are neither macOS, Linux, nor Windows, like Android.

Sources/SwiftFormat/Utilities/URL+isRoot.swift
```diff
@@ -12,21 +12,43 @@
...
+  /// Returns a `Bool` to indicate if the given `URL` leads to the root of a filesystem.
+  /// A non-filesystem type `URL` will always return false.
   @_spi(Testing) public var isRoot: Bool {
+    guard isFileURL else { return false }
+
+    #if os(macOS)
+    return self.path == NSOpenStepRootDirectory()
+    #endif
+
+    #if compiler(>=6.1)
```
ahoppen, line 29: I would change this to `#elseif compiler(>=6.1)` to ensure we don’t get any of the returns below for macOS

Sources/SwiftFormat/Utilities/URL+isRoot.swift
macshome: I had that thought and will fix it.

macshome:
OK. I updated with the suggested changes, but note that I don't have my Windows setup with me right now. If someone can run the tests on Windows it would be great.

ahoppen (approved):

PR #895, opened 2024-12-11
Fix comment length computation

rauhul:
Resolved an issue where "end of line" comments matching the configured line length were incorrectly flagged as exceeding it.


allevato (approved):
Thank you for fixing this!

PR #897, opened 2024-12-12
Update .swift-format

rauhul:
Enables a number of rules that are already followed by this code base.


allevato (approved):
This works for me as long as @ahoppen is OK with it, since he would like to keep the configurations the same between swift-format and swift-syntax.

PR #899, opened 2024-12-13
Document default config values

rauhul:
Updates Configuration.md with a more structured config description and documents the default value for each config.


ahoppen (approved):
Thank you, Rauhul

Issue #894, opened 2024-12-10
#883 regressed performance by ~7% 


ahoppen:
https://github.com/swiftlang/swift-format/pull/883 regressed swift-format’s performance by 7-8%. To reproduce run

```bash
cd /tmp
git clone https://github.com/apple/swift-syntax.git
cd swift-syntax
git checkout -f 509.0.0

cd path/to/swift-format
swift build -c release
.build/release/swift-format lint --measure-instructions --recursive /tmp/swift-syntax
```

And compare the instruction count between swift-format with and without https://github.com/swiftlang/swift-format/pull/883.

ahoppen:
Synced to Apple’s issue tracker as rdar://141233035

macshome:
In case anyone was wondering, here is the difference.

Before #883: Instructions executed: 67167400341
With #883: Instructions executed: 71260352299

macshome:
Changing the code to:
```swift
let lines = text.count { $0 == "\n" }
    lineNumber += lines
    guard lines > 0, let range = text.range(of: "\n", options: .backwards) else {
      // Handle case where no newline exists
      column += text.count
      return
    }
    let lastLine = text[range.upperBound...]
    column = lastLine.count
```

Lowers the instructions executed to: 68853748601

macshome:
After playing around in Godbolt it seems that `lastIndex(of:_)` uses a few less instructions per-loop than `range(of:_)`. Changing the code to:
```swift
 let lines = text.count { $0 == "\n" }
    lineNumber += lines
    guard lines > 1, let lastNewlineIndex = text.lastIndex(of: "\n") else {
      // Handle case where no newline exists
      column += text.count
      return
    }
    let lastLine = text[text.index(after: lastNewlineIndex)...]
    column = lastLine.count
```

Further lowers the instruction executed count to: 68259258073

PR #906, opened 2024-12-20
Revert "PrettyPrinterPerformance Optimized the PrettyPrinter for #894"

bnbarham:
Reverts swiftlang/swift-format#901 which now computes the column on codepoints rather than characters.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrintBuffer.swift
@@ -122,17 +122,16 @@ struct PrettyPrintBuffer {
     // In case of comments, we may get a multi-line string.
     // To account for that case, we need to correct the lineNumber count.
     // The new column is only the position within the last line.
-    var lastLength = 0
-    // We are only interested in "\n" we can use the UTF8 view and skip the grapheme clustering.
-    for element in text.utf8 {
-      if element == UInt8(ascii: "\n") {
-        lineNumber += 1
-        lastLength = 0
-      } else {
-        lastLength += 1
-      }
+    let lines = text.split(separator: "\n")
+    lineNumber += lines.count - 1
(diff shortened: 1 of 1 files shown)


bnbarham:
CC @macshome (sorry, we'll get there!)

I haven't benchmarked this at all, but we could do something like:
```
    guard let lastNewlineRange = text.range(of: "\n", options: .backwards) else {
      // In case there's no newline, the string itself is the only line
      column += text.count
      return
    }

    let lastLine = text[lastNewlineRange.upperBound...]
    column += lastLine.count

    // Now count the rest of the lines in a possible multi-line string. We are only interested in '\n' so we can use
    // the UTF8 view and skip grapheme clustering entirely.
    for element in text[...lastNewlineRange.upperBound].utf8 {
      if element == UInt8(ascii: "\n") {
        lineNumber += 1
      }
    }
```

Which should hopefully result in a single iteration still (though I'm unsure of the slices). But basically - iterate backwards to find the last newline, grab column count during/from there, then iterate forwards up to that index to get the other newlines (which can still use the utf8 view).

Couple extra tests we could add:
```
  func testCharacterVsCodepoint() {
    let input =
      """
      let fo = 1  // 🤥
      
      """

    assertPrettyPrintEqual(
      input: input,
      expected: input,
      linelength: 16,
      whitespaceOnly: true,
      findings: []
    )
  }

  func testCharacterVsCodepointMultiline() {
    let input =
      #"""
      /// This is a multiline
      /// comment that is in 🤥
      /// fact perfectly sized
      
      """#

    assertPrettyPrintEqual(
      input: input,
      expected: input,
      linelength: 25,
      whitespaceOnly: true,
      findings: []
    )
  }
```

allevato (approved):

ahoppen:
@macshome Would you be interested in trying something like what @bnbarham suggested to count characters instead of codepoints?

PR #910, opened 2025-01-04
Fix NoEmptyLinesOpeningClosingBraces to respect consecutive newlines when a function starts or ends with a comment

### LINKED ISSUE #909: `NoEmptyLinesOpeningClosingBraces` should respect comments when determining which blanks lines are at the beginning/end of a function
The `NoEmptyLinesOpeningClosingBraces` swift-format option will remove blank lines from the  beginning/end of blocks, eg.

```
func myFunc() {

    var x = 3
    return x

}
```

is reformatted into:
```
func myFunc() {
    var x = 3
    return x
}
```

But comments are ignored, eg.
```
func myFunc() {
    // Some comment here

    // Do a thing
    var x = doAThing()

    // Some other comment here
}
```

ends up as:
```
func myFunc() {
    // Some comment here
    // Do a thing
    var x = doAThing()
    // Some other comment here
}
```

I would expect the input to remain unchanged in this particular case.

rdar://141589363

TTOzzi:
Resolve #909 

Since we are traversing the `leadingTrivia` in order to determine whether newlines are respected, we needed to modify the `trimmingSuperfluousNewlines` function to distinguish whether the `leadingTrivia` we are traversing is the beginning or end of a block.
For the `leadingTrivia` of the `closingBrace`, the handling of newlines was deferred and revised to process them only when the next `TriviaPiece` is neither a space/tab nor a newline.

### DIFF
--- a/Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
+++ b/Sources/SwiftFormat/Rules/NoEmptyLineOpeningClosingBraces.swift
@@ -69,7 +69,9 @@ public final class NoEmptyLinesOpeningClosingBraces: SyntaxFormatRule {
   }
 
   func rewritten(_ token: TokenSyntax) -> TokenSyntax {
-    let (trimmedLeadingTrivia, count) = token.leadingTrivia.trimmingSuperfluousNewlines()
+    let (trimmedLeadingTrivia, count) = token.leadingTrivia.trimmingSuperfluousNewlines(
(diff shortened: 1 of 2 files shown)


allevato (approved):
Looks reasonable to me, thanks!

allevato:
I'll kick off the full CI, which should at least make sure that this PR doesn't change the expected formatting of swift-syntax as a safety check.

@swift-ci please test

TTOzzi:
Thank you! 😄

PR #915, opened 2025-01-13
[Perf][WhitespaceLinter] Use hand crafted "is whitespace" function

rintaro:
* `UnicodeScalar(_:)` on arbitrary UTF8 code point was wrong. It only works correctly if the code point is < 0x80
* `UnicodeScalar.Properties.isWhitespace` is slow. Profiling `lint` showed it was taking 13.6% of the entire run time
* Whitespaces in Unicode "Basic Latin" block are well defined, there's no need to consult `UnicodeScalar.Properties`

Screenshot 2025-01-13 at 10 25 12 AM

```
swift-format lint --recursive /tmp/swift-syntax-509.0.0`
// Before
Instructions executed: 54264413578
// After
Instructions executed: 49121489746
```

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift
+++ b/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift
@@ -339,9 +339,16 @@ public class WhitespaceLinter {
     startingAt offset: Int,
     in data: [UTF8.CodeUnit]
   ) -> ArraySlice<UTF8.CodeUnit> {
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

allevato (approved):
Very nice improvement, thanks!

(I'm surprised that looking up a Boolean property in the Unicode tables that are baked into the runtime is *that* expensive, though...)

bnbarham (approved):

bnbarham:
Nice, thanks @rintaro!

PR #918, opened 2025-01-29
Improve the DontRepeatTypeInStaticProperties rule to better align with its original intent

### LINKED ISSUE #854: DontRepeatTypeInStaticProperties message doesn't match what it detect
1. Checkout https://github.com/apple/swift-protobuf @ c9fbc332ed293c64ff70af8d82367ca530dce11c
1. `swift format lint Sources/SwiftProtobuf/Message.swift` - Will produce a warning:

    > Sources/SwiftProtobuf/Message.swift:46:16: warning: [DontRepeatTypeInStaticProperties] remove the suffix 'Message' from the name of the variable 'protoMessageName'

The error message says to remove the _suffix 'Message'_, but the variable isn't a suffixed with 'Message'.  The check must be doing an `contains` on the name.  Not sure if the impl or the error message is wrong.

TTOzzi:
Resolve #854 

[DontRepeatTypeInStaticProperties](https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift#L41-L43) checks whether the type is included in the static property's name using contains. 
Since it triggers a diagnostic not only when the type appears as a suffix but also in other positions, I have appropriately modified the message.

### DIFF
--- a/Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
+++ b/Sources/SwiftFormat/Rules/DontRepeatTypeInStaticProperties.swift
@@ -26,19 +26,23 @@ public final class DontRepeatTypeInStaticProperties: SyntaxLintRule {
   /// type name (excluding possible namespace prefixes, like `NS` or `UI`) as a suffix.
   public override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
     guard node.modifiers.contains(anyOf: [.class, .static]),
-      let typeName = Syntax(node).containingDeclName
+      let typeName = Syntax(node).containingDeclName,
+      let variableTypeName = node.typeName,
+      typeName.hasSuffix(variableTypeName) || variableTypeName == "Self"
     else {
       return .visitChildren
     }
 
-    let bareTypeName = removingPossibleNamespacePrefix(from: typeName)
+    // the final component of the top type `A.B.C.D` is what we want `D`.
+    let lastTypeName = typeName.components(separatedBy: ".").last!
+    let bareTypeName = removingPossibleNamespacePrefix(from: lastTypeName)
     for binding in node.bindings {
       guard let identifierPattern = binding.pattern.as(IdentifierPatternSyntax.self) else {
         continue
       }
 
       let varName = identifierPattern.identifier.text
-      if varName.contains(bareTypeName) {
+      if varName.hasSuffix(bareTypeName) {
         diagnose(.removeTypeFromName(name: varName, type: bareTypeName), on: identifierPattern)
       }
     }
@@ -90,8 +94,7 @@ extension Syntax {
       case .identifierType(let simpleType):
         return simpleType.name.text
       case .memberType(let memberType):
-        // the final component of the top type `A.B.C.D` is what we want `D`.
-        return memberType.name.text
+        return memberType.description.trimmingCharacters(in: .whitespacesAndNewlines)
       default:
         // Do nothing for non-nominal types. If Swift adds support for extensions on non-nominals,
         // we'll need to update this if we need to support some subset of those.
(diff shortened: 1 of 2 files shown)


allevato (commented):
I don't think the message is the problem with this rule; the rule implementation itself is far too broad with what it detects. For example, consider one of the examples mentioned in that PR:

`remove the suffix 'Data' from the name of the variable 'postgresDataType'`

Presuming `Data` was the type of that property, that's an absolutely fine property name that doesn't violate this rule.

I don't remember precisely why we used `contains` in this implementation, unless it was a lackluster attempt at also catching plural terms. But at a minimum, I think we should be doing a suffix check instead of a contains check.

If you're really interested in exploring this, we could gain more insight by looking at how the compiler itself (in ClangImporter) stems type-name suffixes off the names of imported Objective-C names, and reimplement similar logic here to detect those cases.

TTOzzi:
> I don't think the message is the problem with this rule; the rule implementation itself is far too broad with what it detects. For example, consider one of the examples mentioned in that PR:
> 
> `remove the suffix 'Data' from the name of the variable 'postgresDataType'`
> 
> Presuming `Data` was the type of that property, that's an absolutely fine property name that doesn't violate this rule.
> 
> I don't remember precisely why we used `contains` in this implementation, unless it was a lackluster attempt at also catching plural terms. But at a minimum, I think we should be doing a suffix check instead of a contains check.
> 
> If you're really interested in exploring this, we could gain more insight by looking at how the compiler itself (in ClangImporter) stems type-name suffixes off the names of imported Objective-C names, and reimplement similar logic here to detect those cases.

Oh, I see. Thank you for your kind advice.
I’ll refer to the ClangImporter logic and reimplement it accordingly!

TTOzzi:
I looked into the suffix-related logic in the [ClangImporter](https://github.com/swiftlang/swift/tree/main/lib/ClangImporter), but I couldn't find anything particularly useful to add to the `DontRepeatTypeInStaticProperties` rule.

- Removal of the "Notification" suffix
- Removal of the "Error" suffix
- Removal of completion-handler suffixes
- Removal of the "Ref" suffix for CoreFoundation types
- Removal of the "Protocol" suffix

Most of these cases already seem to be handled by the existing logic.

Instead, I have modified the `DontRepeatTypeInStaticProperties` rule to better align with its description:
"Static properties of a type that return that type will yield a lint error."

TTOzzi:
Due to implementation limitations, the following case cannot be handled:
```swift
class Foo {
  static let defaultFoo = makeFoo()

  static func makeFoo() -> Foo {
    return Foo()
  }
}
```
However, aside from this, the rule now behaves in a way that aligns well with its description in most cases.

allevato (approved):
Thanks for the updates!

I did a little more digging and it looks like the code used by the compiler for this lives here (https://github.com/swiftlang/swift/blob/b221a92f695973f15ab26a63c2a9d7599f69435f/lib/Basic/StringExtras.cpp#L1252), and ClangImporter calls into that.

But that also looks like quite a lot of logic to bring over, and I'm not sure how much the logic would pay off compared to the effort it would take to port it over. What you've done here is already a solid improvement!

TTOzzi:
> Thanks for the updates!
> 
> I did a little more digging and it looks like the code used by the compiler for this lives here (https://github.com/swiftlang/swift/blob/b221a92f695973f15ab26a63c2a9d7599f69435f/lib/Basic/StringExtras.cpp#L1252), and ClangImporter calls into that.
> 
> But that also looks like quite a lot of logic to bring over, and I'm not sure how much the logic would pay off compared to the effort it would take to port it over. What you've done here is already a solid improvement!

Oh, I wasn’t familiar with C++ enough to have checked that file.
Although the PR has already been merged, I'll take a look, and if I find any areas for further improvement, I'll consider opening another PR.
Thank you for looking into this together 🙇

PR #923, opened 2025-02-06
Add dump-effective-configuration subcommand

### LINKED ISSUE #667: Add command to show the current configuration
Similar to:

```shell
$ git config -l
```

---

The proposed command would show the actual configuration being used.

With

.swift-format
```json
{
  "version": 1
}
```

```shell
$ swift-format show-configuration
```

would display

```
{
  "version": 1,
... default values ...
}
```

---

```shell
$ swift-format dump-configuration
```

displays the default configuration.

andre-richter:
closes #667

Implement the subcommand `dump-effective-configuration`, which dumps the configuration that would be used if `swift-format` was executed from the current working directory (cwd), incorporating configuration files found in the cwd or its parents, or input from the `--configuration` option.

This helps when composing a configuration or with configuration debugging/verification activities.


allevato (commented):
A high-level suggestion before I review this further: Since this is so similar to the existing `dump-configuration` command, I think this would be expressed more simply as an optional flag for that command instead of a completely separate command. Something like `dump-configuration --effective` would work and would probably cut down on the implementation complexity here.

andre-richter:
I thought about this initially, and wasn't happy that it would need to be explained that the `--configuration` option will only be valid in combination with the `--effective` flag, so I went with separate subcommand.

I'll give merging this into the `dump-config` subcommand a try and amend it to this PR.

andre-richter:
@allevato implemented your change request.

ahoppen:
I’m curious what the real-world use case for that options is. swift-format doesn’t have a complicated search algorithm to find the format files (it just walks up the directory structure) and it doesn’t merge files either. So, IIUC, all this effectively does is to find the first `.swift-format` file in the directory structure and print it. Am I missing something?

allevato:
> I’m curious what the real-world use case for that options is. swift-format doesn’t have a complicated search algorithm to find the format files (it just walks up the directory structure) and it doesn’t merge files either. So, IIUC, all this effectively does is to find the first .swift-format file in the directory structure and print it. Am I missing something?

I'm not sure of the PR author's motivation, but one thought I've had in the past is that if we ever make a breaking change to the configuration and bump the version number, it would be nice to have a way for users to explicitly update their configuration to the new version. One way to handle that would be an option like this: since older versions of the config would be translated to the new model in-memory during decoding, printing back that in-memory model afterwards would render it as the latest version.

andre-richter:
Sure, let me elaborate a bit on the motivation:

- Closing a feature gap with similar tools like `clang-format`. Or `git config -l` as mentioned by the author of #667. For both examples, the default behavior is actually what `--effective` would do here (I am not implying this is a better or worse default).

- What @allevato said. To extend on this: In general, having a quick way to verify that the configuration you're entering into the tool is actually being processed by the tool as intended is beneficial. This typically would be used while you're in the process of composing or modifying a `.swift-format` file or a `--configuration` JSON-string.

- Tools or scripts that wrap around `swift-format` likely pass the configuration to the tool using the `--configure` option. `--effective` would allow authors of such wrappers to easily add an option to print the final configuration that would be fed to `swift-format`, so that users get an option to easily retrieve it for review or re-use.

Sources/SwiftFormat/API/Configuration+Dump.swift
```diff
@@ -0,0 +1,39 @@
...
+
+extension Configuration {
+  /// Return the configuration as a JSON string.
+  public func asJsonString() throws -> String {
+    let data: Data
+
+    do {
+      let encoder = JSONEncoder()
+      encoder.outputFormatting = [.prettyPrinted]
+      if #available(macOS 10.13, *) {
```
ahoppen, line 23: We have a deployment target of macOS 12, so `.sortedKeys` should always be available.

Sources/swift-format/Subcommands/DumpConfiguration.swift
```diff
@@ -15,32 +15,55 @@ import Foundation
...
+
+      // Pretend to use stdin, so that the configuration loading machinery in the Frontend base class can be used in the
+      // next step. This produces the same results as if "format" or "lint" subcommands were called.
+      let lintFormatOptions = try LintFormatOptions.parse(["-"])
+
+      let frontend = DumpConfigurationFrontend(
+        configurationOptions: configurationOptions,
+        lintFormatOptions: lintFormatOptions
+      )
+      frontend.run()
```
ahoppen, line 57: I’m not a big fan of running a dummy frontend that stores the configuration used as a side effect. I’d prefer to just call into `Frontend.configuration(fromPathOrString:orInfoerredFromSwiftFileAt:)` instead.

ahoppen (commented):
Thanks for your explanation of the use case @andre-richter. The change makes sense to me. I have a few comments regarding the implementation inline.

Sources/swift-format/Subcommands/DumpConfiguration.swift
ahoppen: Yes, looks like it shouldn’t be too hard. It looks like `ConfigurationLoader` can be a local variable instead of a member and `DiagnsoticsEngine` can be passed in.

andre-richter:
@ahoppen, there was also `checkForUnrecognizedRules()`, which also employs a `DiagnosticsEngine`. Since `Frontend.configuration()` was the only user, passing `DiagnosticsEngine` references around in arguments several levels deep would have made the code quite convoluted in my opinion.

Therefore, I decided to carve out the whole configuration resolution logic into its own helper struct that saves a `DiagnosticsEngine` reference . I believe this also enhanced the frontend design slightly because it leads to a clearer division of responsibilities.

Sources/swift-format/Subcommands/DumpConfiguration.swift
```diff
@@ -15,32 +15,59 @@ import Foundation
...
-        print(jsonString)
+
+        configuration = effectiveConfiguration
+      } else {
+        configuration = Configuration()
+      }
+
+      do {
+        let configurationAsJson = try configuration.asJsonString()
+        print(configurationAsJson)
```
ahoppen, line 67: I guess this could be shortened slightly.

```suggestion
        print(try configuration.asJsonString())
```

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -16,6 +16,118 @@ import SwiftParser
 import SwiftSyntax
 
 class Frontend {
+  /// Provides formatter configurations for given `.swift` source files, configuration files or configuration strings.
+  struct ConfigurationProvider {
+    /// Loads formatter configuration files and chaches them in memory.
+    private var configurationLoader: ConfigurationLoader = ConfigurationLoader()
```
ahoppen, line 22: Actually, I don’t think this needs to be a member at all. A local variable in `provide` should be sufficient, right?

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -16,6 +16,118 @@ import SwiftParser
 import SwiftSyntax
 
 class Frontend {
+  /// Provides formatter configurations for given `.swift` source files, configuration files or configuration strings.
+  struct ConfigurationProvider {
```
ahoppen, line 20: Just wondering: Would it be simpler if we just had a static method (or global function) like `func configuration(forConfigPathOrString pathOrString: String?, orForSwiftFileAt swiftFileURL: URL?, diagnosticsEngine: DiagnosticsEngine)` and then either make `checkForUnrecognizedRules` a nested function or make it take a `DiagnosticsEngine` as well?  To me, having a type here implies that there is some state being stored, which isn’t actually the case.

ahoppen (commented):
Thanks for addressing my comments @andre-richter, this seems a lot cleaner to me. I have a few more questions/comments inline.

Sources/swift-format/Frontend/Frontend.swift
andre-richter: It stores the cache of previously loaded configurations. So it wouldn't work as a local variable (as in: stack-allocated inside `provide`, if that is what you meant?).

Sources/swift-format/Frontend/Frontend.swift
andre-richter: > To me, having a type here implies that there is some state being stored, which isn’t actually the case.

There is state being stored in the form of `configurationLoader`, which stores the cache for previously loaded configs.

ahoppen: Oh, I see. Fair enough then.

ahoppen (approved):

PR #929, opened 2025-02-11
Add pre-commit hook for linting

nikalra:
- `swift-lint` for invoking `swift-format lint` with the default options
- `swift-lint-strict` for invoking `swift-format lint --strict`

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,3 +4,15 @@
   language: swift
   types: [swift]
   require_serial: true
+- id: swift-lint
+  name: swift-lint
+  entry: swift-format lint
+  language: swift
+  types: [swift]
+  require_serial: true
+- id: swift-lint-strict
+  name: swift-lint-strict
+  entry: swift-format lint --strict
+  language: swift
+  types: [swift]
+  require_serial: true
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -6,13 +6,7 @@
   require_serial: true
 - id: swift-format-lint
   name: swift-format-lint
-  entry: swift-format lint
-  language: swift
-  types: [swift]
-  require_serial: true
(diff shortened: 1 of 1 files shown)


allevato:
We should probably avoid the name `swift-lint` to avoid confusion since there is another open-source project called SwiftLint. `swift-format-lint` might be better to reinforce which tool is performing the action.

@ahoppen Thoughts, since you reviewed the previous precommit check PR?

.pre-commit-hooks.yaml
```diff
@@ -4,3 +4,15 @@
   language: swift
   types: [swift]
   require_serial: true
+- id: swift-format-lint
+  name: swift-format-lint
+  entry: swift-format lint
+  language: swift
+  types: [swift]
```
ahoppen, line 11: `swift-format lint` without `--strict` will currently not exit with a non-zero exit code because swift-format currently only emits warnings. So, I think we should only provide the `--strict` option.

Also, I think we should use `--recursive --parallel` here, otherwise I don’t think that we’ll iterate through all directories.

ahoppen (commented):
Thanks for the contribution, @nikalra!

.pre-commit-hooks.yaml
nikalra: I think the hook should receive all of the file names on the command line, but I added the `--recursive` option anyway. Re `--parallel`, is there any benefit to using the built-in `--parallel` over removing `requires_serial: true`?

If we can remove that, it should chunk the filenames into N jobs and run those N in parallel. Though I guess we'd still want the `--parallel` flag to process each file name in the chunk.

ahoppen: Using swift-format's native parallelism only spawns a single process, so if anything, it avoids some process spawning overhead.

ahoppen (approved):
Thanks. Looks good to me.

PR #934, opened 2025-02-15
Modify PrettyPrinter to output diagnostics using shortened relative paths

### LINKED ISSUE #818: Lint is inconsistent on how source paths are reported
1. Checkout https://github.com/apple/swift-protobuf @ c9fbc332ed293c64ff70af8d82367ca530dce11c
2. Run
    ```shell
    swift format lint \
        Sources/SwiftProtobuf/Message.swift \
        Tests/SwiftProtobufTests/Test_Extensions.swift
    ```

You get:

```
Sources/SwiftProtobuf/Message.swift:46:16: warning: [DontRepeatTypeInStaticProperties] remove the suffix 'Message' from the name of the variable 'protoMessageName'
/Users/thomasvl/src/github/swift-protobuf/Tests/SwiftProtobufTests/Test_Extensions.swift:97:68: warning: [EndOfLineComment] move end-of-line comment that exceeds the line length
```

The order of the two arguments doesn't matters, the path to `Message.swift` is reported as the shortened relative path and `Test_Extensions.swift` is always reported as a full path instead.

TTOzzi:
Resolve #818 

Modified `PrettyPrinter`'s diagnostic logic to output diagnostics with shortened relative paths instead of full paths, as I believe this is more appropriate.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -819,7 +819,7 @@ public class PrettyPrinter {
     context.findingEmitter.emit(
       message,
       category: category,
-      location: Finding.Location(file: context.fileURL.path, line: outputBuffer.lineNumber, column: column)
+      location: Finding.Location(file: context.fileURL.relativePath, line: outputBuffer.lineNumber, column: column)
     )
   }
 }


Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -819,7 +819,7 @@ public class PrettyPrinter {
     context.findingEmitter.emit(
       message,
       category: category,
-      location: Finding.Location(file: context.fileURL.path, line: outputBuffer.lineNumber, column: column)
+      location: Finding.Location(file: context.fileURL.relativePath, line: outputBuffer.lineNumber, column: column)
```
ahoppen, line 822: I’m a little skeptical of this. When forming a URL, we provide a `baseURL` and a path relative to it. Generally, I wouldn’t expect it to matter what’s part of the base URL and what’s part of the relative path. For example, I would have expected `url = URL(string: url.absoluteString)` to not have any effect on the output but with this change it does. 

What I’d prefer instead is some way of stripping off a known base path from `context.fileURL` here (though I don’t have concrete suggestions of how to do so at the moment).

allevato: FWIW, this change _would_ bring the pretty printer to be consistent with how we're using the `relativePath` property elsewhere.

When we create the `SourceLocationConverter` that the other non-pretty-printer rules use in [Context.swift](https://github.com/swiftlang/swift-format/blob/fb6340440c36bfaeff045e6044b07ab0b40ce900/Sources/SwiftFormat/Core/Context.swift#L84-L85), we use the `relativePath` property.

Then, in [FileIterator.swift](https://github.com/swiftlang/swift-format/blob/fb6340440c36bfaeff045e6044b07ab0b40ce900/Sources/SwiftFormat/Utilities/FileIterator.swift#L151-L162), we have a bunch of logic that explicitly constructs the `URL`s this way. (In fact, I thought we were using `FileManager.DirectoryEnumerationOptions.producesRelativePathURLs` here, but that doesn't seem to be the case. That might have been historical, or there might have been a platform-specific issue that prevented it.)

ahoppen: Oh, OK. I didn’t realize that we already used `relativePath` elsewhere. While I have the same objection still in principle, I don’t object to adding another use for it if there’s already precedence.

ahoppen (approved):

PR #935, opened 2025-02-15
Improve the formatting logic of OmitExplicitReturns rule

### LINKED ISSUE #812: _lint_ flags things for _OmitExplicitReturns_ but _format_ doesn't fix them
With a dummy test file:

```swift
var f = l.filter { return $0.a != o }
```

And a config file that includes:

```json
    "OmitExplicitReturns" : true,
```

```shell
swift format lint --configuration swift-format.cfg test.swift
```

Produces:
> test.swift:1:20: warning: [OmitExplicitReturns] 'return' can be omitted because body consists of a single expression

However
```shell
swift format format --configuration swift-format.cfg test.swift
```

Doesn't actually return the `return` in what it produces.

TTOzzi:
Resolve #812 and fix an additional minor bug that was discovered.
I will leave the explanation in the comments.

### DIFF
--- a/Sources/SwiftFormat/Rules/OmitExplicitReturns.swift
+++ b/Sources/SwiftFormat/Rules/OmitExplicitReturns.swift
@@ -60,7 +60,7 @@ public final class OmitExplicitReturns: SyntaxFormatRule {
     guard let accessorBlock = binding.accessorBlock,
       let transformed = transformAccessorBlock(accessorBlock)
     else {
-      return node
+      return super.visit(node)
     }
 
     binding.accessorBlock = transformed
@@ -145,7 +145,7 @@ public final class OmitExplicitReturns: SyntaxFormatRule {
     CodeBlockItemListSyntax([
       CodeBlockItemSyntax(
         leadingTrivia: returnStmt.leadingTrivia,
-        item: .expr(returnStmt.expression!),
+        item: .expr(returnStmt.expression!.detached.with(\.trailingTrivia, [])),
         semicolon: nil,
         trailingTrivia: returnStmt.trailingTrivia
       )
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormat/Rules/OmitExplicitReturns.swift
```diff
@@ -60,7 +60,7 @@ public final class OmitExplicitReturns: SyntaxFormatRule {
     guard let accessorBlock = binding.accessorBlock,
       let transformed = transformAccessorBlock(accessorBlock)
     else {
-      return node
+      return super.visit(node)
```
TTOzzi, line 63: If the condition was not met, the `node` was returned immediately, preventing the additional `visit` logic from executing.  
By calling `super.visit`, the `visit` process continues, allowing the formatting logic in `visit(_ node: ClosureExprSyntax)` to be applied.

Sources/SwiftFormat/Rules/OmitExplicitReturns.swift
```diff
@@ -145,7 +145,7 @@ public final class OmitExplicitReturns: SyntaxFormatRule {
     CodeBlockItemListSyntax([
       CodeBlockItemSyntax(
         leadingTrivia: returnStmt.leadingTrivia,
-        item: .expr(returnStmt.expression!),
+        item: .expr(returnStmt.expression!.detached.with(\.trailingTrivia, [])),
```
TTOzzi, line 148: Since `returnStmt.expression` and `returnStmt` had the same `trailingTrivia`, formatting caused `trailingTrivia` to appear twice when present.
To fix this, I detached the `trailingTrivia` of `returnStmt.expression` when rewrapping with `CodeBlockItemSyntax`, ensuring it does not get duplicated.

ahoppen (approved):
Thank you, change looks good to me.

PR #936, opened 2025-02-17
Ignore backticks in documentation abstracts

### LINKED ISSUE #924: invalid `BeginDocumentationCommentWithOneLineSummary` warning
```swift
  /// Open a scope enclosed by `"typewriter"` double-quotes.
  ///
  /// Closed by a ``EndTypewriterDoubleQuotes`` macro.
  ///
  /// __Example Usage__:
  ///   ```swift
  ///   BeginTypewriterDoubleQuotes()
  ///   "Hello, Swift!"
  ///   EndTypewriterDoubleQuotes()
  ///   ```
  public struct BeginTypewriterDoubleQuotes: MDocMacroProtocol {
    public static let kind = "Qo"
    public var arguments: [MDocASTNode]
    /// Creates a new `BeginTypewriterDoubleQuotes` macro.
    public init() {
      self.arguments = []
    }
  }
```

triggers: `Tools/generate-manual/MDoc/MDocMacro.swift:1144:3: warning: [BeginDocumentationCommentWithOneLineSummary] terminate this sentence with a period: "Open a scope enclosed by `"typewriter"` double-quotes."`

natecook1000:
The linguistic tagger doesn't do a great job recognizing backticks as opening vs. closing quotes, so some valid abstracts that contain backticks were being rejected by the `BeginDocumentationCommentWithOneLineSummary` rule. This change removes any `InlineCode` nodes from the abstract paragraph before processing, since they aren't important to that judgment.

Fixes #924.

### DIFF
--- a/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
+++ b/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
@@ -97,8 +97,13 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
     else { return }
 
     // For the purposes of checking the sentence structure of the comment, we can operate on the
-    // plain text; we don't need any of the styling.
-    let trimmedText = briefSummary.plainText.trimmingCharacters(in: .whitespacesAndNewlines)
+    // plain text; we don't need any of the styling. Additionally, the backticks that are
+    // frequently used to denote symbol names cause issues with quote identification, and
+    // aren't necessary for this purpose.
+    let trimmedText = briefSummary.plainText
+      .trimmingCharacters(in: .whitespacesAndNewlines)
+      .replacingOccurrences(of: "``.+?``", with: "ZZZ", options: .regularExpression)
+      .replacingOccurrences(of: "`.+?`", with: "zzz", options: .regularExpression)
     let (commentSentences, trailingText) = sentences(in: trimmedText)
     if commentSentences.count == 0 {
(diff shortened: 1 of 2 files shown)


ahoppen (commented):
Does this cause issues when you have a code block that contains a period and would thus recognize it as a multi-sentence first paragraph?

Eg.
```
/// The same as `Foo.foo()`.
func bar() {}
```

If not, could you add a test case for that?

allevato:
We can assume that a sentence-ending period will never occur _inside_ an inline code block, so I wonder if we should simply remove the content within backticks before doing this check, instead of removing just the backticks.

Since the `briefSummary` field is already a `Paragraph` node, we could do this in a structured way by just locally mutating that tree to remove any `InlineCode` children (and likewise for `SymbolLink` children, to handle the DocC-style double-backtick syntax).

The only thing I'm not sure about is how the linguistic tagger would handle a space before the period, because I'm assuming that removing the `InlineCode` node from this:

```
Foo bar `baz`.
```

would leave us with `"Foo bar ."`, and I'm not familiar enough with that API to know if it would handle that. If that's a problem, then we could address that by instead of removing the `InlineCode/SymbolLink` children, replacing them with a plaintext word like "placeholder", and then run the linguistic tagger on that result.

natecook1000:
I would have sworn I updated this commit before opening the PR, but here we are. @allevato, I switched to use a `MarkupRewriter` to remove `InlineCode` nodes, and it seems to work great! I added a test case using a double-backtick link (which I don't think should have periods, but just in case) and it seemed to do fine.

PR #940, opened 2025-02-19
Skip flaky performance test on Windows

weliveindetail:
Let's skip `testWhitespaceLinterPerformance` on Windows until someone with more linter performance insights can craft a real fix for https://github.com/swiftlang/swift-format/issues/939

### DIFF
--- a/Tests/SwiftFormatPerformanceTests/WhitespaceLinterPerformanceTests.swift
+++ b/Tests/SwiftFormatPerformanceTests/WhitespaceLinterPerformanceTests.swift
@@ -5,7 +5,7 @@ import XCTest
 @_spi(Testing) import _SwiftFormatTestSupport
 
 final class WhitespaceLinterPerformanceTests: DiagnosingTestCase {
-  func testWhitespaceLinterPerformance() {
+  func testWhitespaceLinterPerformance() throws {
     #if os(Windows)
     // https://github.com/swiftlang/swift-format/issues/939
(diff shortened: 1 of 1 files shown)


bnbarham (approved):
Thanks @weliveindetail, was just about to put this up. These run in a VM, though the 200% variance is still odd to me. @allevato was this intended to run in CI anyway, or was it just for local testing?

allevato:
> @allevato was this intended to run in CI anyway, or was it just for local testing?

Aside from being flaky, I can't think of a reason that we _wouldn't_ want them to run in CI. But I'm not opposed to disabling them on platforms where it's currently a problem.

I can't think of anything Windows-specific that would cause the results to have such high variance, but I'm also not a frequent Windows user.

bnbarham:
> Aside from being flaky, I can't think of a reason that we wouldn't want them to run in CI

It's really only testing for variance *within the run* in CI (unless I'm missing something?) and it seems unlikely that should ever occur outside of CI-VM-flakiness.

allevato:
Yeah, that's fair. I imagine there's an environment variable that we can check inside the test to decide whether to run it or not? Or, alternatively, check it inside the package manifest and not create that test target when it builds for CI?

bnbarham:
Yeah, we often use `SWIFTCI_USE_LOCAL_DEPS` for that.

> Or, alternatively, check it inside the package manifest and not create that test target when it builds for CI?

SGTM

PR #943, opened 2025-02-21
Remove severity

bkolb:
Instead of making the severity configurable,
this patch removes severity all together and treats every finding as an error.

Issue: #879

### DIFF
--- a/Sources/SwiftFormat/Core/Parsing.swift
+++ b/Sources/SwiftFormat/Core/Parsing.swift
@@ -63,11 +63,9 @@ func parseAndEmitDiagnostics(
     for diagnostic in diagnostics {
       let location = diagnostic.location(converter: expectedConverter)
 
-      // Downgrade editor placeholders to warnings, because it is useful to support formatting
+      // Ignore editor placeholders, because it is useful to support formatting
       // in-progress files that contain those.
-      if diagnostic.diagnosticID == StaticTokenError.editorPlaceholder.diagnosticID {
-        parsingDiagnosticHandler(downgradedToWarning(diagnostic), location)
-      } else {
+      if diagnostic.diagnosticID != StaticTokenError.editorPlaceholder.diagnosticID {
         parsingDiagnosticHandler(diagnostic, location)
         hasErrors = true
       }
@@ -78,30 +76,4 @@ func parseAndEmitDiagnostics(
     throw SwiftFormatError.fileContainsInvalidSyntax
   }
   return sourceFile
-}
-
-// Wraps a `DiagnosticMessage` but forces its severity to be that of a warning instead of an error.
-struct DowngradedDiagnosticMessage: DiagnosticMessage {
-  var originalDiagnostic: DiagnosticMessage
-
-  var message: String { originalDiagnostic.message }
-
-  var diagnosticID: SwiftDiagnostics.MessageID { originalDiagnostic.diagnosticID }
-
-  var severity: DiagnosticSeverity { .warning }
-}
-
-/// Returns a new `Diagnostic` that is identical to the given diagnostic, except that its severity
-/// has been downgraded to a warning.
-func downgradedToWarning(_ diagnostic: Diagnostic) -> Diagnostic {
-  // `Diagnostic` is immutable, so create a new one with the same values except for the
-  // severity-downgraded message.
-  return Diagnostic(
-    node: diagnostic.node,
-    position: diagnostic.position,
-    message: DowngradedDiagnosticMessage(originalDiagnostic: diagnostic.diagMessage),
-    highlights: diagnostic.highlights,
-    notes: diagnostic.notes,
-    fixIts: diagnostic.fixIts
-  )
 }
--- a/Sources/_SwiftFormatTestSupport/Configuration+Testing.swift
+++ b/Sources/_SwiftFormatTestSupport/Configuration+Testing.swift
@@ -44,4 +44,11 @@ extension Configuration {
     config.indentBlankLines = false
     return config
   }
+
+  public static func forTesting(enabledRule: String) -> Configuration {
+    var config = Configuration.forTesting
+    config.rules = config.rules.mapValues({ _ in false })
+    config.rules[enabledRule] = true
+    return config
+  }
 }
--- a/Sources/swift-format/Utilities/DiagnosticsEngine.swift
+++ b/Sources/swift-format/Utilities/DiagnosticsEngine.swift
@@ -134,15 +134,8 @@ final class DiagnosticsEngine {
   /// Converts a lint finding into a diagnostic message that can be used by the `TSCBasic`
   /// diagnostics engine and returns it.
   private func diagnosticMessage(for finding: Finding) -> Diagnostic {
-    let severity: Diagnostic.Severity
-    switch finding.severity {
-    case .error: severity = .error
-    case .warning: severity = .warning
-    case .refactoring: severity = .warning
-    case .convention: severity = .warning
-    }
     return Diagnostic(
-      severity: severity,
+      severity: .error,
       location: finding.location.map(Diagnostic.Location.init),
       category: "\(finding.category)",
       message: "\(finding.message.text)"
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 swift-format.xcodeproj/
 Package.resolved
 /.vscode
+.index-build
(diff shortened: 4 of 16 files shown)


Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -40,7 +40,7 @@ class Frontend {
       // That way they will be printed out, but we'll continue execution on the valid rules.
       let invalidRules = configuration.rules.filter { !RuleRegistry.rules.keys.contains($0.key) }
       for rule in invalidRules {
-        diagnosticsEngine.emitWarning("Configuration contains an unrecognized rule: \(rule.key)", location: nil)
+        diagnosticsEngine.emitError("Configuration contains an unrecognized rule: \(rule.key)", location: nil)
```
ahoppen, line 43: I think this could break clients in the following scenario: You have a Swift 6.1 rule added to your `.swift-format` file but locally you are running a 6.0 `swift-format` to format your code (not to lint). That should mostly work but with this change, formatting would exit with 1 and thus fail (IIUC Xcode would show an error message in that case). So, I think this would need to become a print to stderr instead of going through the diagnostic engine.

CC @allevato Do you have thoughts on this?

ahoppen (commented):
Thank you, Bernd. Looks good to me overall, just have a few thoughts inline. I would also like to hear @allevato’s opinion before merging this.

Sources/swift-format/Frontend/Frontend.swift
allevato: I'll do a more thorough soon, but for this issue specifically: I'd like to keep all messages from the formatter going through the diagnostics engine instead of having some there and some through just `stderr`.

I think the answer here is that we should only remove the severity from `Finding`s but not `Diagnostic`/`DiagnosticEngine`. That removes the complexity from the rules themselves (and forces all findings to become errors, which is the change we're asking for), but still lets the formatter frontend issue other diagnostics that aren't as severe.

Tests/SwiftFormatTests/Rules/OmitReturnsTests.swift
```diff
@@ -16,7 +16,10 @@ final class OmitReturnsTests: LintOrFormatRuleTestCase {
           }
         """,
       findings: [
-        FindingSpec("1️⃣", message: "'return' can be omitted because body consists of a single expression")
+        FindingSpec(
```
allevato, line 19: Here and below, are these reformattings relevant to this PR? If not, can you revert them to reduce the number of unrelated changes?

Sources/_SwiftFormatTestSupport/Configuration+Testing.swift
```diff
@@ -44,4 +44,24 @@ extension Configuration {
...
+}
+
+extension Configuration {
+  public func disableAllRules() -> Self {
+    var config = self
+    config.rules = config.rules.mapValues({ _ in false })
+    return config
+  }
+
+  public func enable(_ rule: String, enabled: Bool) -> Self {
```
allevato, line 62: This doesn't look used; can it be removed?

Sources/_SwiftFormatTestSupport/Configuration+Testing.swift
```diff
@@ -44,4 +44,24 @@ extension Configuration {
...
+
+  public static func forTesting(enabledRule: String) -> Configuration {
+    var config = Configuration.forTesting.disableAllRules()
+    config.rules[enabledRule] = true
+    return config
+  }
+}
+
+extension Configuration {
+  public func disableAllRules() -> Self {
```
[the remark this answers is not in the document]
allevato: It looks like this is unused outside of the call in `forTesting(...)` above; I would consider inlining it there until we have a separate need for it.

Sources/SwiftFormat/Core/Parsing.swift
```diff
@@ -65,9 +65,7 @@ func parseAndEmitDiagnostics(
 
       // Downgrade editor placeholders to warnings, because it is useful to support formatting
```
allevato, line 66: Please update this comment to indicate that editor placeholder errors are being ignored instead of downgraded to warnings.

(Which I think is fine. Editor placeholders are common in source being interactively edited and formatted that I think it's fine to not report them at all.)

Sources/_SwiftFormatTestSupport/Configuration+Testing.swift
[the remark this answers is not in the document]
bkolb: done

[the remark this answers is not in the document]
ahoppen: Looks like you didn’t push your recent changes 😉

Tests/SwiftFormatTests/Rules/OmitReturnsTests.swift
bkolb: I get them when running 
`xcrun swift-format --recursive --in-place --configuration .swift-format .`
I am not sure why this changed.

Sources/generate-swift-format/PrettyPrintCollector.swift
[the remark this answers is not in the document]
bkolb: yes and no. This was to be able to allow setting a severity also for pretty print issues. Is there a way to disable those otherwise?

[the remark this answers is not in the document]
bkolb: It is related to this: 
Sources/SwiftFormat/Core/RuleRegistry+Generated.swift

Sources/swift-format/Utilities/DiagnosticsEngine.swift
```diff
@@ -92,7 +91,7 @@ final class DiagnosticsEngine {
     for note in finding.notes {
       emit(
         Diagnostic(
-          severity: .note,
+          severity: .warning,
```
allevato, line 94: Should this change be reverted? We're only removing severity from findings, but diagnostics should still support a `note` level severity.

Tests/SwiftFormatTests/Rules/OmitReturnsTests.swift
allevato: Did you have any temporary changes to the swift-format configuration in your directory when you ran it? These lines don't exceed the 120 column limit so they shouldn't have been modified, as far as I can tell.

Sources/generate-swift-format/PrettyPrintCollector.swift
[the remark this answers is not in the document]
allevato: IMO, adding a way to disable pretty-print findings is a significant enough feature that it should be factored out into its own PR, not included in a PR about removing finding severity. Can you please do that?

I think we _do_ want that feature, it'll just be easier to track things as a separate PR.

Tests/SwiftFormatTests/Rules/OmitReturnsTests.swift
bkolb: no, not that I am aware of. I will manually revert it

Sources/generate-swift-format/PrettyPrintCollector.swift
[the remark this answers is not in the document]
bkolb: done

bkolb:
I think I addressed all your comments. Can you please check again?

allevato (approved):
Thanks! This is a lot cleaner.

allevato (commented):
Looks like these need to be added to `api-breakages.txt`:

```
7 breaking changes detected in SwiftFormat:
  💔 API breakage: func Rule.diagnose(_:on:severity:anchor:notes:) has parameter 2 type change from SwiftFormat.Finding.Severity? to SwiftFormat.FindingAnchor
  💔 API breakage: func Rule.diagnose(_:on:severity:anchor:notes:) has parameter 3 type change from SwiftFormat.FindingAnchor to [SwiftFormat.Finding.Note]
  💔 API breakage: enum Finding.Severity has been removed
  💔 API breakage: var Finding.severity has been removed
  💔 API breakage: var FindingCategorizing.defaultSeverity has been removed
  💔 API breakage: var FindingCategorizing.defaultSeverity has been removed
  💔 API breakage: func Rule.diagnose(_:on:severity:anchor:notes:) has been renamed to func diagnose(_:on:anchor:notes:)
```

bkolb:
I updated the api-breakages.txt file. Can you please approve the workflow again?

PR #947, opened 2025-02-25
Clean up symlink following and recursively follow symlink chains.

allevato:
This started out as a weird debugging journey. We had a team running `swift-format` as part of a Bazel action to format generated Swift code. The action basically just ran `swift-format $input > $output`. This action started failing (generating no output) in this environment because the Bazel action workspace was set up such that `$input` was a symlink to the actual file. Even if I updated the action to pass `--follow-symlinks`, the old `FileIterator` code had a `fallthrough` that didn't correctly handle the case where a path passed directly on the command line was a symlink to a *file* instead of a directory.

The reason that this _actually_ started failing was because we finally updated our internal toolchain to use the new swift-foundation instead of the legacy swift-corelibs-foundation (before it was layered on top of swift-foundation). Something in the implementation of `FileManager` changed subtly and started giving us `URL`s back that pointed to symlinks when previously we were getting them pre-resolved, which revealed the above issue.

While I was in here, I decided to add support for following symlinks to other symlinks, just to make sure that didn't become an issue in the future.

### DIFF
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -71,25 +71,18 @@ public struct FileIterator: Sequence, IteratorProtocol {
       if dirIterator != nil {
         output = nextInDirectory()
       } else {
-        guard var next = urlIterator.next() else {
+        guard let next = urlIterator.next() else {
           // If we've reached the end of all the URLs we wanted to iterate over, exit now.
           return nil
         }
-
-        guard let fileType = fileType(at: next) else {
+        guard let (next, fileType) = fileAndType(at: next, followSymlinks: followSymlinks) else {
           continue
         }
 
         switch fileType {
         case .typeSymbolicLink:
-          guard
-            followSymlinks,
-            let destination = try? FileManager.default.destinationOfSymbolicLink(atPath: next.path)
-          else {
-            break
-          }
-          next = URL(fileURLWithPath: destination, relativeTo: next)
-          fallthrough
+          // If we got here, we encountered a symlink but didn't follow it. Skip it.
+          continue
 
         case .typeDirectory:
           dirIterator = FileManager.default.enumerator(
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -173,9 +159,34 @@ public struct FileIterator: Sequence, IteratorProtocol {
...
+    return try? FileManager.default.attributesOfItem(atPath: url.path)[.type] as? FileAttributeType
+  }
+
+  guard var fileType = typeOfFile(at: url) else {
+    return nil
+  }
+
+  var url = url
+  while followSymlinks && fileType == .typeSymbolicLink,
+    let destination = try? FileManager.default.destinationOfSymbolicLink(atPath: url.path)
```
ahoppen, line 183: Should we detect circles here if you have a symlink that points to itself?

ahoppen (commented):
Just because it has bitten me multiple times in the past, could you double check that that this works correctly for `/tmp` on macOS (which is a symlink to `/private/tmp`) but `resolvingSymlinksInPath` returns the normalized `/tmp` regardless.

Sources/SwiftFormat/Utilities/FileIterator.swift
allevato: Good catch; fixed. Symlink cycles (either a link pointing to itself, or a multi-link cycle) are silently ignored now. We could warn on this, but we'd have to plumb `DiagnosticsEngine` down into `FileIterator`, which I didn't really want to do here if we can avoid it.

ahoppen: As long as we don’t infinite loop, I think the behavior is OK.

Also, it only just occurs to me, why can’t we use `resolvingSymlinksInPath`/`realpath` here if we follow all symlinks?

In SourceKit-LSP, we have the following code to avoid the `/tmp` vs `/private/tmp` issue: https://github.com/swiftlang/sourcekit-lsp/blob/3e7f350246dc86072a9dfc88f25782dc86dd1c9a/Sources/SwiftExtensions/URLExtensions.swift#L40-L57

allevato: Hmm, good question. That `realpath` code seems to work on macOS from local testing, so let me see if it works on Linux/Windows. Assuming they both consistently resolve all symlinks deeply or stop when a cycle is reached on all platforms, then that would remove the need for our own loop.

allevato: Sigh, no good. Based on CI results and local testing, it looks like `resolvingSymlinksInPath` on Linux pre-5.x Foundation doesn't correctly resolve through multiple symlinks. So we'd need a special case to handle those anyway, which would essentially be the code I had earlier anyway.

Maybe one day when we drop 5.x support, we can clean this all up.

ahoppen (approved):

PR #948, opened 2025-02-27
Make ValidateDocumentationComments allow `Throws` doc comments for rethrows functions

### LINKED ISSUE #928: Incorrect ValidateDocumentationComments warning for rethrows function
`Sources/ArgumentParser/Utilities/Mutex.swift:53:5: warning: [ValidateDocumentationComments] remove the 'Throws:' sections of 'withLock'; it does not throw any errors`

```swift
  /// Calls the given closure after acquiring the lock and then releases
  /// ownership.
  ///
  /// - Warning: Recursive calls to `withLock` within the closure parameter has
  ///   behavior that is platform dependent. Some platforms may choose to panic
  ///   the process, deadlock, or leave this behavior unspecified. This will
  ///   never reacquire the lock however.
  ///
  /// - Parameter body: A closure with a parameter of `Value` that has exclusive
  ///   access to the value being stored within this mutex. This closure is
  ///   considered the critical section as it will only be executed once the
  ///   calling thread has acquired the lock.
  ///
  /// - Throws: Re-throws any error thrown by `body`.
  ///
  /// - Returns: The return value, if any, of the `body` closure parameter.
  func withLock<U>(
    _ body: (inout T) throws -> U
  ) rethrows -> U {
    self.lock.lock()
    defer { self.lock.unlock() }
    return try body(&self.value)
  }
```

TTOzzi:
Resolve #928 

Respecting the intent of the existing implementation, change `Throws` documentation for `rethrows` functions to optional.

- `throws` → Required; a diagnostic is triggered if missing.
- `rethrows` → Optional; no diagnostic is triggered if missing.
- none → If a `Throws` comment is present, a diagnostic is triggered to remove it.

### DIFF
--- a/Sources/SwiftFormat/Rules/ValidateDocumentationComments.swift
+++ b/Sources/SwiftFormat/Rules/ValidateDocumentationComments.swift
@@ -131,12 +131,13 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
     throwsDescription: Paragraph?,
     node: DeclSyntax
   ) {
-    // If a function is marked as `rethrows`, it doesn't have any errors of its
-    // own that should be documented. So only require documentation for
-    // functions marked `throws`.
+    // Documentation is required for functions marked as `throws`.
+    // For functions marked as `rethrows`, documentation is not enforced
+    // since they don’t introduce new errors of their own.
+    // However, it can still be included if needed.
     let needsThrowsDesc = throwsOrRethrowsKeyword?.tokenKind == .keyword(.throws)
 
-    if !needsThrowsDesc && throwsDescription != nil {
+    if throwsOrRethrowsKeyword == nil && throwsDescription != nil {
       diagnose(
         .removeThrowsComment(funcName: name),
         on: throwsOrRethrowsKeyword ?? node.firstToken(viewMode: .sourceAccurate)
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormat/Rules/ValidateDocumentationComments.swift
```diff
@@ -131,12 +131,13 @@ public final class ValidateDocumentationComments: SyntaxLintRule {
...
-    // own that should be documented. So only require documentation for
-    // functions marked `throws`.
+    // Documentation is required for functions marked as `throws`.
+    // For functions marked as `rethrows`, documentation is not enforced
+    // since they don’t introduce new errors of their own.
+    // However, it can still be included if needed.
     let needsThrowsDesc = throwsOrRethrowsKeyword?.tokenKind == .keyword(.throws)
 
-    if !needsThrowsDesc && throwsDescription != nil {
+    if throwsOrRethrowsKeyword?.tokenKind == nil && throwsDescription != nil {
```
allevato, line 140: Can this just be `throwsOrRethrowsKeyword == nil`?

allevato (approved):
One minor comment, but otherwise looks good. Thanks!

ahoppen:
Windows failures look unrelated. I’m investigating.

PR #950, opened 2025-03-01
Support file-level ignore directive for specific rules

### LINKED ISSUE #927: No syntax to disable rule for file
You should be able to write `// swift-format-ignore-file: <rule>...` to disable rules for an entire file.

TTOzzi:
Resolve #927

I have modified the function that disables specific rules in node-level directive so that it can also be reused for file-level directive, ensuring both follow the same logic.

Previously, if a file-level directive appeared in the topmost comment, we would always skip visiting nodes. However, with this change, we now visit child nodes when only specific rules are disabled. As a result, this caused an unintended side effect in cases where the `OrderedImports` rule was enabled. Specifically, a directive included in the `leadingTrivia` of a sorted import statement was being reordered along with the imports.
To resolve this, I modified the `OrderedImports` rule to treat the `swift-format-ignore-file` directive as part of the file header.

Additionally, to avoid simple string comparisons like:

```swift
if line.description.contains("swift-format-ignore-file") {
```

I have created an `IgnoreDirective` enum to handle ignore directives more cleanly.

For easier review, I have split the commits.
Once the review is complete, I will squash and force push them before merging. 🙂

### DIFF
--- a/Sources/SwiftFormat/Core/RuleMask.swift
+++ b/Sources/SwiftFormat/Core/RuleMask.swift
@@ -35,6 +35,16 @@ import SwiftSyntax
 ///   2. |  let a = 123
 ///   Ignores `RuleName` and `OtherRuleName` for line 2.
 ///
+///   1. |  // swift-format-ignore-file: RuleName
+///   2. |  let a = 123
+///   3. | class Foo { }
+///   Ignores `RuleName` for the entire file (lines 2-3).
+///
+///   1. |  // swift-format-ignore-file: RuleName, OtherRuleName
+///   2. |  let a = 123
+///   3. | class Foo { }
+///   Ignores `RuleName` and `OtherRuleName` for the entire file (lines 2-3).
+///
 /// The rules themselves reference RuleMask to see if it is disabled for the line it is currently
 /// examining.
 @_spi(Testing)
@@ -85,6 +95,29 @@ extension SourceRange {
   }
 }
 
+/// Represents the kind of ignore directive encountered in the source.
+enum IgnoreDirective: CustomStringConvertible {
+  /// A node-level directive that disables rules for the following node and its children.
+  case node
+  /// A file-level directive that disables rules for the entire file.
+  case file
+
+  var description: String {
+    switch self {
+    case .node:
+      return "swift-format-ignore"
+    case .file:
+      return "swift-format-ignore-file"
+    }
+  }
+
+  /// Regex pattern to match an ignore comment. This pattern supports 0 or more comma delimited rule
+  /// names. The rule name(s), when present, are in capture group #3.
+  fileprivate var pattern: String {
+    return #"^\s*\/\/\s*"# + description + #"((:\s+(([A-z0-9]+[,\s]*)+))?$|\s+$)"#
+  }
+}
+
 /// A syntax visitor that finds `SourceRange`s of nodes that have rule status modifying comment
 /// directives. The changes requested in each comment is parsed and collected into a map to support
 /// status lookup per rule name.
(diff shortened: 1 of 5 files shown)


ahoppen:
@TTOzzi The Windows CI failures should be addressed by https://github.com/swiftlang/swift-format/pull/951. Could you rebase your changes on top of `main` so that we can pick up that fix for CI?

TTOzzi:
> @TTOzzi The Windows CI failures should be addressed by #951. Could you rebase your changes on top of `main` so that we can pick up that fix for CI?

Oh, thank you for identifying the cause of the failure 😃
I’ve rebased it.

Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift
```diff
@@ -651,4 +651,36 @@ final class OrderedImportsTests: LintOrFormatRuleTestCase {
       ]
     )
   }
+
+  func testImportsOrderWithFileIgnoreDirective() {
+    assertFormatting(
```
ahoppen, line 656: Does this also end up adding a newline after `// swift-format-ignore-file` if no re-ordering is necessary? Eg. for

```swift
// swift-format-ignore-file
import LibA
import LibB
```

To be clear, I’m fine with that, just want to check and record the behavior in a test case.

--- 

One thing I did notice is that I think this flips comments order around if you have eg.

```swift
// We need to ignore this file because it is generated
// swift-format-ignore-line

import Foundation
```

Maybe `swift-format-ignore-line` just needs to flush the `commentBuffer` into the `headerBuffer` similar to what a blank line does.

Sources/SwiftFormat/Core/RuleMask.swift
```diff
@@ -95,6 +95,34 @@ extension SourceRange {
...
+  /// Regex pattern to match an ignore comment. This pattern supports 0 or more comma delimited rule
+  /// names. The rule name(s), when present, are in capture group #3.
+  private var pattern: String {
+    return #"^\s*\/\/\s*"# + description + #"((:\s+(([A-z0-9]+[,\s]*)+))?$|\s+$)"#
+  }
+
+  /// Rule ignore regex object.
+  fileprivate var regex: NSRegularExpression {
+    return try! NSRegularExpression(pattern: pattern, options: [])
+  }
```
ahoppen, line 123: Since you are working on this already, I’ve got a few thoughts:
- The regex restriction after the `:` seems a little restrictive to me. Eg. I imagine it would be fine to try and debug why a future rule that contains a `_` cannot be ignored using a directive. Maybe we can just make the regex a prefix match. @allevato Do you remember why the regex is this specific?
- This re-compiles the regex every time `regex` is accessed. Compiling regular expressions is fairly expensive, so I’d prefer to cache that.
- Could we use Swift’s Regex instead of `NSRegularExpression`?
- Or, alternatively, would it be easier to write a simple line matcher and avoid going through regex matching at all?

All except for avoiding to re-compile the regex don’t need to be in this PR.

ahoppen (commented):
Thank you and thanks for remembering to not re-order the ignore directive in ordered imports. I’ve got a couple of thoughts inline.

Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift
TTOzzi: > Does this also end up adding a newline after // swift-format-ignore-file if no re-ordering is necessary?

Yes, that's right. In the current implementation,

```swift
// swift-format-ignore-file
import LibA
import LibB
```

is formatted as

```swift
// swift-format-ignore-file

import LibA
import LibB
```

because the `// swift-format-ignore-file` comment is treated as part of the file header and is separated accordingly.

Additionally, I have fixed the issue where the comment order could be flipped when`// swift-format-ignore-line` appeared after the first line, and I have added a test case for it.

Thanks for the feedback!

Sources/SwiftFormat/Core/RuleMask.swift
TTOzzi: Yes, I have updated the implementation to cache the regex instead of recompiling it each time.

And as for other suggestions, would it be a good idea to create a separate issue for these?
- Relaxing the regex restrictions
(If we can raise the minimum deployment version, migrating from `NSRegularExpression` to Swift’s `Regex`)
- Alternatively, parsing ignore directives using string matching instead of regex

ahoppen (approved):
Thanks. Looks good to me 👍🏽

Sources/SwiftFormat/Core/RuleMask.swift
ahoppen: Issues (or PRs 😉) for those items would be good, yes. Also, I am happy to raise the deployment target to macOS 13. SourceKit-LSP, for example, already requires macOS 13.0.

TTOzzi: Got it! I'll work on a follow-up PR for those items 😁

PR #955, opened 2025-03-06
Convert ValidateDocumentationComments to a formatter

natecook1000:
This change enables the `ValidateDocumentationComments` rule to make the swap between outlined and separated parameter descriptions, when necessary. Other diagnostics require user intervention, so they're still only linted.


natecook1000:
Only a draft since it doesn't handle block documentation comments yet. Is this a direction that makes sense? I'm doing the processing directly on the trivia collection, since converting the docs to Markdown doesn't seem to preserve the actual text structure in a way that we can make only minimal changes.

rauhul:
Awesome PR! ❤️

allevato:
> I'm doing the processing directly on the trivia collection, since converting the docs to Markdown doesn't seem to preserve the actual text structure in a way that we can make only minimal changes.

FWIW, the motivation for using the Markdown module in swift-format wasn't just to parse doc comments, but also to apply a canonical formatting to them one day. So I wouldn't mind if that was a prerequisite for doing these kinds of transformations, especially if it makes the implementation easier/cleaner because you'd be operating on Markdown nodes instead of the raw trivia.

PR #956, opened 2025-03-06
Enable the license header check

ahoppen:


allevato (approved):
LG, modulo the Python file that apparently needs to be modified.

bnbarham (approved):

PR #957, opened 2025-03-07
Bump the deployment target to macOS 13 🚀 

TTOzzi:
This PR addresses the follow-up work discussed in https://github.com/swiftlang/swift-format/pull/950#discussion_r1983159771.

- Bump the deployment target to macOS 13
- Relax regex restrictions to allow rules containing `_` or other special characters
- Drop `NSRegularExpression` and migrate to Swift `Regex`


Package.swift
```diff
@@ -137,8 +137,8 @@ if buildOnlyTests {
 let package = Package(
   name: "swift-format",
   platforms: [
-    .macOS("12.0"),
-    .iOS("13.0"),
+    .macOS("13.0"),
+    .iOS("16.0"),
```
TTOzzi, line 141: I also raised the iOS deployment target to 16 to use Swift `Regex`.
One thing I'm curious about—does `swift-format` need to support iOS?

Sources/_SwiftFormatTestSupport/DiagnosingTestCase.swift
```diff
@@ -213,54 +213,45 @@ open class DiagnosingTestCase: XCTestCase {
     file: StaticString = #file,
     line: UInt = #line
   ) {
-    // Use `CollectionDifference` on supported platforms to get `diff`-like line-based output. On
-    // older platforms, fall back to simple string comparison.
-    if #available(macOS 10.15, *) {
```
TTOzzi, line 218: I removed what seemed to be unnecessary OS conditionals. If the GitHub Actions CI passes, can we assume it works fine on Linux and Windows?

Package.swift
allevato: Someone added it a while back. I guess theoretically you could write an editor for iPad and want to format Swift code in it.

Sources/_SwiftFormatTestSupport/DiagnosingTestCase.swift
allevato: Yeah, as long as all the CI comes back green, that should be enough verification that the versions of Foundation we care about have the necessary APIs. (They should anyway, because of the `*` in the `#available` condition.)

TTOzzi:
It seems that `RegexBuilder` is not supported in Swift 5.9 on Windows.. 😨

Sources/SwiftFormat/Core/RuleMask.swift
```diff
@@ -111,10 +111,11 @@ enum IgnoreDirective: CustomStringConvertible {
...
   }
 
-  /// Regex pattern to match an ignore comment. This pattern supports 0 or more comma delimited rule
-  /// names. The rule name(s), when present, are in capture group #3.
-  fileprivate var pattern: String {
-    return #"^\s*\/\/\s*"# + description + #"((:\s+(([A-z0-9]+[,\s]*)+))?$|\s+$)"#
+  /// Regex pattern to match an ignore directive comment.
+  /// - Capture group #1 captures the rule names if `":"` is present.
+  fileprivate func makeRegex() -> Regex<(Substring, Substring?)> {
+    let pattern = #"^\s*\/\/\s*"# + description + #"(?:\s*:\s*(.+))?$"#
```
allevato, line 117: If we can't use RegexBuilder here because of platform limitations, it would still be nice to use real regex literals.

I don't remember from the previous PR, but if the `description` property is only used here, you could lift the regexes up into that instead, something like:

```swift
var regex: Regex {
  switch self {
  case .node: return /...swift-format-ignore.../
  case .file: return /...swift-format-ignore-file.../
}
```

allevato: It would also be great to use named capture groups so that the result tuples elsewhere get labels, to make it more readable.

Sources/SwiftFormat/Core/RuleMask.swift
```diff
@@ -111,10 +111,11 @@ enum IgnoreDirective: CustomStringConvertible {
...
+    return try! Regex(pattern)
```
ahoppen, line 118: Could you add a comment that we can’t use regex literals here because Windows didn’t have support for regex literals until Swift 5.10 (at least I think it was 5.10)? Just so we don’t go back and wonder why we aren’t using regex literals  here.

ahoppen (commented):
Nice

Sources/_SwiftFormatTestSupport/DiagnosingTestCase.swift
TTOzzi: Ah, since there is a `*` it seems that it was already working on Windows and Linux 😅
Thank you for letting me know.

Sources/SwiftFormat/Core/RuleMask.swift
TTOzzi: `description` is used in `OrderedImports` to distinguish directives. 
Also, after modifying it to use regex literals and running the CI, I found that, as @ahoppen mentioned below, regex literals are not supported on Windows Swift 5.9 (5.10). Given this, it seems best to keep the current implementation as is.

TTOzzi: > It would also be great to use named capture groups so that the result tuples elsewhere get labels, to make it more readable.

I have utilized the named capture group in the regex to add the label `ruleNames`.

Sources/SwiftFormat/Core/RuleMask.swift
TTOzzi: Sure, I've added it.

Sources/SwiftFormat/Core/RuleMask.swift
allevato: Every time we try to do some good, Windows is there lurking in the background. 🙁 

Thanks for checking!

allevato (approved):
Thanks!

ahoppen:
This change regressed performance of swift-format by 3% in release mode

To reproduce this regression, run

```
git clone https://github.com/apple/swift-syntax.git
cd swift-syntax
git checkout -f 509.0.0

.build/release/swift-format lint --measure-instructions --recursive path/to/above/checkout/of/swift-syntax 2> /dev/null
```

@TTOzzi Could you check if there’s a good way to recover that performance regression? I suspect that it’s because `Regex` is slower than `NSRegularExpression`. If there isn’t a way to recover the regression, I think we should stick with `NSRegularExpression`.

TTOzzi:
> This change regressed performance of swift-format by 3% in release mode
> 
> To reproduce this regression, run
> 
> ```
> git clone https://github.com/apple/swift-syntax.git
> cd swift-syntax
> git checkout -f 509.0.0
> 
> .build/release/swift-format lint --measure-instructions --recursive path/to/above/checkout/of/swift-syntax 2> /dev/null
> ```
> 
> @TTOzzi Could you check if there’s a good way to recover that performance regression?

Oh, I see! Thank you for letting me know.
I didn't think to check the performance at the time 😓 
I'll look into it by this weekend and open a PR 🙇

ahoppen:
Thank you ❤️

PR #962, opened 2025-03-14
Unify whitespace handling in linter and formatter

### LINKED ISSUE #960: Crash: WhitespaceLinter.contiguousWhitespace(startingAt:in:)
When applying swift-format to my project it crashes with following stack trace:

```
💣 Program crashed: System trap at 0x0000aaaac5d215e0

Thread 0 "swift-format" crashed:

 0 WhitespaceLinter.contiguousWhitespace(startingAt:in:) in swift-format at /home/build-user/swift-format/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift:327:11
 1 WhitespaceLinter.lint() + 660 in swift-format at /home/build-user/swift-format/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift:56:33
 2 SwiftLinter.lint(syntax:operatorTable:assumingFileURL:source:) + 715 in swift-format at /home/build-user/swift-format/Sources/SwiftFormat/API/SwiftLinter.swift:155:8
 3 SwiftLinter.lint(source:assumingFileURL:parsingDiagnosticHandler:) + 307 in swift-format at /home/build-user/swift-format/Sources/SwiftFormat/API/SwiftLinter.swift:101:9
 4 LintFrontend.processFile(_:) + 555 in swift-format at /home/build-user/swift-format/Sources/swift-format/Frontend/LintFrontend.swift:33:18
 5 specialized Sequence.forEach(_:) + 2491 in swift-format at /home/build-user/swift-format/Sources/swift-format/Frontend/Frontend.swift:153:18
 6 implicit closure #7 in Frontend.processURLs(_:parallel:) in swift-format at /home/build-user/swift-format/Sources/swift-format/Frontend/Frontend.swift:153:10
 7 Frontend.processURLs(_:parallel:) + 699 in swift-format at /home/build-user/swift-format/Sources/swift-format/Frontend/Frontend.swift:153:18
 8 Frontend.run() + 199 in swift-format at /home/build-user/swift-format/Sources/swift-format/Frontend/Frontend.swift:102:7
 9 closure #1 in SwiftFormatCommand.Lint.run() + 123 in swift-format at /home/build-user/swift-format/Sources/swift-format/Subcommands/Lint.swift:37:18
10 specialized PerformanceMeasurementsOptions.printingInstructionCountIfRequested<A>(_:) in swift-format at Sources/swift-format/Subcommands/PerformanceMeasurement.swift:24:18
11 SwiftFormatCommand.Lint.run() + 175 in swift-format at /home/build-user/swift-format/Sources/swift-format/Subcommands/Lint.swift:35:41
12 0x0000ffff938f84c4 <unknown> in libc.so.6
13 0x0000ffff938f8598 <unknown> in libc.so.6
```

Here's how you can reproduce the crash:
```sh
# Clone the repository
git clone https://github.com/seemoo-lab/CellGuard.git
# Run swift-format included with the latest swift Docker container
docker run --rm -it -v $(pwd)/CellGuard:/app:ro swift:latest
# Inside the container go to the mounted directory
cd /app/CellGuardAppSwift
# Run swift-format
swift format lint -r .
# At some point, you'll encounter this crash
```

Best
Lukas

TTOzzi:
Resolve #960 

In Xcode, using certain unicode characters (such as `U+2028`) in code causes errors. Following the same logic, I modified the linter so that it no longer treats these characters as valid and instead emits a message prompting their removal. Comments containing these characters will also no longer cause crashes when lint.

To ensure scalability, I introduced a `UnicodeException` enum that maintains a collection of unprocessable unicode characters. When adding a new unexpected unicode character in the future, simply defining a new case and its corresponding `utf8Bytes` will allow it to be properly handled as an unexpected unicode character.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/Comment.swift
+++ b/Sources/SwiftFormat/PrettyPrint/Comment.swift
@@ -21,13 +21,27 @@ extension StringProtocol {
   /// - Returns: The string with trailing whitespace removed.
   func trimmingTrailingWhitespace() -> String {
     if isEmpty { return String() }
-    let scalars = unicodeScalars
-    var idx = scalars.index(before: scalars.endIndex)
-    while scalars[idx].properties.isWhitespace {
-      if idx == scalars.startIndex { return String() }
-      idx = scalars.index(before: idx)
+    let utf8Array = Array(utf8)
+    var idx = utf8Array.endIndex - 1
+    while utf8Array[idx].isWhitespace {
+      if idx == utf8Array.startIndex { return String() }
+      idx -= 1
     }
-    return String(String.UnicodeScalarView(scalars[...idx]))
+    return String(decoding: utf8Array[...idx], as: UTF8.self)
+  }
+}
+
+extension UTF8.CodeUnit {
+  /// Checks if the UTF-8 code unit represents a whitespace character.
+  ///
+  /// - Returns: `true` if the code unit represents a whitespace character, otherwise `false`.
+  var isWhitespace: Bool {
+    switch self {
+    case UInt8(ascii: " "), UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\r"), /*VT*/ 0x0B, /*FF*/ 0x0C:
+      return true
+    default:
+      return false
+    }
   }
 }
 
--- a/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift
+++ b/Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift
@@ -339,16 +339,8 @@ public class WhitespaceLinter {
     startingAt offset: Int,
     in data: [UTF8.CodeUnit]
   ) -> ArraySlice<UTF8.CodeUnit> {
-    func isWhitespace(_ char: UTF8.CodeUnit) -> Bool {
-      switch char {
-      case UInt8(ascii: " "), UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\r"), /*VT*/ 0x0B, /*FF*/ 0x0C:
-        return true
-      default:
-        return false
-      }
-    }
     guard
-      let whitespaceEnd = data[offset...].firstIndex(where: { !isWhitespace($0) })
+      let whitespaceEnd = data[offset...].firstIndex(where: { !$0.isWhitespace })
     else {
       return data[offset..<data.endIndex]
     }
--- a/Tests/SwiftFormatTests/PrettyPrint/CommentTests.swift
+++ b/Tests/SwiftFormatTests/PrettyPrint/CommentTests.swift
@@ -1094,4 +1094,17 @@ final class CommentTests: PrettyPrintTestCase {
 
     assertPrettyPrintEqual(input: input, expected: input, linelength: 80)
   }
+
+  func testUnexpectedUnicodeCharacters() {
+    let input =
+      """
+      // Hello World\u{2028}
+      // Hello\u{20}\u{2028}World
+      // Hello World\u{2028}\u{2029}\u{2029}
+      // Hello World\u{20}\u{20}\u{20}\u{2028}
+
+      """
+
+    assertPrettyPrintEqual(input: input, expected: input, linelength: 80)
+  }
 }
--- a/Tests/SwiftFormatTests/PrettyPrint/WhitespaceLintTests.swift
+++ b/Tests/SwiftFormatTests/PrettyPrint/WhitespaceLintTests.swift
@@ -255,4 +255,29 @@ final class WhitespaceLintTests: WhitespaceTestCase {
       ]
     )
   }
+
+  func testUnexpectedUnicodeCharacters() {
+    assertWhitespaceLint(
+      input: """
+        // Hello World\u{2028}
+        // Hello\u{20}\u{2028}World
+        // Hello World\u{2028}\u{2029}\u{2029}
+        // Hello World              \u{2028}
+        // Hello World\u{2028}1️⃣\u{20}\u{20}\u{20}
+
+        """,
+      expected: """
+        // Hello World\u{2028}
+        // Hello\u{20}\u{2028}World
+        // Hello World\u{2028}\u{2029}\u{2029}
+        // Hello World              \u{2028}
+        // Hello World\u{2028}
+
+        """,
+      linelength: 30,
+      findings: [
+        FindingSpec("1️⃣", message: "remove trailing whitespace")
+      ]
+    )
+  }
 }


ahoppen (commented):
Maybe one high-level question: Why do we need to perform a preliminary validation pass over the input? Wouldn’t it be possible to resolve the crash at the point where it’s happening – I haven’t reproduced the issue myself, so this might not be possible, but I would like to explore that option.

TTOzzi:
> Maybe one high-level question: Why do we need to perform a preliminary validation pass over the input? Wouldn’t it be possible to resolve the crash at the point where it’s happening – I haven’t reproduced the issue myself, so this might not be possible, but I would like to explore that option.

My understanding is that `WhitespaceLinter` generates diagnostics by comparing the formatted text, processed by `PrettyPrinter`, with the original text, specifically focusing on removed whitespace characters.

During linting, it iterates through the text step by step, incrementing the index, and calls `contiguousWhitespace` to extract non-whitespace portions from both the original and formatted text.
Under normal circumstances, the non-whitespace content of both texts should be identical, meaning no issues should arise. However, in this issue, the original text contains more characters that are not recognized as whitespace compared to the formatted text. As a result, during iteration, formattedIndex eventually exceeds the valid range of formattedText, leading to an out-of-bounds issue.

Given this, I believe the current behavior of the linter is reasonable.


After reviewing your comment and re-examining the behavior, I found that the root cause is the discrepancy in whitespace detection criteria between `PrettyPrinter` and `WhitespaceLinter`.
When `PrettyPrinter` generates `formattedText`, it removes trailing whitespace using [`trimmingTrailingWhitespace`](https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/PrettyPrint/Comment.swift#L17-L32), which detects whitespace at the String level. This process removes characters like U+2028.
However, WhitespaceLinter converts the string into `[UTF8.CodeUnit]` and checks for whitespace at the byte level. As a result, it fails to recognize `U+2028` as a single whitespace character and instead processes its UTF-8 representation `[0xE2, 0x80, 0xA8]` as three separate units, failing to filter it out correctly.

IMO, the best approach would be to align the whitespace detection logic between PrettyPrinter and WhitespaceLinter 🤔 
Let me revise this again!

Sources/SwiftFormat/PrettyPrint/WhitespaceLinter.swift
```diff
@@ -339,20 +340,15 @@ public class WhitespaceLinter {
...
+    // To ensure consistency with PrettyPrinter, the data is converted to a String
+    // before processing whitespace.
+    let substring = String(decoding: data[offset...], as: UTF8.self)
+    var stringIndex = substring.startIndex
+    while stringIndex < substring.endIndex, substring[stringIndex].isWhitespace {
+      substring.formIndex(after: &stringIndex)
     }
-    return data[offset..<whitespaceEnd]
+    let utf8Count = substring.utf8.distance(from: substring.startIndex, to: stringIndex)
+    return data[offset..<offset + utf8Count]
```
TTOzzi, line 351: I modified the logic to determine whitespace to match the one used in [PrettyPrinter](https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/PrettyPrint/Comment.swift#L17-L32).
However... after looking at https://github.com/swiftlang/swift-format/pull/915 and running performance tests, I found that it became significantly slower 🫠
```
before: 50321785572
after: 4392453228823
```
I think I'll need to look for an alternative approach.

allevato:
From some quick testing, it looks like this problem only repros when the U+2028 character is at the _end_ of a line. That looks like it throws off all of the subsequent whitespace calculations by one line, causing it to crash at the end. If it's in the middle of a line, it executes fine. (At least, that's the only condition where I could cause it to crash.)

What's more interesting is that in format mode, something in the formatter is _removing_ that U+2028, because it's not in the output. I didn't track down exactly where it's happening, though.

So I think this might be easier to solve. If we can figure out where the U+2028 is being dropped, perhaps we could just keep it? If someone wants to have one of those in a comment, the compiler doesn't care, so maybe we shouldn't either. Then, all the offsets will line up as expected.

It's likely that there's code somewhere that's trimming "whitespace" (where whitespace is defined in the Unicode sense), causing it to cover more possibilities than what whitespace linter expects. If we make that code only recognize ASCII whitespace, that might fix it.

TTOzzi:
> From some quick testing, it looks like this problem only repros when the U+2028 character is at the _end_ of a line. That looks like it throws off all of the subsequent whitespace calculations by one line, causing it to crash at the end. If it's in the middle of a line, it executes fine. (At least, that's the only condition where I could cause it to crash.)
> 
> What's more interesting is that in format mode, something in the formatter is _removing_ that U+2028, because it's not in the output. I didn't track down exactly where it's happening, though.
> 
> So I think this might be easier to solve. If we can figure out where the U+2028 is being dropped, perhaps we could just keep it? If someone wants to have one of those in a comment, the compiler doesn't care, so maybe we shouldn't either. Then, all the offsets will line up as expected.
> 
> It's likely that there's code somewhere that's trimming "whitespace" (where whitespace is defined in the Unicode sense), causing it to cover more possibilities than what whitespace linter expects. If we make that code only recognize ASCII whitespace, that might fix it.

You're right. The issue is that the formatter and linter have different definitions of whitespace. The formatter considers `U+2028` as whitespace and removes it [here](https://github.com/swiftlang/swift-format/blob/main/Sources/SwiftFormat/PrettyPrint/Comment.swift#L17-L32).

I initially thought the linter should align with the formatter's behavior, so I tried addressing it [here](https://github.com/swiftlang/swift-format/pull/962#discussion_r1996668210). But if, as you mentioned, there's no real need to remove `U+2028`, then it might make more sense to adjust the formatter to match the linter instead.

I'll revise the approach accordingly. Thanks for sharing your thoughts!

Sources/SwiftFormat/PrettyPrint/Comment.swift
```diff
@@ -21,13 +21,27 @@ extension StringProtocol {
...
-      idx = scalars.index(before: idx)
+    let utf8Array = Array(utf8)
+    var idx = utf8Array.endIndex - 1
+    while utf8Array[idx].isWhitespace {
+      if idx == utf8Array.startIndex { return String() }
+      idx -= 1
+    }
+    return String(decoding: utf8Array[...idx], as: UTF8.self)
+  }
+}
```
TTOzzi, line 32: I have updated the formatter's whitespace criteria to match the linter's.
The formatter no longer considers characters like `U+2028` as whitespace, so they will no longer be removed.

This change has no significant impact on performance.
```
.build/release/swift-format lint --measure-instructions --recursive /tmp/swift-syntax

before: 50321785572
after: 49557836028
```

allevato (commented):
Can you add a test or two to WhitespaceLintTests to make sure we don't accidentally regress this in the future? We don't need to worry about the situation where a character like U+2028 would appear outside of a comment (or a string literal I guess) because that would be a compiler error, but it would be good to have a test that puts one at the end of a comment line and make sure it doesn't blow up.

That does mean that if we had a comment like this:

```swift
// foo bar baz    <U+2028>
```

the whitespace after `baz` wouldn't be recognized as trailing whitespace. I guess that's "fine"; if we ever wanted to do something more Unicode-smart in the future, we'd want to do it holistically through the whole formatter.

TTOzzi:
> Can you add a test or two to WhitespaceLintTests to make sure we don't accidentally regress this in the future? We don't need to worry about the situation where a character like U+2028 would appear outside of a comment (or a string literal I guess) because that would be a compiler error, but it would be good to have a test that puts one at the end of a comment line and make sure it doesn't blow up.
> 
> That does mean that if we had a comment like this:
> 
> ```swift
> // foo bar baz    <U+2028>
> ```
> 
> the whitespace after `baz` wouldn't be recognized as trailing whitespace. I guess that's "fine"; if we ever wanted to do something more Unicode-smart in the future, we'd want to do it holistically through the whole formatter.

Yes, sounds good. I've added a test case to `WhitespaceLintTests` to demonstrate that the linter does not treat characters like `U+2028` as whitespace.

ahoppen (commented):
Nice. Thank you @TTOzzi, this is a great solution and the improved performance is an amazing bonus 🚀

PR #963, opened 2025-03-16
Import Android for isatty

marcprux:


Sources/swift-format/Utilities/TTY.swift
```diff
@@ -11,6 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+#if canImport(Android)
```
compnerd, line 14: I tend to prefer `#if os(Android)` - this isn't about importing the module if it is available, this is about OS specific support.

allevato:
The format check is failing; can you run swift-format on the change?

I'm curious what this is for; are you actually planning to target Android with swift-format, or is this just to get around a SwiftPM issue where it incorrectly likes to build plugins for the target platform instead of the host?

marcprux:
> I'm curious what this is for

I'm wanting to use the swift-format API from Android app.

allevato (approved):

Issue #964, opened 2025-03-18
Configure PrettyPrinterRules


bkolb:
It should be allowed to enable/disable pretty printer rules

ahoppen:
Synced to Apple’s issue tracker as rdar://147314999

PR #965, opened 2025-03-18
Add support for pretty print categories in rule registry generation

bkolb:
Issue: #964


bkolb:
I close this for now. this will need more work to be done properly.

PR #971, opened 2025-03-28
Update for swift-syntax AttributeSyntax.Arguments changes

rintaro:
Some specialized `AttributeSyntax.Arguments` cases and related types are being removed from swift-syntax.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -1880,11 +1880,6 @@ fileprivate final class TokenStreamCreator: SyntaxVisitor {
     return .visitChildren
   }
 
-  override func visit(_ node: ExposeAttributeArgumentsSyntax) -> SyntaxVisitorContinueKind {
-    after(node.comma, tokens: .break(.same, size: 1))
-    return .visitChildren
-  }
-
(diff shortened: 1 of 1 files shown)


allevato:
Is that the only place in TokenStreamCreator that's actually affected? I thought we covered more of the attributes you mentioned. Maybe it's not as big a savings as I thought it would be. 🙁  (We should still do it though.)

We have a little bit of a chicken-and-egg problem here since the tests will fail unless this is built against a version of swift-syntax that has the parser changes to produce argument lists instead of the old nodes. Do we support paired repos in the new PR testing workflow?

rintaro:
https://github.com/swiftlang/swift-syntax/pull/3028
https://github.com/swiftlang/swift/pull/80351
@swift-ci Please test

rintaro:
Ah no `@swift-ci` PR testing anymore? But at least we can test it from `swift-syntax` repo 
https://github.com/swiftlang/swift-syntax/pull/3028#issuecomment-2761512146

rintaro:
> Is that the only place in TokenStreamCreator that's actually affected? I thought we covered more of the attributes you mentioned. 

As far as I tested locally 🙂 Many are underscored attributes, no wonder they weren't handled.

ahoppen:
Let me try and push for a review on https://github.com/swiftlang/github-workflows/pull/52 and allow cross-PR testing for GitHub actions. In the meantime, we’ll have to find someone with force merge capabilities to merge this.

bnbarham (approved):

PR #973, opened 2025-03-31
Correctly format `unsafe` expressions.

### LINKED ISSUE #972: swift-format can't treat `unsafe` expressions correctly
## Summary

In [SE-0458](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md), the new `unsafe` expression has been added. However, swift-format still can't handle it correctly.

## Prerequisites

- swift-format main branch 88a8900f4fd7902367261756575fdd0039e8aa58
- swift-syntax main branch [94681bad0aa2ef3a277d2895e7f0c9e60ec2bbb1](https://github.com/swiftlang/swift-syntax/commit/94681bad0aa2ef3a277d2895e7f0c9e60ec2bbb1)

## Steps to reproduce

1. Create `test.swift`
    ```swift
    _ = unsafe UnsafeRawPointer(bitPattern: 0)
    ```
2. Run `swift-format format test.swift`

## Expected output

```swift
_ = unsafe UnsafeRawPointer(bitPattern: 0)
```

## Actual output

It breaks the source code.

```swift
_ = unsafeUnsafeRawPointer(bitPattern: 0)
```

allevato:
Since this is another "stackable keyword-modified expression" situation, I've given the same kind of handling as `try` and `await`, where we have a group that tries to encompass the entire sequence of keywords and the following token (if it is an identifier in a callable non-compound expression) to avoid breaking unnecessarily in those situations.

The only subtle difference is that `unsafe` is a contextual keyword, so it _must not_ be separated from the following identifier/keyword by a line break. We use a fixed space to keep it glued to the following token in all circumstances.

I've shuffled the pretty-printer tests for `try`/`await` around a bit so that we have a single file to handle the behavior for all of these keyword-modified expressions. I think it's cleaner.

Fixes #972.


bnbarham (approved):
Nice!

PR #975, opened 2025-04-03
Update publish_release action to commit and push PrintVersion.swift changes if needed

### LINKED ISSUE #834: Output of `swift-format --version` is confusing
I am using Xcode 16 which comes with a bundled swift format. When I run `swift-format --version`, it prints `main`.
Does this mean the swift format bundled with Xcode will always be updated with the latest main branch? Or the bundled version is the latest from the main branch at the time of Xcode release is created?

I tried installing swift-format with homebrew and `swift-format --version` prints the version correctly.

TTOzzi:
Related to #834 

We had hoped that the version print issue would be fixed in the Swift 6.1 release, but unfortunately, it seems that it hasn’t been addressed.

As I understand it, swift-toolchain references swift-format by a specific branch(https://github.com/swiftlang/swift/blob/release/6.1/utils/update_checkout/update-checkout-config.json#L211), so the current behavior of the `publish_release` action — where the change to `PrintVersion.swift` is included only in a tag and not pushed to a branch — means that the update doesn’t make it into the distributed toolchain 🤔 

This PR updates the `publish_release` action so that when `PrintVersion.swift` is updated, the change is committed and pushed to the branch as well.

It seems like another possible approach could be to have `update-checkout-config` in Swift point to a tag of swift-format instead of a branch — but I’m not sure which option would be better.

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -67,6 +67,25 @@ jobs:
           echo "swift_format_version=$SWIFT_FORMAT_VERSION" >> "$GITHUB_OUTPUT"
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Create version commit if version changed
+        run: |
+          # Without this, we can't perform git operations in GitHub actions.
+          git config --global --add safe.directory "$(realpath .)"
+          git config --local user.name 'swift-ci'
+          git config --local user.email 'swift-ci@users.noreply.github.com'
+
+          BRANCH_NAME="${GITHUB_REF#refs/heads/}"
+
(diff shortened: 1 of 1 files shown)


bnbarham:
Unfortunately this is unlikely to have the intended effect, since `publish_release` is going to run around (and possibly after) the toolchain release. It's clear we need to do something here though 🤔

TTOzzi:
> Unfortunately this is unlikely to have the intended effect, since `publish_release` is going to run around (and possibly after) the toolchain release. It's clear we need to do something here though 🤔

Oh, I see — so `publish_release` runs after the toolchain release…
We do run a prerelease using the same workflow before the official release, but since it uses a different version name (e.g. `601.0.0-prerelease-2025-03-28`), that probably won’t help here.
Looks like we’ll need to explore a different approach 🤔

PR #976, opened 2025-04-03
Add concurrency control to cancel in-progress PR workflows

TTOzzi:
When updating a pull request that's already open, each new commit can trigger a workflow. If the PR author doesn't have permission to cancel workflows, this can lead to unnecessary executions.
To avoid redundant executions and reduce resource usage, this change cancels any in-progress workflow for the same pull request when a new one is triggered by a commit push, ensuring only one active workflow per PR.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   tests:
     name: Test


TTOzzi:
If this change looks good, I think we could also consider applying the same setup to the `pull_request` workflow in `swift-syntax`.

shahmishal:
We should look at adding this directly in the swiftlang/github-workflow repository.

shahmishal:
However, it's ok to take this now in this repository.

TTOzzi:
> We should look at adding this directly in the swiftlang/github-workflow repository.

Yes, as you mentioned, it would be even better if we could apply this change across the board in `swiftlang/github-workflows`.
Since it's already applied to `soundness` (https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/soundness.yml#L83-L86), I think we only need to apply it to `swift_package_test`.
Would it be okay if I open a PR to `github-workflows` for that?

shahmishal:
> Would it be okay if I open a PR to github-workflows for that?

Yes, that would be a great help. Thank you!

TTOzzi:
I’ve opened a replacement PR in `swiftlang/github-workflows`, so I’m closing this one. 
Thank you @shahmishal 🙇

TTOzzi:
Sorry about that — I tried making changes over in `swiftlang/github-workflows`, but with the functionality currently provided by GitHub, it seems impossible to fully support all the places where `swift_package_test` is used.
So I've reopened this 😓

bnbarham (approved):

bnbarham:
Happy to take this while we work out something for github-workflows regardless. Same for `swift-syntax`. Thanks @TTOzzi!

PR #1004, opened 2025-04-24
Missing dependency

ahoppen:


ahoppen:
It’s just something I noticed while working on some other build infrastructure for swift-format. It isn’t an actual issue in SwiftPM because there is a transitive dependency swift-format -> SwiftFormat -> SwiftParserDiagnostics -> SwiftDiagnostics. But I think it’s cleaner to make it explicit.

Issue #703, opened 2024-03-15
OrderedImports should create separate groups for different access levels


allevato:
Now that `public/internal/private/etc import` is supported, we should have the `OrderedImports` rule do separate groupings for each access level. Order the groups in descending order of access level (`public`, `package`, `internal`, `fileprivate/private`), with each group sorted lexicographically as usual.

Other details:

* If a user has both `internal import` and `import`, do we unify them by adding/removing `internal`? We don't currently add or drop `internal` anywhere else, so we don't have a precedent here.
* We do need to respect the `fileprivate/private` configuration that's used for other file-scoped declarations.

ahoppen:
Tracked in Apple’s issue tracker as rdar://126948323

allevato:
If we're planning to do it for `internal/private/etc` imports, doing it also for `@_implementationOnly` seems reasonable too.

PR #1013, opened 2025-05-06
OrderedImports should create separate group for @_implementationOnly …

ebgraham:
The OrderedImports rule should create a separate import grouping for `@_implementationOnly` imports to go with the current `@testable` group. The code is refactored a bit to enable a future change for public/internal/private import grouping, and to harden again other import attributes like @preconcurrency.

Discussed in [#703](https://github.com/swiftlang/swift-format/issues/703).

### DIFF
--- a/api-breakages.txt
+++ b/api-breakages.txt
@@ -18,4 +18,4 @@ API breakage: func Configuration.MultilineStringReflowBehavior.hash(into:) has b
 API breakage: func Configuration.MultilineStringReflowBehavior.encode(to:) has been removed
 API breakage: var Configuration.MultilineStringReflowBehavior.hashValue has been removed
 API breakage: constructor Configuration.MultilineStringReflowBehavior.init(from:) has been removed
-API breakage: enum LineType.implementationOnlyImport has been added as a new enum case
+API breakage: enumelement LineType.implementationOnlyImport has been added as a new enum case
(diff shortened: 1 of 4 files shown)


allevato (commented):
Can you make sure to run swift-format on the changes, and add the new enum case to `api-breakages.txt`?

@ahoppen @bnbarham This will likely impact how swift-syntax's sources are formatted. For changes like this, do we just need to fast-follow a change to swift-syntax after this is submitted?

api-breakages.txt
```diff
@@ -18,3 +18,4 @@ API breakage: func Configuration.MultilineStringReflowBehavior.hash(into:) has b
 API breakage: func Configuration.MultilineStringReflowBehavior.encode(to:) has been removed
 API breakage: var Configuration.MultilineStringReflowBehavior.hashValue has been removed
 API breakage: constructor Configuration.MultilineStringReflowBehavior.init(from:) has been removed
+API breakage: enum LineType.implementationOnlyImport has been added as a new enum case
```
allevato, line 21: This check is still failing; the message needs to be copied verbatim from the error in the CI log:

`API breakage: enumelement LineType.implementationOnlyImport has been added as a new enum case`

(`enumelement`, not `enum`)

allevato (approved):
Thanks for adding this!

PR #1014, opened 2025-05-07
Allow comments in `.swift-format`

### LINKED ISSUE #1011: Allow comments in `.swift-format` files
Allow comments in `.swift-format` files. Use JSON5 or some other format that supports comments.

antigluten:
Enable [allowsJSON5](https://developer.apple.com/documentation/foundation/jsondecoder/allowsjson5) on the `JSONDecoder`. With this change, the decoder can now parse JSON5-formatted data.

Resolves #1011.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -293,7 +293,9 @@ public struct Configuration: Codable, Equatable {
   /// Creates a new `Configuration` by decoding it from the UTF-8 representation in the given data.
   public init(data: Data) throws {
     let jsonDecoder = JSONDecoder()
+    #if compiler(>=6)
     jsonDecoder.allowsJSON5 = true
+    #endif
     self = try jsonDecoder.decode(Configuration.self, from: data)
   }
 
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -293,7 +293,7 @@ public struct Configuration: Codable, Equatable {
   /// Creates a new `Configuration` by decoding it from the UTF-8 representation in the given data.
   public init(data: Data) throws {
     let jsonDecoder = JSONDecoder()
-    #if compiler(>=6)
+    #if canImport(Darwin) || compiler(>=6)
     jsonDecoder.allowsJSON5 = true
     #endif
     self = try jsonDecoder.decode(Configuration.self, from: data)
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -292,7 +292,9 @@ public struct Configuration: Codable, Equatable {
 
   /// Creates a new `Configuration` by decoding it from the UTF-8 representation in the given data.
   public init(data: Data) throws {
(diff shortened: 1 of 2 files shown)


ahoppen:
I would prefer to check for a compiler version on non-Darwin OSs because that’s something that’s easier to remove in the future. It’s easy to tell that `#if compiler(>=6)` will always evaluate to true if we no longer support building swift-format using a Swift 5 compiler. But for `#if canImport(FoundationEssentials)` is a bit harder to determine if that’s always true because you need to remember when `FoundationEssentials` was introduced.

antigluten:
Hi @allevato @ahoppen, added checking for compiler version.

Tests/SwiftFormatTests/API/ConfigurationTests.swift
```diff
@@ -99,9 +106,30 @@ final class ConfigurationTests: XCTestCase {
...
+
+    let jsonData = """
+      {
+          // Indicates the configuration schema version.
+          "version": 1,
+      }
+      """.data(using: .utf8)!
+
+    let jsonDecoder = JSONDecoder()
+    #if compiler(>=6)
```
allevato, line 129: This whole test needs to be guarded, not just this line, or it will fail on Swift 5.x.

Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -292,7 +292,11 @@ public struct Configuration: Codable, Equatable {
 
   /// Creates a new `Configuration` by decoding it from the UTF-8 representation in the given data.
   public init(data: Data) throws {
-    self = try JSONDecoder().decode(Configuration.self, from: data)
+    let jsonDecoder = JSONDecoder()
+    #if compiler(>=6)
```
allevato, line 296: This could be `#if canImport(Darwin) || compiler(>=6)`, because on Apple platforms this API existed before Swift 6 so there's no reason to restrict it there.

antigluten: ✅ Fixed

Tests/SwiftFormatTests/API/ConfigurationTests.swift
antigluten: ✅ Fixed

Tests/SwiftFormatTests/API/ConfigurationTests.swift
```diff
@@ -99,9 +106,32 @@ final class ConfigurationTests: XCTestCase {
...
+      #if canImport(Darwin) || compiler(>=6)
+      jsonDecoder.allowsJSON5 = true
+      #endif
+      let config = try jsonDecoder.decode(Configuration.self, from: jsonData)
       XCTAssertEqual(config.reflowMultilineStringLiterals, expectedBehavior)
     }
   }
 
+  func testConfigurationWithComments() throws {
+    #if !canImport(Darwin) && compiler(<6)
```
allevato, line 119: You'll need to wrap the rest of the body in an `#else` here, too, otherwise it will still try to compile the code below in the `compiler(<6)` case (since it's just unguarded code right now).

antigluten: ✅ Fixed

allevato (approved):
Thank you!

PR #1015, opened 2025-05-11
Ensure newline after trailing line comments to prevent formatting issues

### LINKED ISSUE #1003: Import with attribute with trailing comment incorrectly formatted
```
@foo // bar
import Baz
```
gets formatted to
```
@foo //bar import Baz
```

TTOzzi:
Resolve #1003 

When arranging attributes, they were previously placed on a single line regardless of the presence of line comments.
This change preserves the original behavior of keeping attributes on a single line whenever possible, but inserts a newline when a line comment is present to prevent compilation failures.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -146,6 +146,10 @@ public class PrettyPrinter {
   /// enabled (see ``isBreakingSuppressed``).
   private var allowSuppressedDiscretionaryBreaks = false
 
+  /// Overrides break suppression for line or doc comments, or similar cases where a line break is required.
+  /// Reset after handling the break.
+  private var shouldOverrideBreakSuppression = false
+
   /// The computed indentation level, as a number of spaces, based on the state of any unclosed
   /// delimiters and whether or not the current line is a continuation line.
   private var currentIndentation: [Indent] {
@@ -426,7 +430,8 @@ public class PrettyPrinter {
       case .soft(_, let discretionary):
         // A discretionary newline (i.e. from the source) should create a line break even if the
         // rules for breaking are disabled.
-        overrideBreakingSuppressed = discretionary && allowSuppressedDiscretionaryBreaks
+        overrideBreakingSuppressed =
+          shouldOverrideBreakSuppression || (discretionary && allowSuppressedDiscretionaryBreaks)
         mustBreak = true
       case .hard:
         // A hard newline must always create a line break, regardless of the context.
@@ -455,6 +460,7 @@ public class PrettyPrinter {
         outputBuffer.enqueueSpaces(size)
         lastBreak = false
       }
+      shouldOverrideBreakSuppression = false
 
     // Print out the number of spaces according to the size, and adjust spaceRemaining.
     case .space(let size, _):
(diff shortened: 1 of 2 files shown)


allevato:
Having to resolve this in the attribute case specifically feels like the wrong approach. There should _never_ be a situation where a `//` or `///` comment is followed by a break that would do anything *but* force a newline. Can you check that instead?

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -479,6 +479,9 @@ public class PrettyPrinter {
         }
       }
       outputBuffer.write(comment.print(indent: currentIndentation))
+      if wasEndOfLine, (comment.kind == .line || comment.kind == .docLine) {
```
allevato, line 482: I suppose this _works_, but I'm more worried about the token stream creation itself (I should have been more specific in my earlier comment). Usually the newline behavior is controlled by what kind of `break` is inserted into the token stream, and I would expect that a line comment would always be preceded by something whose newline behavior is `soft` (meaning always insert a newline, unless it would contribute more than the maximum number of blank lines allowed). The original issue would seem to imply that that's not the case, and I think we need to understand why.

TTOzzi: Thanks for the feedback!
The immediate cause of the ignored `soft` newline was the logic that suppresses line breaks via `printerControl` tokens when visiting an `ImportDeclSyntax`.

As a result, the following code triggers a formatting issue:
```swift
@foo // comment
@bar
import Baz
```
Whereas this code (with a non-import declaration) does not:

```swift
@foo // comment
@bar
let var = 0
```
Looking at #501, this suppression behavior appears to be intentional.
So instead of changing the suppression logic itself, I modified it to ignore suppression when the token immediately preceding the break is a `//` or `///` comment.

allevato:
I may need a little more time to think about this.

I think what we ultimately want is for the pretty printer to recognize that certain breaks *must* be present in valid code, even if printer control regions say to suppress them, because the idea is that those regions only suppress _elective_ line breaks, not _all_ of them.

That logic shouldn't rely on whether the previous token was a comment or anything like that; it should be generalized into the token stream's breaks itself.

TTOzzi:
I've updated the approach—how about this?

I think that required line breaks which must not be suppressed by printer control are already represented as `hard` breaks.
This revision is similar to initial attempt, but instead of handling it in the `PrettyPrinter`, I've updated the `TokenStreamCreator` to insert a `hard` break when a line comment is present. This ensures the break will not be suppressed by printer control.

Additionally, I added logic to prevent `soft` newlines from being counted as `hard` newlines when appending. This avoids unintentionally inflating the number of `hard` line breaks when one already exists.

TTOzzi:
I've updated the approach—how about this?

I think that required line breaks which must not be suppressed by printer control are already represented as `hard` breaks.  
This revision is similar to initial attempt, but instead of handling it in the `PrettyPrinter`, I've updated the `TokenStreamCreator` to insert a `hard` break when a line comment is present. This ensures the break will not be suppressed by printer control.

Additionally, I added logic to prevent `soft` newlines from being counted as `hard` newlines when appending. This avoids unintentionally inflating the number of `hard` line breaks when one already exists.

TTOzzi:
I think that line breaks which must not be suppressed by printer control are already represented as `hard` breaks, so I tried adding a `hard` line break when a line comment is present during comment handling.
However, this approach caused side effects in some edge cases, and handling those cases seemed to introduce additional complexity, so I decided to revert to the previous approach 🫠

allevato:
I'm going back through the backlog and noticing that we never resolved this (well, I never did 😛 ). Can you fix up the merge conflicts and we can land this?

TTOzzi:
> I'm going back through the backlog and noticing that we never resolved this (well, I never did 😛 ). Can you fix up the merge conflicts and we can land this?

Sure! Thanks for bringing this up 🙌

allevato (approved):

PR #1029, opened 2025-06-02
[6.2] Check version has been updated when publishing a release

bnbarham:
Rather than updating the version, check that it has been. We need the
update in the branch itself so that the toolchain version is also
updated.

Also update the version itself (currently main).


ahoppen (approved):
Nice idea

Sources/swift-format/PrintVersion.swift
```diff
@@ -12,5 +12,5 @@
 
 func printVersionInformation() {
   // TODO: Automate updates to this somehow.
-  print("main")
+  print("6.2.0")
```
TTOzzi, line 15: Wouldn't it be more appropriate if it were in the form of `"602.0.0"`?

.github/workflows/publish_release.yml
```diff
@@ -67,6 +67,9 @@ jobs:
           echo "swift_format_version=$SWIFT_FORMAT_VERSION" >> "$GITHUB_OUTPUT"
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Check version matches
+        run: |
+          grep '"${{ steps.swift_format_version.outputs.swift_format_version }}"' Sources/swift-format/PrintVersion.swift
```
TTOzzi, line 72: I think we should also take into account the prerelease format, such as `"602.0.0-prerelease-$(date +'%Y-%m-%d')"` 🤔

TTOzzi (commented):
Thank you for taking care of this!

Sources/swift-format/PrintVersion.swift
bnbarham: Yeah, I agree. That was my "which also means the check I added here doesn't work" since publishing the release is for the library and not executable. I'd like to do something here to avoid forgetting to update this, but not sure what that something is right now.

PR #1039, opened 2025-06-22
Fix line breaking for attributes in inheritance clauses by grouping them with their type

### LINKED ISSUE #1032: Poor line breaking around `@unchecked Sendable`
```swift
public class ProtocolAddress: ClonableHandleOwner<SignalMutPointerProtocolAddress>, @unchecked Sendable {
  // …
}
```

This admittedly-lengthy inheritance clause gets formatted as

```swift
public class ProtocolAddress: ClonableHandleOwner<SignalMutPointerProtocolAddress>, @unchecked
  Sendable
{
}
```

which is definitely not the best place for a line break. The general principle here is probably to keep attributes with the thing they're attached to when the attachee isn't at the start of a line.

TTOzzi:
Resolve #1032


Tests/SwiftFormatTests/PrettyPrint/FunctionDeclTests.swift
```diff
@@ -1129,8 +1129,8 @@ final class FunctionDeclTests: PrettyPrintTestCase {
           .InnerMember,
       ) {}
       func foo(
-        cmp: @escaping (R) ->
-          ()
+        cmp:
+          @escaping (R) -> ()
```
allevato, line 1133: Nice! I was hoping we'd find some others with that change.

allevato (approved):

Tests/SwiftFormatTests/PrettyPrint/FunctionDeclTests.swift
bnbarham: I believe this one is causing a bunch of differences in eg. sourcekit-lsp:
```
-    requestHandler: @Sendable @escaping (
-      RequestType, Workspace, LanguageService
-    ) async throws ->
+    requestHandler:
+      @Sendable @escaping (
+        RequestType, Workspace, LanguageService
+      ) async throws ->
```
(and others like it)

I wouldn't have expected any change there though 🤔? The `@Sendable @escaping (` still fits on the same line as `requestHandler:`

allevato: Yeah, the group around the whole type now forces the line break since the whole type doesn't fit. I personally prefer that behavior (citing The Rectangle Rule), because I think a case like the one in the test diff above is one where the new version is clearly better, even though the one you posted from SourceKit doesn't seem so bad.

The difference between those two might be that in the test, we took something that didn't fit on a single line and made it fit on a single line, so that's (unambiguously, IMO) an improvement. In the SourceKit example, that's not the case, so it's less of a big win.

A future version of the formatter could try to make decisions like that, but it would require backtracking, so I've tried to resist that complexity.

PR #1044, opened 2025-07-15
Introduce `multilineTrailingCommaBehavior` configuration

### LINKED ISSUE #946: Feature Request: Support for trailing commas in lists (SE-0439)
[SE-0439](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0439-trailing-comma-lists.md) adds support for trailing commas in _more_ places. This is new to Swift 6.1. 

When developing packages for multiple Swift versions it'll be easy to accidentally add a trailing comma which is fine locally but won't get caught on older Swift versions until tested in CI.

It'd be great to have a formatter rule which can enforce they are either always applied, never applied, or ignored.

TTOzzi:
Resolve #946 

Added a configuration to support the use of trailing commas extended by [SE-0439](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0439-trailing-comma-lists.md).

This configuration supports three values:

- `alwaysUsed `: always insert trailing commas in multiline lists
- `neverUsed`: always remove trailing commas
- `keptAsWritten`: preserve existing commas as-is

The default value is `keptAsWritten `.

The main concern was how to integrate this with the existing `multiElementCollectionTrailingCommas` configuration.
For now, `multilineTrailingCommaBehavior ` takes higher precedence, but to minimize its impact on `multiElementCollectionTrailingCommas`, it is implemented so that when set to the default value `keptAsWritten`, the behavior of `multiElementCollectionTrailingCommas` is preserved.

I’d appreciate any suggestions or feedback you may have.

### DIFF
--- a/Sources/SwiftFormat/API/Configuration.swift
+++ b/Sources/SwiftFormat/API/Configuration.swift
@@ -44,6 +44,7 @@ public struct Configuration: Codable, Equatable {
     case rules
     case spacesAroundRangeFormationOperators
     case noAssignmentInExpressions
+    case multilineTrailingCommaBehavior
     case multiElementCollectionTrailingCommas
     case reflowMultilineStringLiterals
     case indentBlankLines
@@ -172,6 +173,22 @@ public struct Configuration: Codable, Equatable {
 
   /// Contains exceptions for the `NoAssignmentInExpressions` rule.
   public var noAssignmentInExpressions: NoAssignmentInExpressionsConfiguration
+
+  /// Determines how trailing commas in comma-separated lists should be handled during formatting.
+  public enum MultilineTrailingCommaBehavior: String, Codable {
+    case alwaysUsed
+    case neverUsed
+    case keptAsWritten
+  }
+
+  /// Determines how trailing commas in multiline comma-separated lists are handled during formatting.
+  ///
+  /// This setting takes precedence over `multiElementCollectionTrailingCommas`.
+  /// If set to `.keptAsWritten` (the default), the formatter defers to `multiElementCollectionTrailingCommas`
+  /// for collections only. In all other cases, existing trailing commas are preserved as-is and not modified.
+  /// If set to `.alwaysUsed` or `.neverUsed`, that behavior is applied uniformly across all list types,
+  /// regardless of `multiElementCollectionTrailingCommas`.
+  public var multilineTrailingCommaBehavior: MultilineTrailingCommaBehavior
 
   /// Determines if multi-element collection literals should have trailing commas.
   ///
@@ -384,6 +401,9 @@ public struct Configuration: Codable, Equatable {
         forKey: .noAssignmentInExpressions
       )
       ?? defaults.noAssignmentInExpressions
+    self.multilineTrailingCommaBehavior =
+      try container.decodeIfPresent(MultilineTrailingCommaBehavior.self, forKey: .multilineTrailingCommaBehavior)
+      ?? defaults.multilineTrailingCommaBehavior
     self.multiElementCollectionTrailingCommas =
       try container.decodeIfPresent(
         Bool.self,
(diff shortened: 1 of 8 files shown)


allevato (commented):
Is there an abstraction we can come up to reduce the repetition in the individual visitors? Something like a helper function that accepts a generic parameter constrained to a collection of `WithTrailingCommaSyntax` elements and does the actual work.

TTOzzi:
> Is there an abstraction we can come up to reduce the repetition in the individual visitors? Something like a helper function that accepts a generic parameter constrained to a collection of `WithTrailingCommaSyntax` elements and does the actual work.

I was struggling because the point where a comma needs to be inserted differs for each syntax, but as I came to know about `WithTrailingCommaSyntax`, I came up with an idea!
I defined an abstract type and refactored the code accordingly. Thanks!

Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
```diff
@@ -686,15 +699,28 @@ public class PrettyPrinter {
...
+        let shouldHandleCommaDelimitedRegion: Bool? =
+          switch configuration.trailingCommasInMultilineLists {
+          case .always:
+            true
+          case .never:
+            false
+          case .ignore:
+            isCollection ? configuration.multiElementCollectionTrailingCommas : nil
+          }
+        if let shouldHandleCommaDelimitedRegion, shouldHandleCommaDelimitedRegion {
```
allevato, line 717: This could just be `if shouldHandleCommaDelimitedRegion == true`. Being written twice made me have to reread it a couple times to make sure they were referring to the same thing.

Sources/SwiftFormat/Core/SyntaxTraits.swift
```diff
@@ -65,3 +65,81 @@ extension ExprSyntax {
     return self.asProtocol(KeywordModifiedExprSyntaxProtocol.self) != nil
   }
 }
+
+/// Common protocol implemented by comma-separated lists whose elements
+/// support a `trailingComma`.
+protocol CommaSeparatedListSyntaxProtocol: SyntaxCollection where Element: WithTrailingCommaSyntax & Equatable {
+  /// Indicates whether the list should be treated as a collection for formatting.
+  /// If `true`, the list is affected by the `multiElementCollectionTrailingCommas` configuration.
+  var isCollection: Bool { get }
```
allevato, line 74: Hate to be nitpicky, but can you change this to `isCollectionLiteral`? `isCollection` feels like it could be talking about the node itself (which are all `Collection`s).

Or... does this have to be a protocol requirement at all? It looks like you could just pass it in as another parameter to `markCommaDelimitedRegion`, since it has to be called individually for each node type anyway. That might be cleaner (but the rename is still valid elsewhere that it's referenced in the pretty printer).

Sources/SwiftFormat/API/Configuration.swift
```diff
@@ -173,6 +174,22 @@ public struct Configuration: Codable, Equatable {
...
+  }
+
+  /// Determines how trailing commas in comma-separated lists are handled during formatting.
+  ///
+  /// This setting takes precedence over `multiElementCollectionTrailingCommas`.
+  /// If set to `.ignore` (the default), the formatter defers to `multiElementCollectionTrailingCommas`
+  /// for collections only. In all other cases, existing trailing commas are preserved as-is and not modified.
+  /// If set to `.always` or `.never`, that behavior is applied uniformly across all list types,
+  /// regardless of `multiElementCollectionTrailingCommas`.
+  public var trailingCommasInMultilineLists: TrailingCommasInMultilineLists
```
allevato, line 191: Looking at this more closely, this feels too similar to the name of the other property below. I don't know if `multilineLists` really communicates the bit about parameter lists, tuples, closure captures, etc.

What about `multilineTrailingCommaBehavior`? Then, let's rename the enum to `MultilineTrailingCommaBehavior` and have the cases be `alwaysUsed`, `neverUsed`, and `keptAsWritten`. I think that helps to better document what's going on.

Later on (not in this PR), we could add a fourth case: `onlyCollectionLiterals`, which could take over the behavior of the other Boolean setting, and deprecate that one.

TTOzzi: I understand. I’ll make sure to be careful when adding or changing rules. Thanks for explaining! 🙇

allevato (approved):
Thanks!

PR #1050, opened 2025-07-21
Include _GenerateSwiftFormat in buildOnlyTests

TTOzzi:


Sources/swift-format/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_executable(swift-format
   Utilities/StderrDiagnosticPrinter.swift
   Utilities/TTY.swift)
 target_link_libraries(swift-format PRIVATE
+  _GenerateSwiftFormat
```
bnbarham, line 30: `swift-format` doesn't depend on `_GenerateSwiftFormat` so this shouldn't be needed. We also never build generate-swift-format in cmake, so technically the added CMakeLists.txt isn't needed either. We could add an extra in generate-swift-format so that it *is* buildable from cmake as well, but that's not the PR testing issue.

The PR testing issue is that we set `SWIFTFORMAT_BUILD_ONLY_TESTS` which then does this:
```
if buildOnlyTests {
  products = []
  targets = targets.compactMap { target in
    guard target.isTest || target.name == "_SwiftFormatTestSupport" else {
      return nil
    }
    target.dependencies = target.dependencies.filter { dependency in
      if case .byNameItem(name: "_SwiftFormatTestSupport", _) = dependency {
        return true
      }
      return false
    }
    return target
  }
}
```

That needs to be updated to allow `_GenerateSwiftFormat` as well

TTOzzi: Ah thanks for letting me know. I wasn't aware of this :bow: 
I’ve updated it as you suggested.

bnbarham (approved):
Thanks @TTOzzi!

PR #1057, opened 2025-08-11
.swift-format: Specify an exhaustive configuration

AnthonyLatsis:
Mirrors https://github.com/swiftlang/swift-syntax/pull/3117.
___

Unexpectedly, an omitted rule tells swift-format to ignore it altogether rather than assume the default setting.
    
This change applies the default configuration, retaining current settings and disabling 3 default-enabled rules: `DoNotUseSemicolons`, `NoAccessLevelOnExtensionDeclaration`, and `DontRepeatTypeInStaticProperties`. The latter is disabled due to undesired effect:
    
```
warning: [DontRepeatTypeInStaticProperties] remove the suffix 'Paths' from the name of the variable 'envSearchPaths'
```

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,20 +1,76 @@
 {
+    "fileScopedDeclarationPrivacy": {
+        "accessLevel": "private"
+    },
     "indentConditionalCompilationBlocks": false,
+    "indentSwitchCaseLabels": false,
     "indentation": {
         "spaces": 2
     },
+    "lineBreakAroundMultilineExpressionChainComponents": false,
+    "lineBreakBeforeControlFlowKeywords": false,
     "lineBreakBeforeEachArgument": true,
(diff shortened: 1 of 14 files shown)


allevato (approved):
The configuration decoder explicitly uses `decodeIfPresent` and falls back to the default values if the key wasn't in the JSON: https://github.com/swiftlang/swift-format/blob/2a1d811e3e7871496dcdd4939d14661601e8f5cf/Sources/SwiftFormat/API/Configuration.swift#L317-L432

...ah, but that still means that any `rules` dictionary being present will be used as the entire thing, rather than assuming the default values for any rules not present. We should probably fix that.

In the meantime, no reason to block this PR on that fix.

bnbarham:
> We should probably fix that.

Indeed. Though it's probably something we need to do under a new "style" given how breaking it could be 😓

PR #1068, opened 2025-09-19
Add workflow to auto-update PrintVersion on release branch creation

TTOzzi:
This retries the work attempted in #975 and #1029.
When a release branch is created, this workflow automatically updates the version output value in PrintVersion.swift to match the release version, commits the change, and opens a PR.

If committing directly to the release branch is acceptable, that would be cleaner. Please let me know your thoughts, and I’ll adjust the workflow accordingly.

### DIFF
--- a/.github/workflows/auto_update_version.yml
+++ b/.github/workflows/auto_update_version.yml
@@ -1,0 +1,67 @@
+name: Update PrintVersion on release branch creation
+
+on:
+  create:
+    branches:
+      - "release/*"
+
+jobs:
+  update-version:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ github.token }}
+
+      - name: Extract version from branch name
+        run: |
+          BRANCH_NAME="${GITHUB_REF_NAME}"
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_update_version.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr create \
+            --base "${{ env.RELEASE_BRANCH }}" \
+            --head "${{ env.UPDATE_BRANCH }}" \
+            --title "[${{ env.VERSION }}] Update version to ${{ env.VERSION }}" \
+            --body "This PR updates the version in PrintVersion.swift to ${{ env.VERSION }}.
+
+          Automatically created when release branch ${{ env.RELEASE_BRANCH }} was created." \
```
bnbarham, line 56: How about the following:
> This PR was automatically opened by a GitHub action on creation of a release branch (${{ env.RELEASE_BRANCH }}). Review the version update and merge if correct, otherwise update the version manually.

?

bnbarham (commented):
Ooo, interesting idea! We do still bump without branch creation, but this would at least catch some of them.

Thoughts @shahmishal?

.github/workflows/auto_update_version.yml
TTOzzi: Yes, I’ve updated it. Thank you!

shahmishal (approved):

PR #1077, opened 2025-10-18
Order conditionally-compiled imports

### LINKED ISSUE #1027: Conditionally compiled imports not sorted
Imports outside of conditional compilation blocks are sorted, but inside, the ordering of imports is ignored. For example, this code:

```swift
import Zebras
import Apples
#if canImport(Darwin)
  import Foundation
  import Darwin
#elseif canImport(Glibc)
  import Glibc
  import Foundation
#endif
import Aardvarks
```

is currently formatted as:

```swift
import Aardvarks
import Apples
import Zebras

#if canImport(Darwin)
  import Foundation
  import Darwin
#elseif canImport(Glibc)
  import Glibc
  import Foundation
#endif
```
 leaving the Foundation imports out of order. The result should be:
```swift
import Aardvarks
import Apples
import Zebras

#if canImport(Darwin)
  import Darwin
  import Foundation
#elseif canImport(Glibc)
  import Foundation
  import Glibc
#endif
```

a7medev:
Recursively sort conditionally-compiled imports the same way other imports are ordered.

Conditionally-compiled imports are always placed after normal non-conditional imports as before, they're only sorted relative to the same `#if` block.

Resolves #1027.

### DIFF
--- a/Sources/SwiftFormat/Rules/OrderedImports.swift
+++ b/Sources/SwiftFormat/Rules/OrderedImports.swift
@@ -17,6 +17,9 @@ import SwiftSyntax
 /// imports, and 4) @testable imports. These groups are separated by a single blank line. Blank lines in
 /// between the import declarations are removed.
 ///
+/// By default, imports within conditional compilation blocks (`#if`, `#elseif`, `#else`) are not ordered.
+/// This behavior can be controlled via the `orderedImports.includeConditionalImports` configuration option.
+///
 /// Lint: If an import appears anywhere other than the beginning of the file it resides in,
 ///       not lexicographically ordered, or  not in the appropriate import group, a lint error is
 ///       raised.
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -411,8 +411,8 @@ The order of the import groups is 1) regular imports, 2) declaration imports, 3)
 imports, and 4) @testable imports. These groups are separated by a single blank line. Blank lines in
 between the import declarations are removed.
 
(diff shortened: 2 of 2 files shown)


a7medev:
Some questions to consider:

1. Should this be enabled by default with "OrderedImports" or should it be a new rule/option?
2. Is a single test enough given that the specifics of ordering is shared in `orderImports` or do we need to add more tests for other cases?

Sources/SwiftFormat/Rules/OrderedImports.swift
```diff
@@ -47,6 +56,10 @@ public final class OrderedImports: SyntaxFormatRule {
       // Perform linting on the grouping of the imports.
       checkGrouping(linesSection)
 
+      if let firstLine = fileHeader.first, firstLine.type == .blankLine {
+        fileHeader.removeFirst()
+      }
```
a7medev: This handles the case in `testPreservesHeaderCommentInConditionalCompilationBlock`.

Previously, the newline separating the comment and `#if FOO` was printed as is as part of `fileHeader` resulting in an extra newline separating the comment and `#if FOO`.

a7medev: @allevato Actually looking at this again, the pretty-printer will always remove that line anyway (and looks like that isn't configurable), so there should be no breaking changes here.

a7medev:
@allevato Tests and real-world project compatibility tests are now green.

allevato (approved):
Thanks for adding this!

PR #1078, opened 2025-10-22
Add Wasm builds to `pull_request.yml`

MaxDesiatov:
This ensures that builds of `swift-format` for WASI don't regress.

### DIFF
--- a/Sources/_SwiftFormatInstructionCounter/include/InstructionsExecuted.h
+++ b/Sources/_SwiftFormatInstructionCounter/include/InstructionsExecuted.h
@@ -14,4 +14,4 @@
 
 /// On macOS returns the number of instructions the process has executed since
 /// it was launched, on all other platforms returns 0.
-uint64_t getInstructionsExecuted();
+uint64_t getInstructionsExecuted(void);


MaxDesiatov:
Yes, removing the Dispatch dependency would be greatly appreciated! 🙏

allevato:
> I'm referring to the SwiftFormat library product, not the executable. I'm not familiar with its API, but isn't there something there would allow in-place in-memory formatting that someone could rely on, say in a browser editor or something like that?

`SwiftFormat` the library API doesn't write files at all, directly. The various APIs can take either a `URL` to read from or a `String` with the source as inputs, and they write the output by calling `TextOutputStream.write`, so the behavior is based on whatever the caller's abstraction does.

Only the `swift-format` executable product needs atomic file writes.

MaxDesiatov:
> swift-markdown
> * [feat: add partial support for WASI swift-cmark#75](https://github.com/swiftlang/swift-cmark/pull/75) has been merged, but not yet released.

It's cherry-picked now to `release/6.2` and is included in [the 0.7.1 release](https://github.com/swiftlang/swift-cmark/releases/tag/0.7.1), so this item can now be crossed off the list.

kkebo:
> `DirectoryEnumerator`

Now, `DirectoryEnumerator` works on WASI.

- https://github.com/swiftlang/swift-corelibs-foundation/issues/5420

Toolchains for which this change is available:

- `main`: The next snapshot following main-snapshot-2026-02-24 and later
- `release/6.3`: 6.3-snapshot-2026-02-26 and later

PR #1088, opened 2025-11-12
Remove added group around base type in `AttributedTypeSyntax`

bnbarham:
Causes quite a few changes in eg. sourcekit-lsp, so presumably will in other codebases as well. Ideally we'd add this behind some new versioning in the config.

As an example of a change this causes:
```
private let logMessageToIndexLog:
  @Sendable (
    _ message: String, _ type: WindowMessageType, _ structure: LanguageServerProtocol.StructuredLogKind
  ) -> Void
```

To:
```
private let logMessageToIndexLog:
  @Sendable
    (
      _ message: String, _ type: WindowMessageType, _ structure: LanguageServerProtocol.StructuredLogKind
    ) -> Void
```

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -2411,9 +2411,6 @@ private final class TokenStreamCreator: SyntaxVisitor {
       )
     }
 
-    before(node.baseType.firstToken(viewMode: .sourceAccurate), tokens: .open)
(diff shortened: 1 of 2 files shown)


allevato:
> We also have a group around `inputs`, so I suspect the above case of:
> 
> ```
> attributes (
>   inputs
> ) effects -> outputs
> ```
> 
> Will end up much more common than:
> 
> ```
> attributes (inputs)
> effects -> outputs
> ```

We might be able to have it both ways: conditionally add the group around the whole `baseType` only if inputs is an empty tuple?

I guess that doesn't prevent `attributes (Int) \n effects -> outputs` though 🤔

PR #1092, opened 2025-11-14
print lint location and message in plain bold instead of white

### LINKED ISSUE #1087: Lint output uses illegible text color in light-mode terminals
It looks like swift-format prints its body text in bold-white when printing lint diagnostics, which is very difficult to read in my light-mode terminal theme:

Image

Should this text be plain (uncolored) instead so that it can inherit the default text color from the active theme?

QuietMisdreavus:
fixes #1087

When colorized lint diagnostics were introduced, the source-location and message were printed in ANSI bold white, which caused an issue when using swift-format in a light-mode terminal theme. This PR changes the behavior to instead print these components in bold plain-color text, allowing them to inherit their color from the active terminal theme.

Using the same example i screenshotted in the above-linked issue, this is what the output looks like now:

image

### DIFF
--- a/Sources/swift-format/Utilities/StderrDiagnosticPrinter.swift
+++ b/Sources/swift-format/Utilities/StderrDiagnosticPrinter.swift
@@ -65,12 +65,12 @@ final class StderrDiagnosticPrinter {
 
       switch diagnostic.severity {
       case .error: stderr.write("\(ansiSGR(.boldRed))error: ")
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
@allevato If the current version is fine with you (i updated the screenshot in the PR description) then can you go ahead and merge it? I don't have access to the button here. 😅

PR #1093, opened 2025-11-14
Format source code and run `git diff` instead of listing first

ahoppen:
The diff yields more insightful results than `swift-format lint`. Example of how the failures look like now: https://github.com/swiftlang/swift-format/actions/runs/19634366801/job/56221408873

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -37,15 +37,29 @@ jobs:
         uses: actions/checkout@v4
         with:
           persist-credentials: false
+      - name: Mark the workspace as safe
+        # https://github.com/actions/checkout/issues/766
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -44,21 +44,34 @@ jobs:
           swift /tmp/cross-pr-checkout.swift "${{ github.repository }}" "${{ github.event.number }}"
       - name: Build
         run: swift build --configuration release
+      - name: Write check script
```
bnbarham: Would it be better to re-use `github-workflows/.github/workflows/scripts/check-swift-format.sh` here and just cd to `$GITHUB_WORKSPACE/<repo>` before hand?

bnbarham (approved):

PR #1106, opened 2025-12-01
Update automerge.yml when new release branches are created

bnbarham:


.github/workflows/auto_update_version.yml
```diff
@@ -67,7 +84,7 @@ jobs:
           git push origin $UPDATE_BRANCH
```
TTOzzi: I explicitly added only `PrintVersion.swift`, so we should change this 🫣

TTOzzi (commented):
Ah… I just tested the changes reflecting the comments I left earlier. 
It turns out that modifying workflow files requires the `workflows:write` permission, and the default GitHub Actions token can’t be granted that permission. 
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions 



It seems this won’t work unless we use a personal access token with workflow permissions. 
I didn’t realize this wasn’t possible — sorry about that 😭

Issue #1152, opened 2026-02-03
`swift lint` should invoke `swift format lint` to make it more discoverable.


bkhouri:
This was reported by @mac9sb https://github.com/swiftlang/swift-package-manager/issues/9654, but I was unable to move the issue to the swift-format repository.

------

### Description

Seems to be happening in the latest version, been using the embedded lint tool for a while now but this is happening recently `6.2.3`. Format still works great.

### Reproduction

```sh
mac@Macs-MacBook-Air Developer % swift lint
error: unknown or missing subcommand 'swift-lint'
```

### Expected behavior

Swift lint should print the help text.

### Environment

swiftc -version

Apple Swift version 6.2.3 (swift-6.2.3-RELEASE)
Target: arm64-apple-macosx26.0
Build config: +assertions

------

The original poster then commented:

> Ignore this I realise now it's been nested under the `swift format` command as `swift format lint` perhaps a note when running `swift lint` that points users in the right direction would be a good idea?

PR #1153, opened 2026-02-05
feat: Introduce swift-lint executable to perform linting actions

rickhohler:
This PR introduces a dedicated `swift-lint` executable target, effectively implementing the request in issue #1152.

It refactors the CLI logic into a shared `_SwiftFormatCLI` library to allow both `swift-format` and `swift-lint` to share common argument parsing and infrastructure.

Resolves: #1152

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,6 +19,10 @@ var products: [Product] = [
     name: "swift-format",
     targets: ["swift-format"]
   ),
+  .executable(
+    name: "swift-lint",
+    targets: ["swift-lint"]
+  ),
   .library(
     name: "SwiftFormat",
     targets: ["SwiftFormat"]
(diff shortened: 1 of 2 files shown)


allevato:
Thanks! Can you back out the refactoring of the CLI into a separate module? I don't think we need that anymore, right?

allevato:
> What about making `swift-lint` a symlink to `swift-format` and make `swift-format` handle `CommandLine.arguments[0] == "swift-lint"` invocation?

That's precisely what I suggested already in https://github.com/swiftlang/swift-format/pull/1153#issuecomment-3854935211 and what the latest version of the PR does, but it's not ready yet because of the issues I mentioned in https://github.com/swiftlang/swift-format/pull/1153#issuecomment-3891307491.

PR #1156, opened 2026-02-16
Add support for .swift-format-ignore files

### LINKED ISSUE #870: Ability To Turn Off Formatting For Subdirectory

bkhouri:
Add supports for file-based ignoring using `.swift-format-ignore` files. These files allow a developer to specify patterns for files and directories that should be completely excluded from formatting and linting operations.

Fixes: #870
Issue: rdar://139269456

### DIFF
--- a/Documentation/IgnoringSource.md
+++ b/Documentation/IgnoringSource.md
@@ -8,8 +8,8 @@ representation is ignored by the formatter.
 
 ## Ignore A File
 
-In the event that an entire file cannot be formatted, add a comment that contains 
-`swift-format-ignore-file` at the top of the file and the formatter will leave 
+In the event that an entire file cannot be formatted, add a comment that contains
+`swift-format-ignore-file` at the top of the file and the formatter will leave
 the file completely unchanged.
 
 ```swift
@@ -90,7 +90,7 @@ struct Foo {
 }
 ```
 In this case, only the DoNotUseSemicolons and FullyIndirectEnum rules are disabled
-throughout the file, while all other formatting rules (such as line breaking and 
+throughout the file, while all other formatting rules (such as line breaking and
 indentation) remain active.
 
 ## Understanding Nodes
@@ -113,3 +113,93 @@ top level nodes that support suppressing formatting are:
   - Any member declaration inside of a declaration (e.g. properties and
     functions declared inside of a struct/class/enum). All code nested
     syntactically inside of the ignored node is also ignored by the formatter.
+
+## File-Based Ignoring with `.swift-format-ignore`
+
+In addition to the comment-based ignoring described above, `swift-format` supports
+file-based ignoring using `.swift-format-ignore` files. These files allow you to
+specify patterns for files and directories that should be completely excluded from
+formatting and linting operations.
+
+### `.swift-format-ignore` File Format
+
+`.swift-format-ignore` files use the same pattern syntax as `.gitignore` files:
+
+```
+# This is a comment
+*.generated.swift
+build/
+**/Generated/
+!important.swift
+src/**/test*.swift
+```
+
+### Pattern Syntax
+
+The `.swift-format-ignore` files supports the following pattern syntax:
+
+- Simple patterns: `file.swift` matches any file named `file.swift` anywhere in the tree
+- Wildcard patterns: `*.swift` matches all files ending with `.swift`
+- Directory patterns: `build/` matches the `build` directory and all files within it
+- Nested wildcards: `**/*.generated` matches files with `.generated` extension at any depth
+- Negation patterns: `!important.swift` excludes `important.swift` from being ignored (even if matched by other patterns)
+- Absolute patterns: `/root.swift` matches `root.swift` only at the project root
+- Complex patterns: `src/**/test*.swift` matches files starting with `test` and ending with `.swift` anywhere under `src/`
+
+### File Discovery and Precedence
+
+`swift-format` searches for `.swift-format-ignore` files by walking up the directory
+tree from each source file being processed. Multiple ignore files can exist in a
+project, with the following precedence rules:
+
+1. Closest wins: Ignore files closer to the source file take precedence
+2. Directory traversal: The tool searches from the file's directory up to the discovery of a `.swift-format` file
+3. Pattern evaluation: Within each ignore file, patterns are evaluated in order
+4. Negation support: Later negation patterns (`!pattern`) can override earlier ignore patterns
+
+### Example Project Structure
+
+```
+project/
+├── .swift-format-ignore          # Root ignore file
+├── src/
+│   ├── .swift-format-ignore      # Overrides root for src/ directory
+│   ├── main.swift
+│   └── generated/
+│       └── Generated.swift
+├── tests/
+│   └── test.swift
+└── build/
+    └── output.swift
+```
+
+**Root `.swift-format-ignore`:**
+```
+# Ignore all generated files
+*.generated.swift
+build/
+```
+
+**`src/.swift-format-ignore`:**
+```
+# Additional rules for src directory
+generated/
+!important.generated.swift
+```
+
+### Integration with swift-format Commands
+
+The `.swift-format-ignore` functionality works automatically with all `swift-format` commands:
+
+```bash
+# Format all files except those matching ignore patterns
+swift-format --recursive src/
+
+# Lint all files except those matching ignore patterns
+swift-format lint --recursive src/
+
+# Both commands will automatically discover and respect .swift-format-ignore files
+```
+
+Files matched by `.swift-format-ignore` patterns are completely excluded from processing -
+they will not be formatted, linted, or appear in any output.
--- a/Sources/SwiftFormat/Utilities/FileIterator.swift
+++ b/Sources/SwiftFormat/Utilities/FileIterator.swift
@@ -47,19 +47,42 @@ public struct FileIterator: Sequence, IteratorProtocol {
   /// The file extension to check for when recursing through directories.
   private let fileSuffix = ".swift"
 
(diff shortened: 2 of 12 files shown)


Sources/SwiftFormat/Utilities/IgnoreManager.swift
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Get relative path from base directory using standardized paths
+    let standardizedFile = file.standardizedFileURL
+    let standardizedBase = baseDirectory.standardizedFileURL
+    let relativePath = standardizedFile.path.replacingOccurrences(of: standardizedBase.path + "/", with: "")
+
+    // Check if file is a directory
+    var isDirectory: ObjCBool = false
+    FileManager.default.fileExists(atPath: file.path, isDirectory: &isDirectory)
```
rintaro: The caller should know if the URL is directory or not. This function should receive `isDirectory` flag.

Sources/SwiftFormat/Utilities/FileIterator.swift
rintaro: We should skip the directory here if it's being ignored.

bkhouri: Will do.

Sources/swift-format/Frontend/Frontend.swift
```diff
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
-@_spi(Internal) import SwiftFormat
+@_spi(Internal) @_spi(Testing) import SwiftFormat
```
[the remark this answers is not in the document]
rintaro: Still there :)

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -102,6 +133,13 @@ public struct FileIterator: Sequence, IteratorProtocol {
       if let out = output, visited.contains(out.standardizedFileURL.path) {
         output = nil
       }
+
+      // Apply ignore filtering if enabled
+      if let out = output, let ignoreManager = self.ignoreManager {
+        if ignoreManager.shouldIgnore(file: out, isDirectory: isDirectory) {
```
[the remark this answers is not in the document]
bkhouri: ack.

Tests/SwiftFormatTests/Utilities/GitIgnorePatternTests.swift
```diff
@@ -0,0 +1,343 @@
...
+
+@Suite
+struct GitIgnorePatternTests {
+
+  // MARK: - Basic Pattern Matching Tests
+
+  @Test func exactMatch() throws {
+    let pattern = try GitIgnorePattern("test.swift")
+
+    #expect(pattern.matches("test.swift", isDirectory: false))
```
allevato: Nit: throughout, when using `#expect` with Boolean values, I've found that explicitly writing `... == false` or `... == true` produces much more readable log messages if the expectation fails, vs. just testing the Boolean or `!` directly.

bkhouri:
@rintaro : I am unable to reply directly to [one of your comments](https://github.com/swiftlang/swift-format/pull/1156#discussion_r2814467523)

> Do we really need the lock in the first place? As far as I can see FileIterator is only used in the main thread.

I removed the locking mechanism, and also marked the `IgnoreManagerTests` to run serially to avoid any collision between tests since `IgnoreManager` is not thread safe.

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -47,19 +47,42 @@ public struct FileIterator: Sequence, IteratorProtocol {
...
-  ///   - workingDirectory: `URL` that indicates the current working directory. Used for testing.
-  public init(urls: [URL], followSymlinks: Bool, workingDirectory: URL = URL(fileURLWithPath: ".")) {
+  ///   - followSymlinks: `Bool` to indicate if symbolic links sthe current working directory. Used for testing.
+  ///   - ignoreManager: Optional `IgnoreManager`hould be followed when iterating.
+  ///   - workingDirectory: `URL` that indicates  to filter files based on .swift-format-ignore files.
+  public init(
+    urls: [URL],
+    followSymlinks: Bool,
+    workingDirectory: URL = URL(fileURLWithPath: "."),
+    ignoreManager: IgnoreManager
```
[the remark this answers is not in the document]
bkhouri: I kept this here in the event we want to have dependency injection testing.  I will change the access modifier to `package`

Sources/SwiftFormat/Utilities/FileIterator.swift
```diff
@@ -145,7 +178,15 @@ public struct FileIterator: Sequence, IteratorProtocol {
...
           relativePath = path
         }
-        output = URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: workingDirectory)
+        let fileURL = URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: workingDirectory)
+
+        // Apply ignore filtering
+        if self.ignoreManager.shouldIgnore(file: item, isDirectory: false) {
+          output = nil
+          continue  // Skip this file and continue to next
+        }
```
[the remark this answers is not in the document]
rintaro: You removed the another check for files in `next()`.
That means `swift-format path/to/File.swift` would perform formatting the specified file even if that file is `.swift-format-ignore`. Is that the desired behavior?

If so, what about the directories? E.g. `swift-format --recursive path/to/Sources/MyLibrary/generated/` Should this directory be formatted even if it's ignored in `.swift-format-ignore` in the parent directories?

Sources/SwiftFormat/Utilities/IgnoreManager.swift
```diff
@@ -0,0 +1,178 @@
...
+  ///   - file: The file URL to check
+  ///   - isFileDirectory: Whether the file is a directory
+  /// - Returns: `true` if the file should be ignored, `false` otherwise
+  public func shouldIgnore(file: URL, isDirectory: Bool) -> Bool {
+    // Determine base directory using the same logic as .swift-format configuration files
+    let baseDirectory: URL = {
+      if let configURL = Configuration.url(forConfigurationFileApplyingTo: file) {
+        return configURL.deletingLastPathComponent()
+      } else {
+        // No config file found, find ignore files and use the directory of the furthest one
```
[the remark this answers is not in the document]
bkhouri: the parent function was no longer used and has been removed.

Sources/SwiftFormat/Utilities/IgnoreManager.swift
```diff
@@ -0,0 +1,194 @@
...
+  /// Errors that can occur during ignore file operations
+  public enum IgnoreError: Error, Equatable {
+    case fileNotFound
+    case invalidPattern(String)
+  }
+
+  /// Shared cache for loaded ignore files to avoid repeated I/O across all instances
+  private var ignoreFileCache: [URL: [GitIgnorePattern]] = [:]
+
+  private var baseDirectoryCache: [URL: URL] = [:]
```
rintaro: Unused

rintaro (commented):
I haven't had time to thoroughly review the GitIgnorePattern logic, but overall mechanism looks fine to me.
https://github.com/swiftlang/swift-format/issues/1192 is really a concern, but I haven't found obvious reasons that cause it.

I'd like to hear from @allevato as well, but I think I'm fine with this PR. 
(after removing the unused `baseDirectoryCache`)

Documentation/IgnoringSource.md
```diff
@@ -113,3 +113,93 @@ top level nodes that support suppressing formatting are:
...
+- Complex patterns: `src/**/test*.swift` matches files starting with `test` and ending with `.swift` anywhere under `src/`
+
+### File Discovery and Precedence
+
+`swift-format` searches for `.swift-format-ignore` files by walking up the directory
+tree from each source file being processed. Multiple ignore files can exist in a
+project, with the following precedence rules:
+
+1. Closest wins: Ignore files closer to the source file take precedence
+2. Directory traversal: The tool searches from the file's directory up to the project root
```
allevato: "project root" isn't really a well-defined thing. Someone might interpret that to mean that we don't look beyond a Package.swift file or some other vague notion of a project.

I think what we want to capture here is that we stop looking if we see a `.swift-format` file.

allevato (approved):
LGTM modulo one documentation comment. Thanks for all the work on this!

Documentation/IgnoringSource.md
bkhouri: I updated the wording.  could you please double-check and let me know if you are ok with the update?

Sources/SwiftFormat/Utilities/IgnoreManager.swift
bkhouri: Removed

bkhouri:
> I haven't had time to thoroughly review the GitIgnorePattern logic, but overall mechanism looks fine to me. #1192 is really a concern, but I haven't found obvious reasons that cause it.

I agree that #1192 is concerning.  I'm not really sure what is occurring there.

rintaro (approved):

PR #1184, opened 2026-04-10
Improve error messages on invalid configuration files

Wilfred:
Given a configuration file like this:

    {
      "orderedImports": {
        "shouldGroupImports": false
      }
    }

swift-format would previously just say this, which sounds like it couldn't find the file:

    <unknown>: error: Unable to read configuration: The data couldn’t be read because it is missing.

after this change, it shows a more detailed error that describes what's wrong with the JSON:

    <unknown>: error: Unable to read configuration: missing key `orderedImports.includeConditionalImports`


allevato (approved):
This looks helpful, thanks!

PR #1197, opened 2026-05-01
Revert "Revert "Add support for .swift-format-ignore files""

bkhouri:
Reverts swiftlang/swift-format#1195

This reverts commit d9812329999062d4ddecd7e11eec0a2fc48f4943 but changes all `package` modifiers to `public`, and put them behind an Internal SPI, as the latter was causing compilation failure when the project was build using cmake.

### DIFF
--- a/Documentation/IgnoringSource.md
+++ b/Documentation/IgnoringSource.md
@@ -8,8 +8,8 @@ representation is ignored by the formatter.
 
 ## Ignore A File
 
-In the event that an entire file cannot be formatted, add a comment that contains 
-`swift-format-ignore-file` at the top of the file and the formatter will leave 
+In the event that an entire file cannot be formatted, add a comment that contains
+`swift-format-ignore-file` at the top of the file and the formatter will leave
 the file completely unchanged.
 
 ```swift
@@ -90,7 +90,7 @@ struct Foo {
 }
 ```
 In this case, only the DoNotUseSemicolons and FullyIndirectEnum rules are disabled
-throughout the file, while all other formatting rules (such as line breaking and 
+throughout the file, while all other formatting rules (such as line breaking and
 indentation) remain active.
 
 ## Understanding Nodes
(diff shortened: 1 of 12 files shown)


allevato:
Can we just make CMake pass the correct `-package-name` when building the targets? That would be better than leaking things as public API.

@compnerd Does CMake's Swift support have some kind of built-in support for `-package-name` or do we just need to provide it as a flag for each of the targets?

bkhouri:
> Can we just make CMake pass the correct `-package-name` when building the targets? That would be better than leaking things as public API.
> 
> @compnerd Does CMake's Swift support have some kind of built-in support for `-package-name` or do we just need to provide it as a flag for each of the targets?

I was talking with @rintaro offline, who suggested
> `-package-name` support in CMake would probably need some bike shedding, like the default package name, how to customize it in CMake configuration, etc. So I'd recommend to use `public` for the ignore file support PR

allevato:
Making something `public` is a big API decision since we would be on the hook to support it indefinitely if someone started to use it. That's not a decision we should make lightly.

If _general_ `-package-name` support in CMake is too much to take on right now, can we just pass the flag directly in our own swift-format targets that need it so that we can continue using `package`? WDYT @rintaro ?

rintaro:
I'm totally fine for proper package support if we are okay for holding this PR until that.

rintaro:
I just recommended public because I didn't think this much:

> we would be on the hook to support it indefinitely if someone started to use it. 

And we already introduced some public symbols in the original PR.

allevato:
Many of those `public` symbols are `@_spi(Internal)` at least (themselves or as members of an `@_spi(Internal)` type).

So let's just drop the `package` discussion for now and make them either `@_spi(Internal)` or `@_spi(Testing)` based on what we need them for. Can you see if that works @bkhouri ?

bkhouri:
I'm more than happy to get the proper support for `-package-name` into the project, and then I can revert back to using `public`.  Does that work for all of you?

bkhouri:
The API's have been marked `public` and put behind a SPI.

compnerd:
> @compnerd Does CMake's Swift support have some kind of built-in support for `-package-name` or do we just need to provide it as a flag for each of the targets?

Should be available properly in CMake 4.4 to be released in June.

allevato (approved):

rintaro:
(waiting for cross repo testing)
https://ci.swift.org/job/swift-PR-Linux-smoke-test/27325/
https://ci.swift.org/job/swift-PR-macos-smoke-test/28012/
https://ci-external.swift.org/job/swift-PR-windows/54363/

bkhouri:
> (waiting for cross repo testing) https://ci.swift.org/job/swift-PR-Linux-smoke-test/27325/ https://ci.swift.org/job/swift-PR-macos-smoke-test/28012/ https://ci-external.swift.org/job/swift-PR-windows/54363/

Thanks for posting this @rintaro :). Hopefully the cross-repo testing will not find any issues :)

rintaro (approved):

bkhouri:
The cross repository builds all passed. @rintaro , @allevato : can one of you please merge this at your earliest convenience as I don't have write permission on this repository. 

Thank you

PR #1203, opened 2026-05-13
Fix double space between `catch`/`where` in `where`-only `catch` clause

### LINKED ISSUE #1076: Two spaces inserted between `catch` and `where` if there is no clause to capture the error
```swift
do {

} catch where true {

}
```

Gets formatted to

```swift
do {

} catch  where true {

}
```

Notice the two spaces between `catch` and `where`.

broken-circle:
Closes #1076.

When formatting a `catch` clause whose item consists solely of a `where` clause with no preceding pattern, `TokenStreamCreator.visit(_ node: CatchClauseSyntax)` emits a `.space` before the `catch` item's first token. With no pattern present, the first token is the `where` keyword. `TokenStreamCreator.visit(_ node: WhereClauseSyntax)` also inserts a preceding break before the same `where` keyword, producing duplicate whitespace.

This PR suppresses the `catch`-clause-level whitespace when a `catch` item has no pattern and allows `WhereClauseSyntax` to remain the sole source.

## Testing

Added `testCatchWhereOnly_noBreakBeforeCatch()` and `testCatchWhereOnly_breakBeforeCatch()` to `DoStmtTests`.

All tests pass when run locally on macOS.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -742,10 +742,19 @@ private final class TokenStreamCreator: SyntaxVisitor {
     // old (pre-SE-0276) behavior (a fixed space after the `catch` keyword).
     if node.catchItems.count > 1 {
       for catchItem in node.catchItems {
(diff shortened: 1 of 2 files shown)


allevato (approved):
Thank you!

PR #1204, opened 2026-05-14
Avoid stranding keyword introducers on own line before their operands

broken-circle:
When formatting code such as:

```swift
throw SomeError
    .memberA
    .memberB
```

at a narrow line length, the formatter strands the keyword on its own line:

```swift
throw 
    SomeError
    .memberA
    .memberB
```

The same happens for `return` before a wrapped chain expression and for `repeat` before its `each` operand in a pack expansion. These shapes are unusual and rarely the user's intent.

This PR addresses three sites where a keyword introducer was being stranded ahead of its operand: `ReturnStmtSyntax`, `ThrowStmtSyntax`, and `PackExpansionExprSyntax`/`PackExpansionTypeSyntax`. The exact gating differs by site:

- For `return` and `throw`, the break ignores discretionary newlines only when the expression begins with a member-access chain. In all other cases the break remains a plain elective, so user-entered newlines before bare identifiers, literals, or other non-chain expressions continue to be respected. A discretionary newline immediately followed by a line comment or doc-line comment is also preserved on chain expressions, since the comment is a strong signal of intent that the formatter should not discard.

- For `repeat` in pack expansion expressions and types, the change is unconditional. The operand is constrained to be `each <reference>`, so preserving a user's newline there is never the right behavior.

These changes bring the affected sites in line with the existing treatment of `try`, `await`, and similar keyword-modified expressions.

### `PrettyPrinter` change

The behavior described above requires a small change in the pretty printer's scan pass. An `.elective(ignoresDiscretionary: true)` break is documented as a "last resort" break, and it fires only when its content cannot fit on the current line. In practice, the scan pass computed its length against the same running counter as ordinary breaks, which is inflated by `maxLineLength` per forced break (`.soft` or `.hard`) within its span. This meant any forced break later in the span, such as one from a wrapped function call inside a chain, would force the elective to fire even when its own content fit, contradicting its stated semantics.

`prettyPrint()` now maintains a second counter, `suffixTotal`, that counts only the literal content emitted by tokens. Forced breaks advance it by `size`, not `maxLineLength`. The two counters diverge only inside the span of a forced break. `.elective(ignoresDiscretionary: true)` breaks close their length against `suffixTotal`. All other breaks and `.open` groups continue to close against `total`. The span over which a break is measured is unchanged.

This change applies to every use of `.elective(ignoresDiscretionary: true)` in the formatter, not only the new ones for `return` and `throw`. The old behavior (being preempted by forced breaks elsewhere in the span) was inconsistent with the "last resort" framing documented on `NewlineBehavior.elective`.

### Testing

Added `ReturnStmtTests`, `ThrowStmtTests`, and `PackExpansionTests` to verify the new visitor logic.

All tests pass when run locally on macOS. No existing tests needed modifications.

### Future directions

Labeled `break` and `continue` have a similar symptom but a different cause. The Swift parser requires the label to appear on the same source line as the keyword, and any newline between them causes the parser to commit to the labelless form and treat the identifier on the next line as a fresh statement. By the time the formatter sees the tree, the construct has already been split into two unrelated `CodeBlockItemSyntax` nodes, and the user's intent is no longer recoverable at the token-stream layer. Fixing this would probably require either a post-parse `SyntaxRewriter` or parser-level error recovery in swift-syntax. I can file a separate issue if there's interest.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -587,7 +587,37 @@ public class PrettyPrinter {
     // Keep track of the indices of the .open and .break token locations.
     var delimIndexStack = [Int]()
     // Keep a running total of the token lengths.
+    //
+    // `total` includes the `maxLineLength` injection contributed by forced breaks (soft and
+    // hard). It is used to compute the span length for ordinary breaks and `.open` groups,
+    // so that an enclosing context sees a length large enough to force wrapping when its
+    // span contains a forced break.
+    //
+    // `suffixTotal` excludes those injections, counting only the literal content emitted by
+    // the token stream. Last-resort elective breaks (`.elective(ignoresDiscretionary: true)`)
+    // compute their span length against `suffixTotal` instead of `total`, so that forced
+    // breaks occurring later in the same span do not force them to fire. The span itself
+    // (the range from the break to the next break or enclosing `.close`) is identical to
+    // that of any other break. Only the counter differs.
     var total = 0
+    var suffixTotal = 0
+
+    /// Returns `true` if the break at `tokenIndex` is a last-resort elective
+    /// (`.elective(ignoresDiscretionary: true)`).
+    func isLastResortElectiveBreak(at tokenIndex: Int) -> Bool {
+      if case .break(_, _, .elective(ignoresDiscretionary: true)) = tokens[tokenIndex] {
+        return true
+      }
+      return false
+    }
+
+    /// Returns the counter value to use when closing the length of the break or group at
+    /// `tokenIndex`. Last-resort elective breaks use `suffixTotal` so that `maxLineLength`
+    /// injections from forced breaks within their scope do not contribute to their length.
+    /// Everything else uses `total`.
+    func closingCounter(for tokenIndex: Int) -> Int {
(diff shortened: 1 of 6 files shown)


allevato:
I've cited it before, but the current behavior is an example of [the Rectangle Rule](https://github.com/google/google-java-format/wiki/The-Rectangle-Rule), which states that

> When a source file is formatted, each subtree gets its own bounding rectangle, containing all of that subtree’s text and none of any other subtree’s.

We have a few exceptions for that rule (such as assignment of a function call to a left-hand-side—we allow the function name and open parenthesis to remain on the same line as the `=`), but for the most part, the rule is meant to foster uniformity and consistency.

But regardless, this isn't a simple as an obvious bug fix (like the `catch  where` issue) because various projects that use CI to enforce formatting might behave differently if we have rules that start ignoring certain discretionary newlines. We don't really want to make a bunch of individual changes to formatting that will cause churn for clients. Another option would be to introduce a configuration setting to control it, but I don't know if the juice is worth the squeeze for this particular case, and I've also said before that I want the formatter to be _more_ opinionated, not more configurable.

So I don't think there's a very viable way forward for this at this time.

broken-circle:
@allevato Thanks for the quick review and the link to the Rectangle Rule. That makes sense to me. Agreed on keeping the formatting opinionated since too much configuration eventually just adds noise.

PR #1210, opened 2026-05-24
Fix assertion failure when swift-format-ignore is used inside #if

### LINKED ISSUE #1079: swift-format-ignore in conditionally-compiled code results in an assertion failure in PrettyPrint.swift
When running a debug build of swift-format on a file with the following content:

```swift
#if FOO
// swift-format-ignore
print("Hi")
#endif
```

swift-format crashes with the following assertion failure:

```
SwiftFormat/PrettyPrint.swift:720: Assertion failed: Too many unresolved delimiter token lengths.
```

adityasingh2400:
Fixes #1079.

### Problem

Running swift-format (debug build) on a `// swift-format-ignore` directive applied to the last element of a conditional-compilation clause crashes with:

```
SwiftFormat/PrettyPrint.swift:725: Assertion failed: Too many unresolved delimiter token lengths.
```

Minimal reproducer from the issue:

```swift
#if FOO
// swift-format-ignore
print("Hi")
#endif
```

### Root cause

`visit(_ node: IfConfigClauseSyntax)` attaches the clause body’s closing tokens with `after(node.elements?.lastToken, tokens: ..., .close)`, matching the `.open` it emits right after the condition.

When the last element of the body is formatter-ignored, the whole element is emitted as a single `.verbatim` token by `appendFormatterIgnored` and the visit returns `.skipChildren`, so that element’s tokens are never visited. The `after` group registered on its last token is therefore never flushed, the matching `.close` is dropped, and the `.open` is left unclosed. During length calculation in `prettyPrint()`, `delimIndexStack` ends with two unresolved entries and the assertion fires (in release builds the lengths would be resolved incorrectly).

The same hazard is already handled elsewhere: the `CodeBlockItemListSyntax` and `MemberBlockItemListSyntax` visitors skip emitting per-item `.open`/`.close` for ignored items, and the `SwitchCase` visitor falls back to a `before(nextToken:)` anchor when its body has no trailing token of its own.

### Fix

Detect when the body’s last token belongs to a formatter-ignored item, and in that case attach the closing tokens `before` the following token (the next `#elseif`/`#else`/`#endif`), which is always visited. This mirrors the existing empty-clause branch. Non-ignored `#if` blocks take the unchanged `after(lastElemTok:)` path, so their output is byte-for-byte identical.

### Testing

- Added four cases to `IgnoreNodeTests`: ignore in a single `#if`, ignore in each of `#if`/`#elseif`/`#else`, ignore in a nested `#if`, and an ignored member inside `#if` within a type. Each asserts the ignored region is preserved verbatim while the surrounding directive lines are formatted; the helper’s idempotency check also runs.
- Verified the reproducer (and `#elseif`/`#else`, nested, member-block, postfix `#if`, and `indentConditionalCompilationBlocks = false` variants) no longer crashes and produces correct output.
- Full `swift test` suite is green with no regressions.

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -1523,10 +1523,21 @@ private final class TokenStreamCreator: SyntaxVisitor {
     // Unlike other code blocks, where we may want a single statement to be laid out on the same
     // line as a parent construct, the content of an `#if` block must always be on its own line;
     // the newline token inserted at the end enforces this.
-    if let lastElemTok = node.elements?.lastToken(viewMode: .sourceAccurate) {
+    //
+    // The closing tokens are normally attached after the last token of the body. If that last token
+    // belongs to a formatter-ignored item, however, it is never visited (its node is emitted as a
+    // single verbatim token), so an `after` group on it would be dropped and the `.open` above would
+    // be left unclosed. In that case, attach the closing tokens before the following token (the next
+    // `#elseif`/`#else`/`#endif`) instead, which is always visited.
+    if let lastElemTok = node.elements?.lastToken(viewMode: .sourceAccurate),
+      !isFormatterIgnored(lastElemTok)
+    {
       after(lastElemTok, tokens: .break(breakKindClose, newlines: .soft), .close)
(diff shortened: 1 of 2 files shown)


allevato (approved):
Thanks!

PR #1211, opened 2026-05-26
Move @_spi attributes to members in NoAccessLevelOnExtensionDeclaration

### LINKED ISSUE #714: `NoAccessLevelOnExtensionDeclaration` should move `@_spi` attribute with access level attribute
If an extension is marked with an access level attribute and has the `@_spi` attribute, the `NoAccessLevelOnExtensionDeclaration` rule should move both attributes to each declaration in the extension.

```swift
func testExtensionWithSpiAttribute() {
  assertFormatting(
    NoAccessLevelOnExtensionDeclaration.self,
    input: """
      @_spi(Something) public extension Foo {
        var bar: String { "" }
      }
      """,
    expected: """
      extension Foo {
        @_spi(Something) public var bar: String { "" }
      }
      """,
    findings: [
      // TODO: Add findings
    ]
  )
}
```

<hr>

Originated from: https://github.com/apple/swift-syntax/pull/2602#discussion_r1563091753

adityasingh2400:
Fixes #714.

### Problem

When `NoAccessLevelOnExtensionDeclaration` moves an explicit access level off an extension and onto its members, it leaves any `@_spi` attribute behind on the extension:

```swift
@_spi(Something) public extension Foo {
  var bar: String { "" }
}
```

currently becomes

```swift
@_spi(Something) extension Foo {
  public var bar: String { "" }
}
```

An `@_spi` attribute on an extension only takes effect when the extension also carries an explicit access level, and it distributes to the members the same way that access level does. Stripping the access level while leaving `@_spi` on the extension is inconsistent with the rule's job of pushing member-affecting modifiers down to each member.

### Fix

The rule now collects `@_spi` attributes from the extension and prepends them to each member alongside the access level, producing the output from the issue:

```swift
extension Foo {
  @_spi(Something) public var bar: String { "" }
}
```

Attributes that genuinely belong on the extension, such as `@objc` and `@available`, are left in place. When the access level is the redundant `internal` keyword the members are not touched, so a `@_spi` attribute has nothing to attach to and stays on the extension.

I confirmed with `swiftc -emit-private-module-interface-path` that the member-level SPI surface is identical before and after the transformation: both forms expose `@_spi(Something) public var bar`, so this only changes where the attribute is written, not the SPI it vends.

### Verification

Added tests in `NoAccessLevelOnExtensionDeclarationTests` covering the single-member case, multiple `@_spi` groups, `@objc` staying while `@_spi` moves, placement before existing member attributes, and the redundant-`internal` case. The four behavioral tests fail before this change and pass after. The full suite passes (`swift test`: 922 tests, 0 failures, 1 pre-existing skip), and the changed source passes `swift-format lint --strict` against the repo's own configuration.

### DIFF
--- a/Sources/SwiftFormat/Rules/NoAccessLevelOnExtensionDeclaration.swift
+++ b/Sources/SwiftFormat/Rules/NoAccessLevelOnExtensionDeclaration.swift
@@ -25,8 +25,9 @@ public final class NoAccessLevelOnExtensionDeclaration: SyntaxFormatRule {
     /// The rule is currently visiting top-level declarations.
     case topLevel
 
-    /// The rule is currently inside an extension that has the given access level keyword.
-    case insideExtension(accessKeyword: Keyword)
+    /// The rule is currently inside an extension that has the given access level keyword, along with
+    /// any `@_spi` attributes that should be moved down to each member alongside that keyword.
+    case insideExtension(accessKeyword: Keyword, spiAttributes: [AttributeListSyntax.Element])
   }
 
   /// Tracks the state of the rule to determine which action should be taken on visited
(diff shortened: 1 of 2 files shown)


allevato (approved):
Thanks!

PR #1215, opened 2026-06-06
Don't insert trailing commas into attribute argument lists

ayush-that:
fixes #1178.

alwaysUsed inserted a trailing comma into multiline @attached(...) arguments, which the parser rejects (see swiftlang/swift-syntax#3306). macro role and custom attributes are syntactically indistinguishable, so never insert into any attribute's argument list. an existing comma is still managed as configured.


allevato (approved):
Thanks! This seems like a straightforward solution for now vs. trying to hard-code which attributes support this and which don't.

There may be some divergence between what we do here and what tools like SwiftLint might do, to which my answer is still "you shouldn't use multiple linters to check the same rule".

PR #1216, opened 2026-06-14
Fix relative path matching for nested ignore files on Windows

### LINKED ISSUE #1192: Remove `withKnownIssue` in the  `fileIteratorWithMultipleNonEmptySwiftIgoreFiles` test
While working on #1156, a test was marked with `withKnownIssues` on a Windows platform as the FileIterator did not return the expected set of file when there are nested `.swift-format-ignore` files.

This has been opened to find the root cause, and address the issue.

somiljain2006:
Resolves #1192

Improve path handling for nested `.swift-format-ignore` evaluation on Windows.

This change updates `IgnoreManager.getRelativePath(from:to:)` to make path matching more robust across platforms:

* Perform case-insensitive path comparisons on Windows.
* Enforce directory-boundary checks when determining whether a file is contained within an ignore file's directory hierarchy, preventing false matches such as `Sources` matching `SourcesTests`.
* Normalize relative paths to use POSIX-style forward slashes before evaluating `GitIgnorePattern`s.

With these changes, `fileIteratorWithMultipleNonEmptySwiftIgoreFiles` passes without the `withKnownIssue` workaround, so the wrapper has been removed.

Windows CI will provide final validation that this addresses the reported failure.


somiljain2006:
@allevato Can you review this pr?

somiljain2006:
The current approach doesn't appear to address the Windows failure, so I'm going to investigate this from a different angle. I'll reproduce and debug the issue on a local Windows environment to better understand the path and ignore-rule behavior before proposing another fix.

PR #1218, opened 2026-06-15
Fix rule-based formatting for non-zero offset selections

### LINKED ISSUE #1114: Rule-based formatting is not applied when using offsets starting at 1 or greater
When `swift-format` is run with an offset range starting at 1 or greater, the formatter applies general formatting but does not apply any rule-based formatting.

### **Example**
`.swift-format`:
```json
{
  "rules": {
    "NoParensAroundConditions": true,
    "DoNotUseSemicolons": true
  }
}
```
input:
```swift
/// Some String....
 if (a = 0) {             NSLog("print: \(a)");}
```

Environment variable for convenience:
```bash
export CODE="/// Some String....\n if (a = 0) {             NSLog(\"print: \\(a)\");}"
```

And expected result:
```swift
/// Some String....
if a = 0 { NSLog("print: \(a)") }
```

---
These commands produce the expected result:
```bash
echo $CODE | swift-format format - --configuration .swift-format
echo $CODE | swift-format format - --configuration .swift-format --offsets 0:999
```

But the following commands do not:
```bash
echo $CODE | swift-format format - --configuration .swift-format --offsets 1:999
echo $CODE | swift-format format - --configuration .swift-format --offsets 0:1 --offsets 1:999
```

Actual output:
```swift
/// Some String....
if (a = 0) { NSLog("print: \(a)");  }
```
Although the sliced input beginning at offset 1 still forms valid Swift syntax (`// Some String....`), rule-based formatting are not applied, while other non-rule formatting is applied correctly.

somiljain2006:
Fixes #1114 

Rule-based formatting was skipped when a selection started inside a node's leading trivia. Use positionAfterSkippingLeadingTrivia when checking whether a rule should be applied. Adds a regression test covering formatting with a non-zero offset range.

### DIFF
--- a/Sources/SwiftFormat/Core/Rule.swift
+++ b/Sources/SwiftFormat/Core/Rule.swift
@@ -25,6 +25,9 @@ public protocol Rule {
   /// Whether this rule is opt-in, meaning it is disabled by default.
   static var isOptIn: Bool { get }
 
+  /// The scope of the syntax node that this rule operates on.
+  static var targetScope: RuleTargetScope { get }
+
   /// Creates a new Rule in a given context.
   init(context: Context)
 }
@@ -45,9 +48,31 @@ public enum FindingAnchor {
   case trailingTrivia(Trivia.Index)
 }
 
+/// Describes the portion of a syntax node that a rule targets.
+@_spi(Rules)
+public enum RuleTargetScope {
+  /// The rule operates on the non-trivia content of the node.
+  case content
+
+  /// The rule operates on the leading trivia of the node.
+  case leadingTrivia
+
+  /// The rule operates on the trailing trivia of the node (such as end-of-line comments).
+  case trailingTrivia
+
+  /// The rule operates on either leading or trailing trivia of the node.
+  case trivia
+
+  /// The rule operates on the entire node, including all trivia and content.
+  case all
+}
+
 extension Rule {
   /// By default, the `ruleName` is just the name of the implementing rule class.
   public static var ruleName: String { String("\(self)".split(separator: ".").last!) }
+
+  /// By default, rules target the non-trivia content of a syntax node.
+  public static var targetScope: RuleTargetScope { .content }
 
   /// Emits the given finding.
   ///
--- a/Sources/SwiftFormat/Core/Rule.swift
+++ b/Sources/SwiftFormat/Core/Rule.swift
@@ -26,8 +26,7 @@ public protocol Rule {
   static var isOptIn: Bool { get }
 
   /// The scope of the syntax node that this rule operates on.
-  static var targetScope: RuleTargetScope { get }
-
+  static var affectedContent: AffectedContent { get }
   /// Creates a new Rule in a given context.
   init(context: Context)
 }
@@ -50,7 +49,7 @@ public enum FindingAnchor {
 
 /// Describes the portion of a syntax node that a rule targets.
 @_spi(Rules)
-public enum RuleTargetScope {
+public enum AffectedContent {
   /// The rule operates on the non-trivia content of the node.
   case content
 
@@ -72,7 +71,7 @@ extension Rule {
   public static var ruleName: String { String("\(self)".split(separator: ".").last!) }
 
   /// By default, rules target the non-trivia content of a syntax node.
-  public static var targetScope: RuleTargetScope { .content }
+  public static var affectedContent: AffectedContent { .content }
 
   /// Emits the given finding.
   ///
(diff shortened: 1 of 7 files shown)


somiljain2006:
@allevato Can you review this pr?

allevato:
I'm not sure this is the right approach. What if a rule only operated on the doc comments of a node (and we do have a couple of those)? Since this change skips leading trivia, if the user only selected the doc comment lines of the node, it would be skipped.

Given the way trivia is represented in swift-syntax, I think we might need a way for a rule to indicate whether it operates on just the leading trivia of a node, or just the non-trivia content of the node, or both, and have the `shouldFormat` check take that into account when it checks for containment.

somiljain2006:
@allevato Thanks for pointing that out. I've pushed an update that adds a rule-level targetScope so rules can specify whether they operate on content, leading/trailing trivia, trivia, or the entire node, and Context.shouldFormat now checks selection containment against the appropriate region. I also updated trivia-based rules such as UseTripleSlashForDocumentationComments and NoBlockComments, and added tests for leading/trailing trivia selections.

Sources/SwiftFormat/Core/SyntaxFormatRule.swift
```diff
@@ -32,7 +37,12 @@ public class SyntaxFormatRule: SyntaxRewriter, Rule {
   public override func visitAny(_ node: Syntax) -> Syntax? {
     // If the rule is not enabled, then return the node unmodified; otherwise, returning nil tells
     // SwiftSyntax to continue with the standard dispatch.
-    guard context.shouldFormat(type(of: self), node: node) else { return node }
+    guard context.shouldFormat(type(of: self), node: node) else {
+      if type(of: self).targetScope != .content && context.selectionOverlaps(node) {
```
allevato: Why is this additional check necessary; what is it handling that `shouldFormat` doesn't?

Sources/SwiftFormat/Core/Rule.swift
```diff
@@ -25,6 +25,9 @@ public protocol Rule {
   /// Whether this rule is opt-in, meaning it is disabled by default.
   static var isOptIn: Bool { get }
 
+  /// The scope of the syntax node that this rule operates on.
+  static var targetScope: RuleTargetScope { get }
```
allevato: `targetScope` is kind of a vague name, and "target" and "scope" in a programming language environment both make me think of different things, like brace-delimited regions. Can we call it `affectedContent` and the type `AffectedContent`?

Tests/SwiftFormatTests/API/SwiftFormatterSelectionTests.swift
```diff
@@ -232,6 +233,99 @@ final class SwiftFormatterSelectionTests: XCTestCase {
     try assertFormatting(source, expected: expected, selection: Selection(lineRanges: [0...0]))
   }
 
+  func testRuleFormattingWithNonZeroOffsetRange() throws {
+    let source = "/// Some String....\n if (a = 0) {             NSLog(\"print: \\(a)\");}"
```
allevato: Please use multiline strings for these so they're easier to read.

Tests/SwiftFormatTests/API/SwiftFormatterSelectionTests.swift
```diff
@@ -232,6 +233,99 @@ final class SwiftFormatterSelectionTests: XCTestCase {
...
+      sourceFileSyntax: tree,
+      ruleNameCache: ruleNameCache
+    )
+
+    XCTAssertTrue(trailingContext.shouldFormat(NoBlockComments.self, node: funcNode))
+    XCTAssertFalse(trailingContext.shouldFormat(UseTripleSlashForDocumentationComments.self, node: funcNode))
+  }
+
+  func testContentSelectionIgnoresTrivia() throws {
+    let source = """
```
allevato, line 323: It's not clear to me what this is testing; the formatting wouldn't change the doc comment even if it was within the selection range.

What if the code being formatted was this:

```swift
/* documentation */
    func foo() {}
```

I would expect a line range selection of 2...2 to fix the indentation of the function but not touch the comment.

Sources/SwiftFormat/Core/SyntaxFormatRule.swift
somiljain2006: This was meant to avoid stopping traversal too early for trivia-based rules. If shouldFormat returned false, SwiftSyntax would skip descendants, which could prevent rules like NoBlockComments from reaching trivia in the selected region. The overlap check keeps traversal going in that case.

somiljain2006:
@allevato, can you review the updated changes?

PR #1225, opened 2026-06-25
Fix #endif gluing onto the following token

### LINKED ISSUE #1223: Decl following #endif gets joined on the same line when `respectsExistingLineBreaks` is `false`
When a declaration is preceded by a trailing `#if … #endif` block that wraps an attribute, and `respectsExistingLineBreaks` is `false`, the line break after `#endif` is collapsed and the declaration is glued onto the `#endif` line, producing source that no longer parses.

## Input

```swift
#if FLAG
@frozen
#endif
struct S {}
```

## Configuration

`respectsExistingLineBreaks: false` (otherwise defaults).

## Expected output

```swift
#if FLAG
  @frozen
#endif
struct S {}
```

## Actual output

```swift
#if FLAG
  @frozen
#endif struct S {}
```

`#endif struct S {}` is invalid Swift (`error: extra tokens following conditional compilation directive`). Re-running the formatter over its own output collapses it further to `#endifstruct S {}`.

## Notes

- Requires the `#if` to wrap an attribute belonging to the following declaration; a non-attribute statement inside the `#if` does not trigger it.
- With `respectsExistingLineBreaks: true` the output is correct.

dduan:
Add test for #1223. Fix it.

rdar://180652604

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -1496,7 +1496,15 @@ private final class TokenStreamCreator: SyntaxVisitor {
 
   override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
     // there has to be a break after an #endif
-    after(node.poundEndif, tokens: .break(.same, size: 0))
+    let newlines: NewlineBehavior
+    if node.parent?.is(AttributeListSyntax.self) == true {
+      newlines = .elective
+    } else if isNestedInPostfixIfConfig(node: Syntax(node)) {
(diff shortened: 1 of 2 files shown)


allevato (approved):
Thanks!

allevato:
Looks like CI is busted until https://github.com/swiftlang/github-workflows/pull/290 is merged.

bkhouri:
> Looks like CI is busted until [swiftlang/github-workflows#290](https://github.com/swiftlang/github-workflows/pull/290) is merged.

Sorry for introducing the regression.  https://github.com/swiftlang/github-workflows/pull/290 has been merged.  I would have restarted the failed workflow, but I don't have permissions to do so in this project.

PR #1233, opened 2026-06-30
Ignore backticks when sorting imports.

allevato:
If raw identifiers are used for module names, we ignore the backticks so that the sort order is based on the content of the name and not whether or not it's been escaped. This prevents all raw identifiers from being grouped together simply on the basis that they're raw identifiers.

### DIFF
--- a/Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift
+++ b/Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift
@@ -1202,4 +1202,93 @@ final class OrderedImportsTests: LintOrFormatRuleTestCase {
       configuration: configuration
     )
   }
+
+  func testRawIdentifierSorting() {
+    assertFormatting(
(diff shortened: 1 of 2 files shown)


Tests/SwiftFormatTests/Rules/OrderedImportsTests.swift
```diff
@@ -1202,4 +1202,53 @@ final class OrderedImportsTests: LintOrFormatRuleTestCase {
       configuration: configuration
     )
   }
+
+  func testRawIdentifierSorting() {
```
dylansturg, line 1206: Maybe add an example of an import with an attribute like `@testable` and a declaration import like

```
import class `Foundation`.Data
```

allevato: Done, PTAL.

dylansturg (approved):

PR #1236, opened 2026-07-01
Add a `SwiftTestingNamingConventions` rule.

allevato:
This adds a lint-only rule that enforces some opinionated style guidelines about Swift Testing usage. All options are off by default, so the rule is a no-op until a user enables any of them:

* `forbidSuiteWithoutParameters`: Since a type containing any `@Test`s is automatically a suite, diagnose if the user has written a "redundant" `@Suite` (one that doesn't specify any parameters, like traits).
* `forbidSuiteDescription` and `forbidTestDescription`: Diagnose if the user has provided separate string literal descriptions for a suite or test.
* `requireRawIdentifierTestNames`: Diagnoses if the name of a `@Test` function isn't a raw identifier (surrounded by backticks). This is meant to be combined with the previous option to enforce a single description for tests and that description is the function name.

I didn't create a similar `requireRawIdentifierSuiteNames`. We could add that later if we want (in my own planned usage, I don't plan to use that.)


dylansturg (approved):

tshortli:
This broke Windows PR testing for swiftlang: https://ci-external.swift.org/job/swift-PR-windows/57696/console. I think CMakeLists.txt just needs to be updated to remove `Core/WithAttributesSyntax+Convenience.swift`.

PR #1238, opened 2026-07-01
Add `git-swift-format` to the `Scripts` directory.

allevato:
*Consider this experimental for now.*

This is a shamelessly AI-guided adaption of `git-clang-format`, but for swift-format. Logic has been simplified a slight bit:

* There is no way to specify file extensions; it assumes `.swift`.
* Instead of a "style", you can pass a configuration that corresponds to the `--configuration` flag in swift-format. If omitted, swift-format's standard configuration resolution is used (finding the `.swift-format` file relative to the sources being formatted).

Other than that, it's basically the same script as `git-clang-format`. To use it, copy it somewhere into your `PATH` and then run `git swift-format` in a directory that has staged changes (or use one of the flags to specify what should be formatted).

### DIFF
--- a/.licenseignore
+++ b/.licenseignore
@@ -9,3 +9,4 @@
 **/*.modulemap
 CODEOWNERS
 Package.swift
+Scripts/git-swift-format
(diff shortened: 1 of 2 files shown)


allevato:
Adding this script to `.licenseignore` doesn't seem great, but I don't see an obvious way to tell the license checker script to treat the extension-less script as if it were a `.py` file. From what I can tell it just fails for anything without an extension.

dylansturg (approved):

PR #1240, opened 2026-07-02
Update CMakeLists.txt

tshortli:
In https://github.com/swiftlang/swift-format/pull/1236/ files were renamed and added but CMakeLists.txt was not updated, breaking swiftlang CI builds on Windows.

### DIFF
--- a/Sources/SwiftFormat/CMakeLists.txt
+++ b/Sources/SwiftFormat/CMakeLists.txt
@@ -19,6 +19,7 @@ add_library(SwiftFormat
   API/SwiftFormatError.swift
   API/SwiftFormatter.swift
   API/SwiftLinter.swift
+  Core/Attributes+Convenience.swift
   Core/Context.swift
   Core/DocumentationComment.swift
   Core/DocumentationCommentText.swift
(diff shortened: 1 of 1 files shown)


rintaro (approved):

allevato:
Sorry about that, thanks for the fix.

I thought our PR testing exercised the CMake build at least on Windows. I guess that's not the case? We should fix this because adding or renaming a source file shouldn't require doing the full `swift-ci test` flow just to cover one platform.

rintaro:
swift-tools-protocols is testing CMake builds in PR GitHub actions
https://github.com/swiftlang/swift-tools-protocols/blob/0a4bf639ede626ee9776714b26738bd47a2345db/.github/workflows/pull_request.yml#L29-L43
But it's implemented as a SwiftPM plugin.
https://github.com/swiftlang/swift-tools-protocols/blob/0a4bf639ede626ee9776714b26738bd47a2345db/Package.swift#L174-L183
I don't really know why it's like so.

rintaro:
Changing it to shell scripts so swift-format can easily adopt it
https://github.com/swiftlang/swift-tools-protocols/pull/71

PR #1244, opened 2026-07-10
Rework `ImportsXCTestVisitor` class

### LINKED ISSUE #1243: `NeverUseForceTry` is applied to Swift Testing test code.
The `NeverUseForceRetry` is documents as follows:

> NeverUseForceTry
> 
> Force-try (try!) is forbidden.
> 
> This rule does not apply to test code, defined as code which:
> 
> Contains the line import XCTest
> The function is marked with @Test attribute
> Lint: Using try! results in a lint error.
> 
> TODO: Create exception for NSRegularExpression
> 
> NeverUseForceTry is a linter-only rule.

I have a force retry in Swift Testing code, and the lint violating is being emitted.

Here's sample code that was found in a Swift Package test target.

```swift
import Testing

public struct ValidationError: Error, CustomStringConvertible {
  public internal(set) var message: String

  public init(_ message: String) {
    self.message = message
  }

  public var description: String {
    message
  }
}

struct Name {
  var rawValue: String

  init(rawValue: String) throws {
    if rawValue == "bad" {
      throw ValidationError("Bad input for name")
    }
    self.rawValue = rawValue
  }
}

private struct Bar {
  var firstName: Name = try! Name(rawValue: "none")
}
```

bkhouri:
The `ImportsXCTestVisitor` class was too restrictive.  Rename the class to `ImportsAnyTestingLibrary` and have the visitor class support both XCTest and/or Swift Testing library.

The list of "supported" testing modules are store in a property, which can be augmented in the future.

Closes #1243


bkhouri:
The **API breakage check** build failed with the following

```
3 breaking changes detected in SwiftFormat:
  💔 API breakage: var Context.importsXCTest has declared type change from SwiftFormat.Context.XCTestImportState to SwiftFormat.Context.AnyTestImportState
  💔 API breakage: accessor Context.importsXCTest.Get() has return type change from SwiftFormat.Context.XCTestImportState to SwiftFormat.Context.AnyTestImportState
  💔 API breakage: accessor Context.importsXCTest.Set() has been removed
```

These are expected as those have been deprecated and replaced with another.  how do we address this?

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
```diff
@@ -0,0 +1,75 @@
...
+    // If we already know whether or not a supported test library is imported, don't bother doing
+    // anything else.
+    guard context.importsAnyTestLibrary == .notDetermined else { return .skipChildren }
+
+    // If the first import path component matches one of the supported test library module names,
+    // record that fact. Checking in this way lets us catch a plain `import <Library>` but also
+    // specific decl imports like `import class <Library>.SomeType`, if someone wants to do that.
+    let containsImports = supportedTestLibraryModuleNames.map { node.path.first!.name.tokenKind == .identifier($0) }
+
+    if containsImports.reduce(false, { x, y in x || y }) {
```
allevato: Why not just:

```
if supportedTestLibraryModuleNames.contains(where: { node.path.first!.name.tokenKind == .identifier($0) }) {
```

Could put a fileprivate extension on `ImportDeclSyntax` to get rid of the inlined closure, if you prefer. I'm mainly commenting on the `reduce` part, which is just a `contains` in disguise.

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+
+package let supportedTestLibraryModuleNames = [
```
allevato, line 15: This is nice, because in the future we can consider making this configurable (someone could add Nimble, Quick, we could add things like StoreKitTest, and so forth).

allevato (commented):
> The **API breakage check** build failed with the following
> 
> ```
> 3 breaking changes detected in SwiftFormat:
>   💔 API breakage: var Context.importsXCTest has declared type change from SwiftFormat.Context.XCTestImportState to SwiftFormat.Context.AnyTestImportState
>   💔 API breakage: accessor Context.importsXCTest.Get() has return type change from SwiftFormat.Context.XCTestImportState to SwiftFormat.Context.AnyTestImportState
>   💔 API breakage: accessor Context.importsXCTest.Set() has been removed
> ```
> 
> These are expected as those have been deprecated and replaced with another. how do we address this?

Unfortunately you'll just need to add them to the breakage file. The API digester doesn't consider SPI, IIRC, so it just treats them like any other public declaration even if they're not supposed to be used outside of the package.

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
bkhouri: This was my exact thinking for pulling out the "supported test libraries" in a "global property".  I wanted to also include a list of test libraries provided by the `.swift-format` configuration, but as I briefly looked at the code base, I opted to not make this change in this PR.

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
bkhouri: I was looking for a Swift equivalent of python `any()`.  I updated to use your suggestion.

allevato (approved):
Thanks!

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
```diff
@@ -0,0 +1,70 @@
...
+  "XCTest",
+  "Testing",
+]
+
+/// A visitor that determines if the target source file imports a supported test library.
+private class ImportsAnyTestingLibrary: SyntaxVisitor {
```
rintaro: I don't think `SyntaxVisitor` is the right choice for this kind of operation. `import` is only valid at the top-level, so we should just iterate `SourceFileSyntax.statements[].item.as(ImportDeclSyntax.self)`

bkhouri: This is a rename of `ImportsXCTestVisitor`. For some reason, GitHub is showing this and a whole new file. The main delta I is renaming the class, file, and to return `true` if any of the test libraries are imported, which was moved to a "file property" array.

bkhouri: The function has been updated to not use a `SyntaxVisitor`

allevato: It's not entirely true that we can only look at top-level statements. The suggested change won't catch this:

```swift
#if SOME_FEATURE_IS_ENABLED
  import XCTest

  final class SomeFeatureTests: XCTestCase {}
#endif
```

That was the main reason for having this be a visitor, so we don't have to manually recurse through arbitrarily deep `IfConfigDecl`s. Can you change it back?

bkhouri: @allevato : I have a test that checks the conditionals, and it also checks nested conditionals (two levels)

```swift
      TestCase(
        id: "\(library): import test library inside nested conditional",
        source: """
          import Foundation
          #if SOME_FEATURE_FLAG
            import FooBar
            #if os(macOS)
              import \(library)
            #endif
          #endif
          """,
        expected: .importsATestLirary
      ),
```

With that said, I'm happy to change it back, but I would like @allevato and @rintaro (both code owners) to agree on the logic to use here :)

allevato: I feel like this change has spun a bit out of control from just "look for `Testing` instead of `XCTest`" but since you've already written the code, it seems silly to revert it. I'd make sure to include some tests that cover `#else`/`#elseif`, just so we have confidence that it handles all the clauses correctly.

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
```diff
@@ -0,0 +1,62 @@
...
+    } else if let ifConfigDecl = codeBlockItem.item.as(IfConfigDeclSyntax.self) {
+      for clause in ifConfigDecl.clauses {
+        guard let nestedStatements = clause.elements?.as(CodeBlockItemListSyntax.self) else {
+          continue
+        }
+        if statementsImportAnyTestLibrary(nestedStatements) {
+          return true
+        }
+      }
+    }
```
allevato: I usually prefer to write choices like this as a `switch` over `codeBlockItem.item.as(SyntaxEnum.swift)` to avoid the repeated casts, since you can extract the associated value of the correct type from that easily.

Tests/SwiftFormatTests/Rules/SetImportsAnyTestLibraryTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        source: """
+          import Foundation
+          """,
+        expected: .doesNotImportATestLibrary
+      ),
+      TestCase(
+        id: "imports all supported test libraries",
+        source: """
+          import Foundation
+          """ + Self.importsAllTestLibraries,
```
bkhouri, line 145: I originally had the following, but the Windows and Linux Swift 6.0 encountered a type-checker issue, which is fixed in later Swift versions.

```suggestion
        source: """
          import Foundation
          \(supportedTestLibraryModuleNames.map { "import \($0)" }.joined(separator: "\n"))
          """,
```

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
bkhouri: PR has been updated with additional tests.  Please let me know if you'd like any additional tests added.

Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
bkhouri: Updated.

allevato (approved):

rintaro (approved):

compnerd:
This seems to have broken CI:
```
10:46:47  C:\Windows\system32\cmd.exe /C "cd . && T:\x86_64-unknown-windows-msvc\Stage1Compilers\toolchain\usr\bin\swiftc.exe -target x86_64-unknown-windows-msvc -j 36 -num-threads 36 -emit-library -o bin\SwiftFormat.dll -module-name SwiftFormat -module-link-name SwiftFormat -emit-module -emit-module-path swift\SwiftFormat.swiftmodule -emit-dependencies -DSwiftFormat_EXPORTS -sdk "T:/Program Files/Swift/Platforms/Windows.platform/Developer/SDKs/Windows.sdk" -gnone -Xlinker /INCREMENTAL:NO -Xlinker /OPT:REF -Xlinker /OPT:ICF -O -wmo -libc MD -output-file-map Sources\SwiftFormat\CMakeFiles\SwiftFormat.dir\Release\output-file-map.json -I T:\x86_64-unknown-windows-msvc\Markdown\swift -I C:\Users\swift-ci\jenkins\workspace\swift-main-windows-toolchain\swift-markdown\Sources\CAtomic\include -I T:\x86_64-unknown-windows-msvc\Stage2Compilers\.\lib\swift\host -I "T:\Program Files\Swift\Toolchains\0.0.0+Asserts\usr\include\cmark_gfm" -I "T:\Program Files\Swift\Toolchains\0.0.0+Asserts\usr\include\cmark_gfm_extensions" @CMakeFiles\SwiftFormat.rsp    -Xlinker -implib:lib\SwiftFormat.lib && cd ."
10:46:47  C:\Users\swift-ci\jenkins\workspace\swift-main-windows-toolchain\swift-format\Sources\SwiftFormat\Core\ImportsAnyTestingLibrary.swift:15:1: error: the package access level used on 'supportedTestLibraryModuleNames' requires a package name; set it with the compiler flag -package-name

10:46:47  13 | import SwiftSyntax

10:46:47  14 | 

10:46:47  15 | package let supportedTestLibraryModuleNames = [

10:46:47     | `- error: the package access level used on 'supportedTestLibraryModuleNames' requires a package name; set it with the compiler flag -package-name

10:46:47  16 |   "XCTest",

10:46:47  17 |   "Testing",

10:46:47  


```
https://ci-external.swift.org/job/swift-main-windows-toolchain/2206/console

allevato:
@compnerd That should be fixed by https://github.com/swiftlang/swift-format/pull/1247.

PR #1241, opened 2026-07-02
CI: cmake-smoke-test workflow

rintaro:
Add a cmake-smoke-test job to the pull request workflow, mirroring the setup in swift-tools-protocols. It builds swift-format with CMake on Linux, macOS, and Windows to validate the CMake build.

The supporting scripts under `.github/scripts/` install `cmake/ninja` (`prebuild.{sh,ps1}`) and drive the CMake configure/build (`cmake-smoke-test.{sh,ps1}`).

PR #1247, opened 2026-07-15
Update access level

### LINKED ISSUE #1243: `NeverUseForceTry` is applied to Swift Testing test code.

bkhouri:
Uupdate the access level of a function to match the deprecated equivalent.

Closes #1243
Issue: rdar://181871593

### DIFF
--- a/api-breakages.txt
+++ b/api-breakages.txt
@@ -1,6 +1,7 @@
 6.4
 ---
 API breakage: accessor Context.importsXCTest.Set() has been removed
+API breakage: var supportedTestLibraryModuleNames is now with @_spi
 
 6.2
 ---
--- a/Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
+++ b/Sources/SwiftFormat/Core/ImportsAnyTestingLibrary.swift
@@ -12,7 +12,8 @@
 
 import SwiftSyntax
 
-package let supportedTestLibraryModuleNames = [
+@_spi(Testing)
+public let supportedTestLibraryModuleNames = [
   "XCTest",
   "Testing",
 ]
@@ -35,7 +36,8 @@ public func setImportsAnyTestLibrary(context: Context, sourceFile: SourceFileSyn
   }
 }
 
-private func statementsImportAnyTestLibrary(_ statements: CodeBlockItemListSyntax) -> Bool {
+@_spi(Testing)
+public func statementsImportAnyTestLibrary(_ statements: CodeBlockItemListSyntax) -> Bool {
   for codeBlockItem in statements {
     switch codeBlockItem.item {
     case .decl(let decl):


allevato (commented):
Since I can't see the Radar, what was the actual issue here? This was new code, why does it need to be public if its only caller is in this file?

bkhouri:
> Since I can't see the Radar, what was the actual issue here? This was new code, why does it need to be public if its only caller is in this file?

Some Windows PR CI build are failing with the following

```
C:\Users\swift-ci\jenkins\workspace\swift-PR-windows\swift-format\Sources\SwiftFormat\Core\ImportsAnyTestingLibrary.swift:15:1: error: the package access level used on 'supportedTestLibraryModuleNames' requires a package name; set it with the compiler flag -package-name
```

I'm hoping this will fix the issue.

allevato:
@rintaro has been working on a CMake smoke test PR action in https://github.com/swiftlang/swift-format/pull/1241, but I believe it's blocked on some of our dependencies' repos needing tweaks.

Rather than add another CMake workflow in this PR, we should just land the code fix (assuming it fixes the CI problem) and then converge on that check.

Longer term, being unable to use basic Swift language features like `package` in the CMake builds is unsustainable and we need to figure out what's needed to get those working correctly.

bkhouri:
I was unaware of @rintaro work in adding a CMake smoke build.  I still think there is value adding this CMake build in this PR to prevent accidentally breaking the CMake build until #1241 lands.

bkhouri:
Will close and re-open the PR to trigger the Actions that failed due to an infrastructure issue

allevato:
I'll defer to @rintaro whether he wants another CMake workflow added in the meantime. CMake isn't really my wheelhouse.

rintaro:
let's just merge this _without the CMake workflow_.

Until https://github.com/swiftlang/swift-format/pull/1241 is merged, we need to do a cross-repo testing to exercise CMake build, which is also no unsustainable. So I will land #1241 asap.

bkhouri:
Although I don't agree with removing the CMake build as it gives additional coverage, I did remove it to unblock the fix ASAP.

api-breakages.txt
```diff
@@ -1,6 +1,7 @@
 6.4
 ---
 API breakage: accessor Context.importsXCTest.Set() has been removed
+API breakage: var supportedTestLibraryModuleNames is now with @_spi
```
rintaro, line 4: I'm not sure we want this. Was this ever a public API?

allevato: Is this an artifact of the API differ seeing that it was previously `package` level without SPI and now having an SPI added to it? If so, I don't know if we're able to avoid the false/harmless positive.

bkhouri: exactly what @allevato said.  This "property" was added in #1244.

rintaro (approved):

Issue #1235, opened 2026-06-30
There should be a way to join hard-wrapped line comments


itingliu:
## Background

swift-format supports reflowing prose in some scenarios: 
- `respectsExistingLineBreaks: false` collapses manually-wrapped **code** that fits within `lineLength`
- `reflowMultilineStringLiterals` does the same for multiline **string literals**. 
 
Line comments should be supported too.

## Current behavior
When `respectsExistingLineBreaks == false` and `lineLength` is permissive (e.g. 250), the formatter will collapse a hand-broken function call onto one line but leave this untouched:

```swift
// This comment was written at an 80-column
// terminal and never updated.
```

There is no existing rule or configuration flag for this. This results in inconsistent format when the project with wider line lengths use `respectsExistingLineBreaks: false` to collapse code

itingliu:
I don't think this is the same? I want to combine the comments into one line, but the other issue #469 is saying it doesn't respect split into multiple lines? @allevato?

allevato:
You're right! I misread it. Reopening.

That being said, I imagine that this could be implemented as part of a single change that would simply look for any contiguous blocks of line comments and treat them as a single entity for the purposes of wrapping, solving both issues. That's the obvious thing to do for doc comments, so line comments could be done the same.

PR #1251, opened 2026-07-21
Add ReflowComments rule

itingliu:
This opt-in rule joins comments that were hard-wrapped across multiple
lines back into a single line when the combined text still fits within
the configured line length.

It reflows only line comments (`//`) and documentation line comments
(`///`), and only joins lines that belong to the same Markdown paragraph.
Lists, headings, code blocks, block quotes, thematic breaks, blank lines
between paragraphs, divider lines such as `//===----===//`, and file
headers are left untouched. The `reflowComments.reflowedCommentKinds` and
`reflowComments.preservedLinePrefixes` options control which comments the
rule reflows.

### DIFF
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -41,6 +41,7 @@ Here's the list of available rules:
 - [OneVariableDeclarationPerLine](#OneVariableDeclarationPerLine)
 - [OnlyOneTrailingClosureArgument](#OnlyOneTrailingClosureArgument)
 - [OrderedImports](#OrderedImports)
+- [ReflowComments](#ReflowComments)
 - [ReplaceForEachWithForLoop](#ReplaceForEachWithForLoop)
 - [ReturnVoidInsteadOfEmptyTuple](#ReturnVoidInsteadOfEmptyTuple)
 - [SwiftTestingNamingConventions](#SwiftTestingNamingConventions)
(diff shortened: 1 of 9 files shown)


itingliu:
Resolves #1235

allevato:
Thanks for taking a look at this!

I think the problem is a bit trickier than it appears, and a pure syntax-node-based rule isn't going to suffice I'm afraid. Since this behavior deals with line breaking/reflowing, it can't really be applied until the pretty printer phase. Consider this example:

```swift
struct S {
func f() {
// these lines are intentionally unindented, assume the column
// width is 80 and these lines will go past column 80
}
}
```

swift-format works in two discrete phases:

* First, syntax-node-based rules are applied. These only deal with structural correctness, not line breaking or indentation.
* Second, the pretty printer reflows the tokens based on line breaking rules, recomputing indentation and taking column width into account.

Since `ReflowComments` is a node-based rule, it will run before the correct indentation for those lines is known. Assuming an indentation width of 2 spaces, if you reflow this comment in its current form, you're reflowing it based on its current incorrect indentation (zero) and effectively granting it 4 more spaces than it should actually have once properly indented.

I think the proper (and only correct) way to do this is to update the pretty printer to treat clusters of whole-line comments as atomic units. The place to start would be looking at how we handle comments in `TokenStreamCreator` and `PrettyPrinter`. For block comments, we already gather the whole (multiline) comment as a block so that we can write it back out. For line comments, we only take a line at a time. I think what we'd need to do instead is fold multiple line comments into a single `Comment` value to pass to the pretty printer. Then when we process that formatting token in the pretty printer, it's happening right before we're supposed to print it back out, so we know the _exact_ current indentation and the column limit and we can compute the difference to get the correct width for the text.

itingliu:
> I think the proper (and only correct) way to do this is to update the pretty printer to treat clusters of whole-line comments as atomic units. The place to start would be looking at how we handle comments in `TokenStreamCreator` and `PrettyPrinter`. For block comments, we already gather the whole (multiline) comment as a block so that we can write it back out. For line comments, we only take a line at a time. I think what we'd need to do instead is fold multiple line comments into a single `Comment` value to pass to the pretty printer. Then when we process that formatting token in the pretty printer, it's happening right before we're supposed to print it back out, so we know the _exact_ current indentation and the column limit and we can compute the difference to get the correct width for the text.

Thanks for the pointer! I underestimated this overwhelmingly since indentation has never been a problem in the repos that I work with 😅

I'll take a stab with this direction and come back with whatever I have. Thanks!

PR #1252, opened 2026-07-22
Fix broken Configuration link in RuleDocumentation

ZayanKhan-12:
`Documentation/RuleDocumentation.md` links `Documentation/Configuration.md` — from inside `Documentation/`, that resolves to `Documentation/Documentation/Configuration.md` and 404s. The file sits in the same directory → `Configuration.md`.

One-line docs fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -4,7 +4,7 @@
 
 Use the rules below in the `rules` block of your `.swift-format`
 configuration file, as described in
-[Configuration](Documentation/Configuration.md). All of these rules can be
+[Configuration](Configuration.md). All of these rules can be
 applied in the linter, but only some of them can format your source code
 automatically.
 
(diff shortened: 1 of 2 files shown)


allevato (commented):
I could have sworn we'd made this change before, and we have! It turns out the real culprit is in the file's generator, so every time we generate the code we end up going back to the broken link: https://github.com/swiftlang/swift-format/blob/main/Sources/_GenerateSwiftFormat/RuleDocumentationGenerator.swift#L36

Can you fix it there as well, and then future generations will stay correct?

ZayanKhan-12:
Good catch, thank you! Fixed at the source: `RuleDocumentationGenerator.swift` now emits `Configuration.md`, so future `generate-swift-format` runs keep the link correct. The generated file's fix stays in the PR so the two are in sync.

allevato (approved):
Thanks!

PR #1253, opened 2026-07-23
[PrettyPrint] Keep ternary formatting idempotent

### LINKED ISSUE #1248: 6.3.0: Non-idempotent output with long lines and ternary operator
The following input (reduced from generated code, don't blame me)
```swift
struct Outer {
    func test() {
        let ffiStructKeepAlive: (FixedByteArrayConverter<FixedByteArrayHelper15>.KeepAlive?, IdentityConverter<Int32>.KeepAlive?, )? = (media_id_keepalive != nil || cdn_keepalive != nil || false) ? (media_id_keepalive, cdn_keepalive,) : nil
    }
}
```
formats to
```swift
struct Outer {
    func test() {
        let ffiStructKeepAlive:
            (FixedByteArrayConverter<FixedByteArrayHelper15>.KeepAlive?, IdentityConverter<Int32>.KeepAlive?, )? =
                (media_id_keepalive != nil || cdn_keepalive != nil || false) ? (media_id_keepalive, cdn_keepalive,)
                : nil
    }
}
```
but if you run it through the formatter again you get
```swift
struct Outer {
    func test() {
        let ffiStructKeepAlive:
            (FixedByteArrayConverter<FixedByteArrayHelper15>.KeepAlive?, IdentityConverter<Int32>.KeepAlive?, )? =
                (media_id_keepalive != nil || cdn_keepalive != nil || false)
                ? (media_id_keepalive, cdn_keepalive,)
                : nil
    }
}
```

Our .swift-format:
```json
{
    "version": 1,
    "lineLength": 120,
    "indentation": { "spaces": 4 },
    "indentConditionalCompilationBlocks": false,
    "lineBreakBeforeEachArgument": true,
    "lineBreakBetweenDeclarationAttributes": true,
    "prioritizeKeepingFunctionOutputTogether": true
}
```

thliu21:
## Summary

- prevent a formatter-inserted newline before a ternary colon from changing the preceding `?` break on the next formatting pass
- continue honoring the newline when the colon's leading trivia contains a comment
- add regression coverage for the reported long typed declaration and for comments before the colon

## Testing

- `swift test --filter 'TernaryExprTests|RespectsExistingLineBreaksTests'`
- `swift test --parallel`
- formatted both changed files with the modified formatter and confirmed no diff

Fixes #1248

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
+++ b/Sources/SwiftFormat/PrettyPrint/PrettyPrint.swift
@@ -426,8 +426,8 @@ public class PrettyPrinter {
 
       var overrideBreakingSuppressed = false
       switch newline {
-      case .elective, .escaped: break
-      case .soft(_, let discretionary):
+      case .elective, .electiveIgnoringGroupLength, .escaped: break
+      case .soft(_, let discretionary), .softIgnoringGroupLength(_, let discretionary):
         // A discretionary newline (i.e. from the source) should create a line break even if the
         // rules for breaking are disabled.
         overrideBreakingSuppressed =
@@ -667,7 +667,7 @@ public class PrettyPrinter {
         delimIndexStack.append(i)
 
         switch newline {
-        case .elective, .escaped:
+        case .elective, .electiveIgnoringGroupLength, .softIgnoringGroupLength, .escaped:
           total += size
         default:
           // `size` is never used in this case, because the break always fires. Use `maxLineLength`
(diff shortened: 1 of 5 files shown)


allevato:
It looks like this removes some discretionary newlines in places where we would like to allow them; see https://github.com/swiftlang/swift-format/actions/runs/30044416399/job/89485056904?pr=1253.

Can you fix the idempotence issue without breaking those cases?

thliu21:
Thanks for the report. I updated the ternary newline handling so a user-authored discretionary break before `:` is preserved without forcing the earlier `?` break during the next formatting pass. I added a regression test for that case; the existing idempotence test and the full local suite now pass. I also re-ran the reported swift-syntax files locally with no formatting diff.

Pushed in 9e870fc.

PR #1256, opened 2026-08-02
Skip self-hosted CI jobs on forks

kkebo:
PR #1241 has added the `cmake-smoke-test` job, and it requires a self-hosted runner for macOS. However, self-hosted runners are not available for forked repositories. This causes CI to hang and fail after a long period of time on forked repositories. This is frustrating for contributors.

- https://github.com/swiftlang/github-workflows/issues/305

To solve that, I've changed `enable_macos_checks` of `cmake-smoke-test` from `true` to `${{ github.repository == 'swiftlang/swift-format' }}`. If the workflow is running in the upstream repository, macOS checks remain enabled. In forked repositories, they are skipped.

Similar PR: https://github.com/swiftlang/swiftly/pull/471


kkebo:
Rebased.

kkebo:
Rebased.

kkebo:
Could you please rerun the failing "Check compatibility with real world projects" job? It looks flaky.

PR #1258, opened 2026-08-03
fix: avoid breaking after discard assignments

### LINKED ISSUE #1033: Weird formatting (extra line break) around assigning to `_`
```swift
func test() {
    _ = try! ServiceId.parseFrom(
        serviceIdString: Self.TEST_UUID_STRING) as! Aci
}
```

formats as

```swift
func test() {
    _ =
             try! ServiceId.parseFrom(
                 serviceIdString: Self.TEST_UUID_STRING
             ) as! Aci
}
```

I wouldn't expect the line break after `=`, especially with a 4-space indent but even with a 2-space one. That might apply to any sufficiently short variable name, but if that's too detailed a rule than a special case for `_` could still be an improvement.

aryansk:
Fixes #1033.

When a discard assignment has a compound right-hand side, the pretty printer can add an unnecessary break immediately after `_ =`. This change avoids adding that outer grouping for discard assignments while preserving the existing grouping behavior for normal assignments.

Validation:
- `swift test --parallel` (943 tests)
- `swift test --filter AssignmentExprTests`
- strict formatter lint on changed files
- `git diff --check`

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -2200,7 +2200,10 @@ private final class TokenStreamCreator: SyntaxVisitor {
         // When the RHS is a simple expression, even if is requires multiple lines, we don't add a
         // group so that as much of the expression as possible can stay on the same line as the
         // operator token.
-        if isCompoundExpression(rhs) && leftmostMultilineStringLiteral(of: rhs) == nil {
+        if isCompoundExpression(rhs)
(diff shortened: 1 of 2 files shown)


allevato:
This would be a format-breaking change (as in the format of existing code would change without a change to the configuration), so we can't accept this without making it configurable. However, I also don't know if a configuration option here would hold its weight; it seems like more complexity for relatively little benefit.

aryansk:
Closing this draft after maintainer feedback that the proposed formatting change is format-breaking and cannot be accepted without configurability. I am not pursuing a configurable variant in this contribution cycle.

PR #1259, opened 2026-08-05
Recognize CJK full stops in the non-linguistic sentence fallback

### LINKED ISSUE #1237: Platform differences in `BeginDocumentationCommentWithOneLineSummary`
## Code

demo.swift:

```swift
/// 何かする関数。
func doSomething() {
  // do something
}

/// 何かする関数．
func doSomething2() {
  // do something
}
```

.swift-format:

```json
{
  "rules" : {
    "BeginDocumentationCommentWithOneLineSummary" : true
  },
  "version" : 1
}
```

## Expected behavior

`swift format lint demo.swift` should behave identically on all supported platforms.

## Actual behavior

macOS:

```console
$ swift format lint demo.swift
(No warnings)
```

Linux:

```console
$ swift format lint demo.swift
demo.swift:2:1: warning: [BeginDocumentationCommentWithOneLineSummary] terminate this sentence with a period: "何かする関数。"
demo.swift:7:1: warning: [BeginDocumentationCommentWithOneLineSummary] terminate this sentence with a period: "何かする関数．"
```

## The root cause

These platform differences are caused by the difference in implementation between macOS and the others. While the NaturalLanguage framework is used for macOS, the simple `.hasSuffix(".")` is used for the other platforms. Both [`。`](https://en.wikipedia.org/wiki/Japanese_punctuation#Full_stop) and `．` are common [full stops](https://en.wikipedia.org/wiki/Full_stop) in Japanese. The NaturalLanguage framework appears to support them. This is why the platform differences have arisen.

https://github.com/swiftlang/swift-format/blob/b42c1debf3de4667b2d7385edd7e5e7812aeb825/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift#L133-L178

adityasingh2400:
`BeginDocumentationCommentWithOneLineSummary` uses NaturalLanguage on Darwin and a period-splitting fallback everywhere else, and the fallback only understood the ASCII period. A doc summary such as `/// 何かする関数。` therefore linted clean on macOS but produced "terminate this sentence with a period" on Linux. The fallback now scans for any full stop, adding U+3002, U+FF0E, and U+FF61, which NLTagger already reports as sentence terminators, and ends a sentence at each one. An ASCII period still only ends a sentence when a space or the end of the text follows it, so decimal points and abbreviations behave exactly as before, and the CJK stops end a sentence directly because they are not written with a following space. Verified with two new tests, one per implementation, covering a single sentence, two sentences on one line, and an unterminated summary. The fallback test fails on main and passes here, and the full `BeginDocumentationCommentWithOneLineSummaryTests` suite passes.

Note that this deliberately covers only full stops. NLTagger also treats `!` and `?` as sentence terminators, but accepting those in the fallback would change what the rule asks for rather than just aligning the platforms, so I left that out.

Fixes #1237

### DIFF
--- a/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
+++ b/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
@@ -184,7 +184,12 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
   /// halfwidth forms, which are the ordinary full stops in CJK text. The linguistic APIs on Apple
   /// platforms already report all of these as sentence terminators, so recognizing them here keeps
   /// the two implementations in agreement.
-  private static let fullStops: Set<Character> = ["\u{002E}", "\u{3002}", "\u{FF0E}", "\u{FF61}"]
+  private static let fullStops: Set<Character> = [
+    "\u{002E}",  // FULL STOP
+    "\u{3002}",  // IDEOGRAPHIC FULL STOP
+    "\u{FF0E}",  // FULLWIDTH FULL STOP
+    "\u{FF61}",  // HALFWIDTH IDEOGRAPHIC FULL STOP
+  ]
 
   /// Returns the best approximation of sentences in the given text using string splitting around
   /// full stops.
--- a/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
+++ b/Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
@@ -178,8 +178,16 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
     #endif
   }
 
+  /// The characters that the fallback implementation treats as full stops.
+  ///
+  /// Besides the ASCII period, this contains the ideographic full stop and its fullwidth and
+  /// halfwidth forms, which are the ordinary full stops in CJK text. The linguistic APIs on Apple
+  /// platforms already report all of these as sentence terminators, so recognizing them here keeps
+  /// the two implementations in agreement.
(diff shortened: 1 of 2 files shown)


Sources/SwiftFormat/Rules/BeginDocumentationCommentWithOneLineSummary.swift
```diff
@@ -178,8 +178,16 @@ public final class BeginDocumentationCommentWithOneLineSummary: SyntaxLintRule {
     #endif
   }
 
+  /// The characters that the fallback implementation treats as full stops.
+  ///
+  /// Besides the ASCII period, this contains the ideographic full stop and its fullwidth and
+  /// halfwidth forms, which are the ordinary full stops in CJK text. The linguistic APIs on Apple
+  /// platforms already report all of these as sentence terminators, so recognizing them here keeps
+  /// the two implementations in agreement.
+  private static let fullStops: Set<Character> = ["\u{002E}", "\u{3002}", "\u{FF0E}", "\u{FF61}"]
```
allevato: It wouldn't be a bad idea to put each of these on their own line with a trailing comment that says the official Unicode name of each of these code points.

adityasingh2400: Good call, done in 05dfb91.

```swift
private static let fullStops: Set<Character> = [
  "\u{002E}",  // FULL STOP
  "\u{3002}",  // IDEOGRAPHIC FULL STOP
  "\u{FF0E}",  // FULLWIDTH FULL STOP
  "\u{FF61}",  // HALFWIDTH IDEOGRAPHIC FULL STOP
]
```

I took the names from `unicodedata` rather than typing them from memory, so they are the official character names for those code points.

Build is clean, the 10 `BeginDocumentationCommentWithOneLineSummary` tests pass, and `swift format lint --strict` is happy with the file.

allevato (approved):
Seems reasonable to me!

PR #1261, opened 2026-08-06
Revert "Add regression tests for config search on paths with redundant separators (#1213)

tshortli:
This reverts commit 16d9f1afaf6d4c34c95d6e270851c45b89671254 because it appears to be causing timeouts on macOS in Swift CI.


allevato (approved):
Interesting; I wonder what's different about the full CI environment here vs. the GitHub PR checks.

tshortli:
I don't have a merge button so I'll need someone who does to merge when appropriate.

allevato:
Unfortunately I don't believe I have permissions to override the check requirement here. @rintaro or @shahmishal might be able to land this.

PR #1263, opened 2026-08-15
Don't break after a block comment the user kept inline

### LINKED ISSUE #872: Formatted code does not compile anymore when throwing an error with a comment between the error type and its init arguments with the comments crossing the max column width
Alright the title says it all but an example will probably help:
```swift
throw MyError/*Some comment; why not? Swift syntax allows it, though it’s very very weird…*/()
```

This will be formatted like this (assuming 80-col max width and tab indents):
```swift
throw
	MyError /*Some comment; why not? Swift syntax allows it, though it’s very very weird…*/
()
```
This does not compile.

ayaangazali:
Fixes #872.

**What's wrong today.** An inline block comment produces output that doesn't compile:

```swift
throw MyError/*Some comment; why not? Swift syntax allows it, though it's very very weird…*/()
```

at 80 columns becomes

```swift
throw
    MyError /*Some comment; why not? Swift syntax allows it, though it's very very weird…*/
()
```

Subscripts split the same way, `values/*long comment*/[0]` puts `[0]` on its own line.

**Why.** Both places that emit a trailing block comment append an elective `.break(.same)` right after it (`afterTokensForTrailingComment`, and the `.blockComment` case in `extractLeadingTrivia`). The comment on the first one says what it is for: preserving a discretionary newline the user wrote after the comment. But an elective break is also a wrap candidate, and this one sits wherever the comment happens to be, including between a callee and its argument list, where a newline is not legal. The comment being long is what pushes the line over the limit and fires it.

**What this does.** Emit the break only when the source actually has a newline after the comment. If the comment and the next token were adjacent in the source, there is no discretionary newline to preserve, so the break has nothing left to do except cause the wrong wrap. Where the break is dropped in `extractLeadingTrivia`, a plain space takes its place so the existing `breakSize` spacing is unchanged.

The second site is included because it is the same defect: with two adjacent block comments (`MyError/*a*/ /*b*/()`) the break comes from that path instead, and splits the call the same way.

**Verification.** I can't run `swift test` on this machine (no Xcode, so no `XCTest` for `_SwiftFormatTestSupport`), so I leaned on differential testing instead and the suite will run here in CI.

- Built `swift-format` before and after and formatted every `.swift` file in sourcekit-lsp (326), swift-syntax (320), and this repo's `Sources` and `Tests` (258) with each project's own config. 904 files, zero output differences, which is what the compatibility job checks.
- Pulled all 94 multiline literals containing a block comment out of `Tests/SwiftFormatTests` and formatted each with both binaries across line lengths 10 through 120 and across `indentBlankLines`, `respectsExistingLineBreaks`, `lineBreakBeforeEachArgument` and tab indentation. The only differences appear at line lengths below 40; every one of those snippets belongs to a test that runs at 60 or 100, where output is identical. Included #1169's two `RespectsExistingLineBreaks` cases explicitly since they touch the same `.blockComment` branch.
- New output parses under `swiftc -parse` and is idempotent.

The added test covers the call, the subscript, and a user newline after a block comment still being honored, so it fails on all three counts without the fix.

Also worth flagging that CI on this will need a maintainer to approve the run, since it's my first PR here.

---

apologies if there are any obvious errors in here, i worked through the logic myself and used claude code as a sounding board on the design choices. if i got something wrong i'd genuinely like to hear it. freshman in college, trying to be useful where i can :)

### DIFF
--- a/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
+++ b/Sources/SwiftFormat/PrettyPrint/TokenStreamCreator.swift
@@ -3485,17 +3485,20 @@ private final class TokenStreamCreator: SyntaxVisitor {
       )
 
     case .blockComment(let text):
-      return (
-        false,
-        [
-          .space(size: 1, flexible: true),
-          .comment(Comment(kind: .block, leadingIndent: nil, text: text), wasEndOfLine: false),
-          // We place a size-0 break after the comment to allow a discretionary newline after
-          // the comment if the user places one here but the comment is otherwise adjacent to a
-          // text token.
-          .break(.same, size: 0),
-        ]
-      )
+      var tokens: [Token] = [
+        .space(size: 1, flexible: true),
+        .comment(Comment(kind: .block, leadingIndent: nil, text: text), wasEndOfLine: false),
+      ]
+      // We place a size-0 break after the comment to allow a discretionary newline after the
+      // comment if the user places one here but the comment is otherwise adjacent to a text
+      // token. It's only added when the user actually wrote that newline: otherwise it is an
+      // elective break in a spot the pretty printer never chose to break, and firing it to
+      // satisfy the line length can separate tokens that have to stay glued together, like a
+      // callee from its argument list.
+      if trivia.dropFirst().contains(where: { $0.isNewline }) {
+        tokens.append(.break(.same, size: 0))
+      }
+      return (false, tokens)
 
     default:
       return (false, [])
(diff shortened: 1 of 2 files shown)


ayaangazali:
Status from my side, no changes needed so far. I rebased this locally on 9c9a9fa and it still applies cleanly, and re-ran the reproducer from #872 against that commit to confirm the fix still produces the right output. I have not force-pushed, since the branch is already conflict free and a no-op push would only add noise here.

The checks on this one have never started. I am fairly sure that is just the first-time-contributor approval that fork PRs need rather than anything wrong with the diff, so there is nothing for me to fix or re-push to chase. Flagging it only in case it had not been noticed.

No rush from me. If you would rather this were scoped differently, or you want the second emit site left out and only the trailing-comment one changed, say the word and I will redo it.

PR #1266, opened 2026-08-16
Fix access levels on nested protocols

### LINKED ISSUE #1265: `NoAccessLevelOnExtensionDeclaration` produces invalid code for a nested protocol
Given `Example.swift`:

```swift
public enum Namespace {}

public extension Namespace {
  protocol Boundary: Sendable {
    func work()
  }
}
```

Run:

```shell
swift format Example.swift
```

Expected output:

```swift
public enum Namespace {}

extension Namespace {
  public protocol Boundary: Sendable {
    func work()
  }
}
```

Current output with swift-format 6.3.0:

```swift
public enum Namespace {}

extension Namespace {
  protocol Boundary: Sendable {
    public func work()
  }
}
```

The current output is invalid Swift:

```text
error: 'public' modifier cannot be used in protocols
```

Protocol requirements implicitly have the protocol’s access level. The formatter should move `public` from the extension to the nested protocol declaration, not to the protocol requirement. The current rewrite also incorrectly changes `Boundary` from public to internal.

Dave861:
Fixes #1265

## Summary
- apply extension access levels to nested protocol declarations
- avoid descending into protocol requirements, which cannot have explicit access modifiers
- regenerate the rule pipeline and documentation

## Testing
- added coverage for ordinary protocols, explicit access levels, private extensions, and `@_spi`
- `swift test --filter NoAccessLevelOnExtensionDeclarationTests` (19 tests, 0 failures)
- full suite: 948 tests passed

### DIFF
--- a/Documentation/RuleDocumentation.md
+++ b/Documentation/RuleDocumentation.md
@@ -239,6 +239,8 @@ Lint: Specifying an access level for an extension declaration yields a lint erro
 Format: The access level is removed from the extension declaration and is added to each
         declaration in the extension; declarations with redundant access levels (e.g.
         `internal`, as that is the default access level) have the explicit access level removed.
+        For a nested protocol, the access level is added to the protocol declaration itself and
+        not to its requirements, which implicitly have the protocol's access level.
 
 `NoAccessLevelOnExtensionDeclaration` rule can format your code automatically.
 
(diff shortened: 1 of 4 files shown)


allevato (approved):
Thanks! This rule existed before it was possible to nest protocols and never got updated.

PR #1270, opened 2026-08-22
Run swift-format test suite in parallel

shahmishal:


allevato (approved):
I'm surprised this wasn't already the case. Thanks!

ahoppen:
swift-format runs tests faster sequentially than in parallel last time I checked ~2 years back. I don’t think much has changed since then.

allevato:
Good call. After doing a warm-up run to make sure everything was built and using the `time`:

* `swift test  11.19s user 2.28s system 113% cpu 11.895 total`
* `swift test --parallel  53.73s user 63.81s system 1006% cpu 11.684 total`

Looks like the improvement in wall time is negligible (and in a couple other runs, I was seeing wall times of 11.675 seconds, so the difference between them may be statistical noise).

This was on an M1 Max Mac Studio w/ 128GB RAM. I'm not sure what the numbers on the CI runners look like. It wouldn't surprise me if the subprocess invocations in a parallel run dominate those, since they probably have fewer resources/cores than my workstation.

So I guess we should measure this on CI before landing anything.

shahmishal:
with `--parallel` flag https://ci.swift.org/job/swift-PR-macos/31093/console
```
09:48:23.174  0.3%              	 105.38                	 Running tests for swiftformat
```
without `--parallel` flag https://ci.swift.org/job/oss-swift-pr-test-macoss/6293/console

```
09:28:29.574  0.3%              	 90.78                 	 Running tests for swiftformat
```

I wonder why parallel testing would be slower

ahoppen:
I think the tests run sufficiently fast that the overhead added by spawning separate test runners in `--parallel` adds more overhead than it saves.

PR #1279, opened 2026-08-29
Preserve trailing comments on case labels in NoCasesWithOnlyFallthrough

### LINKED ISSUE #1274: `NoCasesWithOnlyFallthrough` silently deletes a comment trailing the `case` label
## Problem                                                         
                                                                             
When `NoCasesWithOnlyFallthrough` merges a fallthrough-only case into the next
one, a comment on the same line as the `case` label is **dropped from the          
output entirely** — not moved, not preserved anywhere.                        
                                                                     
Input:                                                                     
                                                                                
```swift                                                                         
func f(x: Int) {                                                            
  switch x {                                                                   
  case 1:  // trail on 1                               
    fallthrough
  case 2:                                                            
    print("hi")                                                            
  default:                                                                     
    break                                                                      
  }                                                                          
}      
```                                                                             
                                                                       
`swift-format format` output:      

```swift           
func f(x: Int) {
  switch x {                                                               
  case 1, 2:                                      
    print("hi")
  default:                                                                     
    break                                              
  }                                                                           
}                                                                       
```

`// trail on 1` is gone. It also happens with a block comment
(`case 1:  /* block trail */`), and every comment in a chain is lost — three
cases with `// one` and `// two` collapse to `case 1, 2, 3:` with both gone.

`swift-format lint` reports only the ordinary rule finding
(`combine this fallthrough-only 'case' and the following 'case' into a single
'case'`), so nothing warns that applying it will discard source text.

## Expected behavior

The comment should be preserved. Suppressing the merge, as the rule already
does for the neighbouring comment positions, would be consistent and sufficient.

## Environment

- `swift-format 6.3.3` (Swift 6.3.3, `x86_64-unknown-linux-gnu`), default
  configuration

shubhransh-gupta:
### Motivation

Fixes #1274.

When `NoCasesWithOnlyFallthrough` merged a fallthrough-only `case` into the following case, comments trailing the `case` label (e.g. on the same line as the colon: `case 1:  // comment`) were discarded from the output AST because `isMergeableFallthroughOnly(_:)` only checked for comments adjacent to the statement or on the fallthrough token itself.

### Modifications

- In `Sources/SwiftFormat/Rules/NoCasesWithOnlyFallthrough.swift`, added checks for trailing comments on tokens within `switchCase.label` and trailing comments on the label itself before marking the case as mergeable.
- In `Tests/SwiftFormatTests/Rules/NoCasesWithOnlyFallthroughTests.swift`, added `testFallthroughCasesWithTrailingCaseLabelCommentsAreNotCombined` to ensure line and block comments on case labels prevent unwanted collapsing and preserve comments intact.

### Result

Comments trailing `case` labels in fallthrough-only cases are no longer silently deleted during formatting.

### DIFF
--- a/Sources/SwiftFormat/Rules/NoCasesWithOnlyFallthrough.swift
+++ b/Sources/SwiftFormat/Rules/NoCasesWithOnlyFallthrough.swift
@@ -145,9 +145,10 @@ public final class NoCasesWithOnlyFallthrough: SyntaxFormatRule {
     {
       return false
     }
-    // Check for any comments on or trailing the case label (e.g. after the colon).
+    // Check for any comments on or trailing the case label (e.g. after the colon or between patterns).
     if switchCase.label.tokens(viewMode: .sourceAccurate).contains(where: { token in
-      token.trailingTrivia.contains(where: { $0.isComment })
+      token.trailingTrivia.contains(where: { $0.isComment }) ||
+      token.leadingTrivia.contains(where: { $0.isComment })
     }) {
       return false
     }
--- a/Sources/SwiftFormat/Rules/NoCasesWithOnlyFallthrough.swift
+++ b/Sources/SwiftFormat/Rules/NoCasesWithOnlyFallthrough.swift
@@ -145,6 +145,17 @@ public final class NoCasesWithOnlyFallthrough: SyntaxFormatRule {
     {
       return false
     }
+    // Check for any comments on or trailing the case label (e.g. after the colon).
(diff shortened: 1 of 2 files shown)


shubhransh-gupta:
Hi @allevato @MaxDesiatov — Gentle ping when you have a moment to review. This PR resolves #1274 by ensuring comments trailing case labels in NoCasesWithOnlyFallthrough are preserved intact. All unit tests pass cleanly. Thanks!

PR #1287, opened 2026-09-08
cmake: recognize FreeBSD powerpc64le/powerpc64 in get_swift_host_arch

pkubaj:
- **Explanation**:
get_swift_host_arch() maps CMAKE_SYSTEM_PROCESSOR to a Swift architecture
name using the Linux spellings (ppc64le, ppc64).  On FreeBSD the
processor is reported as powerpc64le / powerpc64, so the PowerPC host was
falling through to the "Unrecognized architecture" error.  Accept the
FreeBSD spellings alongside the existing ones.
- **Scope**:
Nothing is removed, PR only adds new platform support.
- **Risk**:
No risk.
- **Testing**:
Build and runtime testing on FreeBSD/powerpc64le has been done.


rintaro (approved):

allevato (approved):
I don't know the CMake support well enough to have a feeling for whether this is something we should be sharing with other repositories instead of doing on our own, but this seems reasonable.

